PR #1, opened 2017-12-24
WIP - complete printing of all node types

mgrip:
Node kinds left:

- [ ] inline
- [ ] magic
- [ ] nowdoc
- [x] yield
- [x] yieldfrom
- [x] sys
- [x] trait
- [x] constant
- [x] classconstant
- [x] retif
- [ ] eval
- [x] exit
- [ ] halt
- [x] clone
- [x] declare
- [x] global
- [x] static
- [x] include
- [x] goto
- [ ] silent
- [x] try
- [x] catch
- [x] throw
- [ ] closure
- [x] new
- [x] label
- [x] traituse
- [x] traitalias
- [x] traitprecedence
- [ ] error


azz:
Cool stuff. `@` me if you run into any teething issues with the new API!

mgrip:
will do, on first glance looks awesome  - definitely makes sense. thanks @azz!

azz:
🤣 

image

tests/traits/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,67 @@
...
+trait C extends A implements B {
+  public function someOtherTestFunction() {
+    // 4
+  }
+  public function testFunction() {
+    // 5
+  }
+}
+class ImplementingClass {
+
```
vjeux: We’ll want to reuse the logic that ensures that there’s at most one empty line in between lines and none at the beginning or end

tests/sys/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,36 @@
...
+echo $test, $other;
+$info = array('coffee', 'brown', 'caffeine');
+list($drink, $color, $power) = $info;
+list(
+  $reallyReallyReallyReallyLongName,
+  $otherReallyReallyReallyLongName,
+  $lastOne
+) = $info;
+print "test print";
+print ("parens test");
```
vjeux: This looks weird with a space between the two

tests/declare/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,27 @@
...
+declare(ticks=1);
+$test = 1;
+declare(ticks=1) {
+  $test = 1;
+}
+declare(encoding='ISO-8859-1');
+$test = 1;
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+declare(ticks=1):
```
vjeux: In javascript we put spaces around =, is it idiomatic not to in php?

src/printer.js
```diff
@@ -613,21 +688,154 @@ function printStatement(node) {
...
+        concat([
+          "static",
+          indent(
+            concat([
+              line,
+              join(
+                concat([",", line]),
+                node.items.map(item => {
+                  // @TODO: hacking this for now. assignments nested inside a static
+                  // declaration doesn't have the operator set, so printing manually
```
vjeux: Would be nice to send issues/prs on the parser implementation. Prettier has found a ton of edge cases on all the parsers we used and having a good relationship with them was critical to make the project work.

vjeux:
This is weird, it passes locally

vjeux:
I reverted https://github.com/prettier/prettier-php/pull/1/commits/90a11a8a1493afad3fbd65a1ca30223eac79667d which is not passing, I'm not sure why as it's passing locally. Let's merge this and the other PR so that more people can contribute at the same time :)

tests/declare/__snapshots__/jsfmt.spec.js.snap
mgrip: typically yes - this is a bit of an edge case, its a system level directive http://php.net/manual/en/control-structures.declare.php

tests/traits/__snapshots__/jsfmt.spec.js.snap
mgrip: ya I agree - was thinking we can come back to it. I'll open an issue right now. Do you know where that logic exists in the js printer?

vjeux: In printStatementSequence (and in a few other places):

```js
if (util.isNextLineEmpty(text, stmt) && !isLastStatement(stmtPath)) {
      parts.push(hardline);
    }
```

PR #2, opened 2017-12-24
Exapand "encapsed" support

czosel:
Support for curly syntax is not complete - see
https://github.com/glayzzle/php-parser/issues/101


mgrip:
I know the parser doesn't fully handle the curly's - should we just comment out that test case or something for now? would be good to keep the tests passing if we can

azz:
You could use https://facebook.github.io/jest/docs/en/configuration.html#testpathignorepatterns-array-string

vjeux:
What we've done for prettier is to keep those as github issues. Committing code that's not working is usually a bad idea: people get confused as to why it's there and not working.

PR #4, opened 2017-12-25
Adding errors

vjeux:
Trying to figure out why it didn't pass CI in #1


vjeux:
The issue was that the ast name is `always` but we need to print `finally`. It was caught by the pp(pp(x)) == pp(x) check which doesn't run by default locally.

PR #5, opened 2017-12-25
Add support for cast, silent, halt; minor fixes and cleanup

czosel:


tests/operations/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -104,4 +105,5 @@ $a . $b;
 $a .= $b;
 $a instanceof stdClass;
 $output = \`ls -al\`;
+$a = (boolean) "test";
```
vjeux: (bool) got turned into (boolean)?

mgrip: I think both are acceptable - http://php.net/manual/en/language.types.type-juggling.php. do people have thoughts on enforcing one convention or reusing what was originally written?

vjeux: We have the same situation for flow and we normalize it to one with prettier

nicoder: both are acceptable for casting, but it seems only the short form is valid for type hints, so it may be better to use the short form also when casting ?

http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

PSR 12 seems to lean that way: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#25-keywords-and-types
`Short form of type keywords MUST be used i.e. bool instead of boolean, int instead of integer etc.`

czosel: I agree, always using the short form seems most clear and consistent. I opened a PR in the parser, i think it should be fixed there: https://github.com/glayzzle/php-parser/pull/102

czosel:
Thanks for merging this one! The PR in the parser has just been merged too and will be released as 2.2.0 soon. https://github.com/glayzzle/php-parser/pull/102

vjeux:
@czosel I just added you as contributor so you can merge PRs :)

czosel:
@vjeux Awesome, thanks! With great power comes great responsibility, i hope i can handle it :wink:

PR #10, opened 2017-12-30
Bump php-parser

czosel:


yarn.lock
```diff
@@ -2593,51 +2575,6 @@ preserve@^0.2.0:
...
-    remark-frontmatter "1.1.0"
-    remark-parse "4.0.0"
-    resolve "^1.5.0"
-    semver "5.4.1"
-    string-width "2.1.1"
-    typescript "2.6.2"
-    typescript-eslint-parser "11.0.0"
-    unicode-regex "1.0.1"
-    unified "6.1.6"
-
```
czosel: @azz not sure why running `yarn` after bumping php-parser caused this change - do you know more?

azz: Maybe we forgot to checkin the file last time Prettier was bumped.

Issue #12, opened 2018-01-06
Implementation question: Usage of fastPath


czosel:
I wanted to tackle #7, so I compared the implementation of the JS and python printers with the PHP one. I noticed that both JS and python rely heavily on calls to `path.call`, `patch.each` and so on, for example

https://github.com/prettier/prettier-python/blob/0e5c2a77ceebe2814707b7c61d1d003b1186b564/src/printer/index.js#L125

while in the PHP printer we're usually just calling to the internal function `printNode`, for example
https://github.com/prettier/prettier-php/blob/cd78cbb906aabeaed8c1e9588d7d89a6e89bae07/src/printer.js#L123

While `fastPath` seems to be used mostly for performance reasons (please correct me if I'm wrong), the call to `print` seems important because it calls the core function [`printGenerically`](https://github.com/prettier/prettier/blob/e4e3c757fb85f47f0336fc2f13678ae93facde40/src/main/ast-to-doc.js#L22-L49) which handles comments, for example.

Does this mean that we should refactor the printer to use `path.call`, `path.each`, ...? Since this would be a major change and I guess it should've come up earlier, I thought I'd open an issue before getting started :wink: Thanks!

vjeux:
Yes, you need to refactor everything to use path. This is annoying but it gives you two things:
1) comment handling around every single node
2) access to the parents all the way to the root

It may be used for other things that I’m not aware of though. This whole fastPath stuff is very weird and I still don’t fully understand why it’s there (it’s one of the few heritage from the original recast fork) but it worked for prettier so I guess it’s useful :p

czosel:
I also took a while to understand what it's doing! :sweat_smile:
Thanks for the clarification, I hope I'll find the time to refactor this soon!

azz:
It also allows the multiparser feature to exist 😄

PR #13, opened 2018-01-07
Refactor printNode to print

czosel:
As discussed in #12


czosel:
Tests are passing! 🎉

mgrip:
I still need to wrap my head around what's going on the hood with `path` 😅 Thanks for doing this!

Issue #7, opened 2017-12-25
Line spacing within files


mgrip:
Make sure that when printing code there’s at most one empty line in between lines and none at the beginning or end. We should hopefully be able to reuse the same logic implemented in the JS printer

mgrip:
Fixed w/ @czosel's PR #14 - closing out

PR #14, opened 2018-01-13
Fix line spacing - refs #7

czosel:
Even though this is still work in progress, I thought I'd share this already because the way locations are specified by the PHP parser seems to require patching prettier core -
 see https://github.com/czosel/prettier/commit/0d8251b99969f5e8f1e7207213379dc23254d8ca

It's a small change and seems reasonable to me, but maybe someone knows a better way to do this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ node_js:
   - 6
   - 8
   - 9
+cache:
+  yarn: true
+  directories:
+  - node_modules
 install:
   - yarn install
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,6 @@ node_js:
   - 6
   - 8
   - 9
-cache:
-  yarn: true
-  directories:
-  - node_modules
 install:
   - yarn install
 script:
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   ],
   "dependencies": {
     "php-parser": "2.2.0",
-    "prettier": "prettier/prettier#fac08034e78f5eaf8ef3bbcb57d12893f84c3583"
+    "prettier": "prettier/prettier#b449a526ce0f16703cdf5b4b22731b2a94775473"
   },
   "devDependencies": {
     "eslint": "4.1.1",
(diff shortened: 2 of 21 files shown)


vjeux:
Yeah, the location extraction code has not been split out into plugins yet. This is the only way to do it right now. I won't have the time to split it out but maybe @azz or you could do it. It shouldn't be hard: we need to expose a locStart and locEnd function to plugins and move all the existing ones to their plugins.

azz:
This is because `src/common/util` is not part of the Prettier API. The bundled version on npm does not have this file. It is also a very large interface that it probably doesn't make sense to support. 

Maybe we can look at supporting a subset of these util functions?

vjeux:
locStart and locEnd are ast-dependent, so they should live in the plugin at least

azz:
Yeah but the only reason `locStart` is called is because this plugin is calling a Prettier internal `util.isNextLineEmpty()`

vjeux:
isNextLineEmpty() is very useful for comments and empty lines.

azz:
We could do something like `babel` does with their plugin API, instead of importing `prettier/lib/util` you get it as an argument to `print()`. 

```js
function genericPrint(path, options, print, util) {
  util.isNextLineEmpty(...);
}
```

@hzoo is this something that you'd recommend?

czosel:
My first idea was to pass `options` to the various methods in `util` that end up calling `locStart` or `locEnd` - which looks quite messy and requires many changes. 
Passing `util` as an argument to print seems like a good idea - probably it should only contain functions which are 
* explicitly meant to be consumed by plugins, and 
* correctly preconfigured if necessary (e.g. with the correct `locStart` and `locEnd` implementation)

Maybe it would make sense to split the current `util` in a private and  a public part?

hzoo:
> We could do something like babel does with their plugin API, instead of importing prettier/lib/util you get it as an argument to print().

@azz I think depends on how you want to do versioning/deps on prettier? Only issue with passing it into the plugin is that you get the version passed down from the dep. We are going to move to a peerDep for `@babel/core`, since you may get version mismatches or the out of date version if the prettier version in php is old?

czosel:
@azz Can't seem to get CI to run even with the explicit export - does this only work if im loading prettier from npm?

czosel:
Tests are passing 🎉

PR #19, opened 2018-01-27
Fixed the start command

williamdes:


azz:
Neither of those flags should be required.

Plugins are automatically loaded when they are named `@prettier/plugin-*`, and the parser is inferred from the file extension.

williamdes:
Good to know, dit not work for me...

azz:
Hmm. Could be an outdated package on npm.

Issue #17, opened 2018-01-23
Is this going to enforce PSR-1? PSR-2? What about upcoming PSR-12?


QWp6t:
Unlike JavaScript community, PHP has widely adopted code style already that's used by all the major framework vendors. Will this library be implementing those code styles? Or will you guys be doing your own thing?

Currently, [PSR-2](http://www.php-fig.org/psr/psr-2/) (and therefore also [PSR-1](http://www.php-fig.org/psr/psr-1/)) is what everyone uses, but [PSR-12](https://github.com/php-fig/fig-standards/blob/eb3618d4bc3e1daf0423ef826a4b3f4335d7fff4/proposed/extended-coding-style-guide.md) is currently under review as a replacement for PSR-2.

czosel:
In my opinion, it makes most sense to output idiomatic PHP code, which is probably PSR-2 compliant. So no, I wouldn't say we want to do our own thing! 😉 
On a side note: I haven't looked at PSR-12 yet - is it very different from PSR-2?

QWp6t:
PSR-12 is the same as PSR-2 for the most part from what I can tell. It's a little more strict (it has some opinions on things that were omitted in PSR-2, such as more rules within PHP headers), but its primary focus is to create rules for things that didn't exist within the language back when PSR-2 was drafted, such as type hinting and `declare()` statements.

theofidry:
@czosel in that case I recommend you to take a look at:

- [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) which is used for [styleci.io](styleci.io) as well
- [PHP-CS](https://github.com/squizlabs/PHP_CodeSniffer)

It might be worth to leverage one of those in order to achieve what you want, it will likely save you a lot of work.

MichaelDeBoey:
My vote goes to using [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) then 🙂

czosel:
@theofidry I'm not completely sure if i understand your comment correctly - it was my understanding that PHP-CS(-Fixer) has the same relationship to prettier-php as eslint (--fix) has to prettier - which means that it solves a different problem but can be used together. 
Assuming that's correct - how could we use PHP-CS to save a lot of work?

Narayon:
PHPCS has a built-in beautifier and fixer called PHPCBF.  
It follows whatever standard you choose - I'm using WordPress-Extra.  
[...]

PR #21, opened 2018-02-09
(WIP) adding test case and fixing formatting to align with psr2

mgrip:
Addressing #17 

Main thing was just moving braces on a new line for methods. Adding as WIP for now since I'm not sure how to increase indentation spacing from 2 spaces to 4...

Using this as a guide: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md - probably more examples that will need to be added

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -427,18 +427,18 @@ function printStatement(path, options, print) {
               indent(
                 concat([
                   node.extends
-                    ? concat([line, "extends ", path.call(print, "extends")])
+                    ? concat([" extends ", path.call(print, "extends")])
                     : "",
                   node.implements
                     ? concat([
-                        line,
-                        "implements ",
+                        " implements ",
                         join(", ", path.map(print, "implements"))
                       ])
                     : ""
                 ])
               ),
-              " {"
+              hardline,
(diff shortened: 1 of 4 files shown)


czosel:
Looks good, shall we handle the indent size in a separate PR?

mgrip:
ya i was trying to leave the test to just match the actual example from the psr2 docs. Took a quick look and I think this just gets solved by changing the `tabWidth` option from 2 to 4. I can change the default for our tests [here](https://github.com/prettier/prettier-php/blob/master/tests_config/run_spec.js#L141), but do you guys have an opinion more largely on plugins being able to override the default options?

czosel:
@mgrip this might work already by exporting `options` in `index.js`: https://github.com/prettier/prettier/pull/3924

mgrip:
Ugh I couldn't get the plugin-specific options to work. This PR is already starting to get bigger than I'd like so I'll open up a new issue to figure out how to override the default value for `tabWidth` - sound good @czosel?

azz:
@mgrip Let me know if you need any assistance with the plugin-specific options!

PR #23, opened 2018-02-13
Optimize wrapping of long calls

czosel:
I wanted to optimize the handling of long calls like this one:
```php
$this->something->method($argument, $this->more->stuff($this->even->more->things->complicatedMethod()));
```
which used to be turned into
```php
$this->something->method($argument, $this->more->stuff($this->even->more->
  things->
  complicatedMethod()));
```
and would now be turned into
```php
$this->something->method(
  $argument,
  $this->more->stuff($this->even->more->things->complicatedMethod())
);
```
This seems to be PSR-2 compliant as far as i can tell ([see here](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#46-method-and-function-calls)).

However, it's this would also change the way functions like `array_map` would be printed like:
```php
array_map(
  function ($entry) {
    return $entry * 2;
  },
  $arr
);
```
instead of
```php
array_map(function ($entry) {
    return $entry * 2;
  }, $arr);
```
which looks a bit strange. Let me know what you think 😄

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -737,12 +737,22 @@ function printStatement(path, options, print) {
         "}"
       ]);
     case "call":
-      return concat([
-        path.call(print, "what"),
-        "(",
-        join(", ", path.map(print, "arguments")),
-        ")"
-      ]);
+      return group(
+        concat([
+          path.call(print, "what"),
+          concat([
+            "(",
+            indent(
+              concat([
+                softline,
+                join(concat([",", line]), path.map(print, "arguments"))
+              ])
+            ),
+            softline,
+            ")"
+          ])
+        ])
+      );
     case "usegroup":
       return concat([
         "use ",
--- a/tests/class/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/class/__snapshots__/jsfmt.spec.js.snap
@@ -35,6 +35,8 @@ abstract class ReallyReallyReallyLongClassName extends AbstractModelFactoryResou
     return self::$staticTest[0];
   }
 }
+
+$this->something->method($argument, $this->more->stuff($this->even->more->things->complicatedMethod()));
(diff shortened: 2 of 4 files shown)


mgrip:
Awesomeee that looks so much better. Plus I think in the case you mentioned having them on separate lines is actually the way to go, if I'm interpreting this correctly
> Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line.

czosel:
Alright, then let’s go with this - it’s certainly an improvement over the current implementation ;)

nicoder:
As far as I can tell, both are PSR-2-compliant, and PSR-12 adds precisions because there was debate about that:

https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#47-method-and-function-calls

for example PSR-12 lists this is valid:

```php
$app->get('/hello/{name}', function ($name) use ($app) {
    return 'Hello ' . $app->escape($name);
});
```

but this would not be :

```php
$app->get('/hello/{name}',
    function ($name) use ($app) {
        return 'Hello ' . $app->escape($name);
    }
);
```

I understand it this way:
- if one argument is defined starting on a new line, then all of the arguments must start on their own line
- but arguments may take up more than one line (independently of whether they start on their own line or not)

For example it may be nicer to see:

```php
foo([
   $longExpression,
   $prettyLongExpression,   
]);
```

than:

```php
foo(
    [
       $longExpression,
       $prettyLongExpression,   
    ]
);
```

Or nicer to see:

```php
$db->Execute($sql, [
    $foo,
    $somewhatLongParameter,
]);
```

than:

```php
$db->Execute(
    $sql,
    [
        $foo,
        $somewhatLongParameter,
    ]
);
```

czosel:
Thanks @nicoder! I created an issue for this.

Issue #18, opened 2018-01-24
@prettier/plugin-php not found in the registry


lgraziani2712:
Right now, the name of the npm lib is `@prettier/php`.

Thank you for your work!

azz:
Yeah I haven't published it yet since it's under active development. I'll publish the latest version to the new name soon.

czosel:
@azz Maybe it would be a good time to publish a first version? Then we could close #24 and the README would not be misleading anymore. I think the big `WORK IN PROGRESS` note in the README is clear enough that users don't expect production quality yet 😄

andreiglingeanu:
That is a great idea @czosel, would love to try the prettifier in my day-to-day work. I believe this will actually just speed things up, more people would instantly be able to provide feedback

azz:
Just published master as `@prettier/plugin-php` and renamed the repo to `prettier/plugin-php` for consistency.

PR #24, opened 2018-02-13
Fix yarn install command

dabroek:


azz:
Thanks for this but it was just not yet published :) see https://github.com/prettier/plugin-php/issues/18#issuecomment-366414177

PR #28, opened 2018-02-14
Use multi-line print for all multi-line node types

mgrip:
Reusing solution from #14 for other functions that need to print multiple lines

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -843,23 +843,16 @@ function printStatement(path, options, print) {
         const directive = Object.keys(path.getValue().what)[0];
         return concat([directive, "=", path.call(print, "what", directive)]);
       };
-      const printDeclareChildren = function(path) {
-        return concat(
-          path.map(child => {
-            return concat([
-              print(child),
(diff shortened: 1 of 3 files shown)


mgrip:
@czosel i just moved the functions you made to handle line spacing to the top w/ the other helper functions - but let me know if you think there's a better way to organize things!

src/printer.js
```diff
@@ -1033,15 +1054,14 @@ function printNode(path, options, print) {
...
-                        lineShouldEndWithSemicolon(adaptationPath) ? ";" : ""
-                      ]);
-                    }, "adaptations")
-                  )
+                  concat([
+                    line,
+                    path.call(
+                      adaptationsPath =>
+                        printLines(adaptationsPath, options, print),
+                      "adaptations"
```
czosel: indentation issue?

mgrip: so this is actually how prettier formats it haha - `"adaptations"` is the second argument for `call()`, the inline arrow function is the 1st argument. I agree its a little muddy though...

czosel: Oh, right ^_^

Issue #34, opened 2018-02-14
Function parameters passed by reference


mgrip:
input:
```php
function pass_by_reference_test($x, &$a) {
  $a + 1;
}
```
output:
```php
function pass_by_reference_test($x, $a) {
  $a + 1;
}
```

input:
```php
class Test {
  public function test_pass_by_reference(&$x) {
    $x + 1;
  }
}
```
output:
```php
class Test {
  public function test_pass_by_reference($x) {
    $x + 1;
  }
}
```

czosel:
Fixed in #35

PR #35, opened 2018-02-14
Fixing pass by reference for functions and methods

mgrip:
Addressing #34


mgrip:
Added in a fix for #36 because I didn't want to get merge conflicts. One note is that the parser currently doesn't handle all native types - once [this](https://github.com/glayzzle/php-parser/issues/113) is addressed we can update the test case

Issue #44, opened 2018-02-15
implement continue node


mgrip:
https://github.com/glayzzle/php-parser/blob/master/docs/AST.md#continue

mgrip:
Fixed in #48 - closing

PR #48, opened 2018-02-16
include optional number after `break` statements, add support for continue

### LINKED ISSUE #16: `break` statements stripped of optional number
I've created a new branch (nikulis/prettier-php@7ee8034) to address this, but did not want to PR at the moment since it is based on the already-open #14.

---

The php `break` statement should accept an optional trailing integer denoting the number of nested control statements to break. Currently, any trailing integer is removed.

Input:

```php
<?php

while(true) {
  while(true) {
    break 2;
  }
}
```

Current output:

```php
<?php

while(true) {
  while(true) {
    break;
  }
}
```

---

There are a couple other quirks surrounding the use of the `break` statement that may or may not fall under the purview of this project to auto-correct and/or offer configurable parameters for:

1. [Apparently](https://secure.php.net/manual/en/control-structures.break.php#121615), php also allows for parentheses around the trailing integer, making it look like a function call, e.g. `break(2);`. Could we automatically remove the parentheses so that all breaks follow the same convention?

2. `break 1;` could be collapsed to `break;`.

czosel:
Fixes #16 and #44.
(just rebased @nikulis 's branch)


czosel:
In order to get the tests running, we still need to implement `massageAstNode` ([example for js](https://github.com/prettier/prettier/blob/master/src/language-js/clean.js))

czosel:
Tests are passing! 🎉

czosel:
It might have been forgotten at an earlier time - when I'm installing from the current lockfile i'm getting `TypeError: Cannot read property 'isNextLineEmpty' of undefined` (outdated core dependency)

PR #51, opened 2018-02-17
tests: a lot of arrays test

evilebottnawi:
Just tests


czosel:
Hi @evilebottnawi, thanks for the contribution! 

I'm always a fan of thorough testing, but in this case i'm not 100% sure if we need this level of detail - but i'd be curious to hear @mgrip 's opinion!

MichaelDeBoey:
The more detailed test, the less regression we can get I think?
If the tests are PSR-2 compliant of course 🙂 (haven't had a look at them)

mgrip:
Ya I'd say in general there are some more glaring issues for us to cover first but it probably can't hurt to have more detailed test cases for this, thanks @evilebottnawi! These look good, I'm game to merge in

PR #56, opened 2018-02-19
fix: don't output extra semicolon

evilebottnawi:
:star: Wait invite :smile:

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -377,7 +377,7 @@ function printStatement(path, options, print) {
           ";",
           concat(
             path.map(
-              usegroup => concat([hardline, print(usegroup), ";"]),
+              childrenPath => concat([hardline, print(childrenPath)]),
               "children"
             )
           )
@@ -812,7 +812,8 @@ function printStatement(path, options, print) {
               concat([",", line]),
               path.map(item => concat([print(item)]), "items")
             ),
-            node.name ? "}" : ""
+            node.name ? "}" : "",
+            ";"
           ])
         )
       );
--- a/tests/formatting/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/formatting/__snapshots__/jsfmt.spec.js.snap
@@ -50,6 +50,6 @@ class Foo extends Bar implements FooInterface
   {
     // method body
   }
-};
+}
 
 `;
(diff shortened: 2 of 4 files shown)


src/printer.js
```diff
@@ -811,7 +811,8 @@ function printStatement(path, options, print) {
               concat([",", line]),
               path.map(item => concat([print(item)]), "items")
             ),
-            node.name ? "}" : ""
+            node.name ? "}" : "",
+            ";"
```
czosel: Can you add a test for this?

evilebottnawi: @czosel we already have ugly output https://github.com/prettier/plugin-php/pull/56/files#diff-07d375fafb7ea50b83dfe5d31b5b2d33L53

czosel: But i don't see a test for the added semicolon in `usegroup`?

czosel: Ah sorry, i was misreading the code changes - forget what i said :wink:

src/printer.js
```diff
@@ -376,7 +376,7 @@ function printStatement(path, options, print) {
           ";",
           concat(
             path.map(
-              usegroup => concat([hardline, print(usegroup), ";"]),
+              usegroup => concat([hardline, print(usegroup)]),
```
czosel: Shall we rename `usegroup` to `child` here? I've been confused about this because i didn't realize that the children of the namespaces can be not only usegroups, but also classes...

evilebottnawi: @czosel yep,
```php
namespace My {
 // Here we can have any stuff
}
```
Let's do this.

evilebottnawi: @czosel but parser always interpreted as `usegroup`, let wait what says @mgrip

czosel: Are you sure? In my debugger ist just looked like in the test case there were 3 `usegroup`s and one `class` in `children`. Since we're just calling `print` on the children, the parser doesn't care what kind of node it finds.

evilebottnawi: @czosel let's double check :+1:

evilebottnawi:
@czosel done, also add tests for use without names, it is remove `;` from each `use`

evilebottnawi:
@czosel In the near future I have about a month of free time, if everything is good, then in a month we will be able to make the first stable release

czosel:
@evilebottnawi awesome, I really hope we'll have a first working version soon!

czosel:
@azz can you invite @evilebottnawi to this repository?

mgrip:
just make sure to get tests passing before this gets merged! Some of the errors are just updating snapshots I think - the other ones look like they might be related to whatevers going on in #61...

PR #57, opened 2018-02-19
fix: respec special characters inside string

### LINKED ISSUE #29: Don't change \n to an actual newline
First up, awesome work on this project, I'm really excited. I seem to recall there was some discussion about this for JavaScript but I can't find the issue.

Right now this converts
```php
$foo = "hello\nworld"
```
into
```php
$foo = "hello
world"
```

Personally I think it is nicer to leave the `\n` alone.

evilebottnawi:
Fixes #29

Why parse by default don't escape special characters :confused:

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -11,6 +11,20 @@ const indent = docBuilders.indent;
 const hardline = docBuilders.hardline;
 const softline = docBuilders.softline;
 
+const makeString = util.makeString;
+
+function stringEscape(str) {
+  return str
+    .replace(/[\\]/g, "\\\\")
+    .replace(/["]/g, '\\"')
+    .replace(/[/]/g, "\\/")
+    .replace(/[\b]/g, "\\b")
+    .replace(/[\f]/g, "\\f")
+    .replace(/[\n]/g, "\\n")
+    .replace(/[\r]/g, "\\r")
+    .replace(/[\t]/g, "\\t");
+}
+
 // polyfill for node 4
 function includes(array, val) {
   return array.indexOf(val) !== -1;
@@ -218,7 +232,7 @@ function printExpression(path, options, print) {
         // use setting from options. need to figure out how this works w/ complex strings and interpolation
         // also need to figure out splitting long strings
         const quote = node.isDoubleQuote ? '"' : "'";
-        return quote + node.value + quote;
+        return makeString(stringEscape(node.value), quote, true);
       }
       case "number":
         return node.value;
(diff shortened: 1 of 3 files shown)


mgrip:
Awesome! @evilebottnawi do you think its worth opening an issue on the [parser repo](https://github.com/glayzzle/php-parser/issues)? Not sure if this is something that should eventually be changed there or not

evilebottnawi:
@mgrip looks on other parsers https://astexplorer.net/#/gist/070b3c0b39f0af268d909e3ba9efa2a3/2ba4c9d3ab6b3e3cf31f6148a28966c638948790
They always have `value` and `rawValue` for this

evilebottnawi:
@mgrip i am not familiar with php-parser maybe some option(s) missed?

mgrip:
Ah got it, that makes sense. In that case I think it just makes sense to do this here

src/printer.js
```diff
@@ -11,6 +11,20 @@ const indent = docBuilders.indent;
 const hardline = docBuilders.hardline;
 const softline = docBuilders.softline;
 
+const makeString = util.makeString;
+
+function stringEscape(str) {
```
mgrip: can you just add a comment explaining why we need to do this? 🙏

evilebottnawi: @mgrip I think I will first create the problem in php parser, let's wait answer

mgrip: sounds good to me 👍

mgrip: if you want to push this just with a comment w/ a link to the issue in the parser repo that works for me - that way we dont have to wait for it to get fixed on their end

evilebottnawi:
Ref: https://github.com/glayzzle/php-parser/issues/116

src/printer.js
evilebottnawi: @mgrip I think we will do this if we do not receive an answer at the earliest acceptable time

evilebottnawi:
/cc @mgrip @czosel 
I suggest you such a strategy:
1. If parser have problem.
2. Determine whether it is possible to solve it at the level of parser/printer inside plugin
3. If yes - add solution with `@Todo` comments with reference on issue in `php-parser`, If no - sorry

This will allow you to adapt the prettier in the early stages to understand where improvements are possible, at this time this and `inline` problem don't adapt prettier on popular code bases.

mgrip:
Def agree. I think we've pretty much been doing this already for the most part but I think it makes sense to formalize a plan going forward

mgrip:
@evilebottnawi does that mean you're going to get this ready to deploy, to workaround the parser? I think in terms of outstanding issues this is one of the most glaring breaking issues

evilebottnawi:
@mgrip I do not quite understand, i want just add in `util` some function with workaround parser problem, today i am do this and we will review this :+1:

mgrip:
Yup, sounds good! I was just voicing my desire to get a fix deployed for this, if not in the parser then a work-around on our end 😄

evilebottnawi:
Close in favor https://github.com/prettier/plugin-php/pull/136

PR #58, opened 2018-02-19
fix: always output newline in EOF

evilebottnawi:
:star:


czosel:
:+1: Didn't even know about this rule :laughing:

PR #62, opened 2018-02-19
fix: `abstract` method

evilebottnawi:
:star:

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -61,6 +61,9 @@ function lineShouldEndWithSemicolon(path) {
   if (node.kind === "traituse") {
     return !node.adaptations;
   }
+  if (node.kind === "method" && node.isAbstract) {
+    return true;
+  }
   return includes(semiColonWhitelist, node.kind);
 }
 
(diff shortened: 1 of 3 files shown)


src/printer.js
```diff
@@ -526,6 +527,7 @@ function printStatement(path, options, print) {
           ),
           ")",
           node.type ? concat([": ", path.call(print, "type")]) : "",
+          node.isAbstract ? ";" : "",
```
mgrip: should we handle this in `lineShouldEndWithSemiColon` instead?

evilebottnawi: @mgrip it is part of `method` construction and it is not solve in `lineShouldEndWithSemicolon`, we only add semicolon in `lineShouldEndWithSemicolon` after body

mgrip: `lineShouldEndWithSemicolon` gets run on every line of code, and checks if that node type should have a semi colon - in this case i would think if if we see the "method" node type we can just check if its abstract or not, and if it is say that it should end with a semicolon? Don't think its a huge deal, but I think its a nice pattern to have that function be solely responsible for handling semi-colon usage

evilebottnawi: @mgrip let'd do this :+1:

evilebottnawi:
@mgrip done :+1:

PR #63, opened 2018-02-19
fix: clean-ast breaks when break level is null

mgrip:


czosel:
I forgot the most common case 🤦‍♂️

Issue #70, opened 2018-02-19
Block comments with incorrect spacing are missing lines


mgrip:
input:
```php
/**
 * Multi line
* test with wrong indentation
*/
```
output:
```php
/**
 * Multi line
 */
```

mgrip:
Fixed in #71

PR #71, opened 2018-02-19
fix: comments with incorrect spacing

mgrip:
#70


mgrip:
Great question - I honestly don't know. Should we merge this for now and then investigate further? I think this is a little more stable than the way we're currently handling comments

evilebottnawi:
@mgrip i agree, merge this as is right now and create issue about this

PR #73, opened 2018-02-20
feat: lowercase all keywords

evilebottnawi:
:star: 

Other improvement:
1. Support alternative syntax
2. NULL => `null`
3. magic constant to upper case `__line__` => `__LINE__`

Parser problem:
1. https://github.com/glayzzle/php-parser/issues/117

### DIFF
--- a/src/clean.js
+++ b/src/clean.js
@@ -16,10 +16,34 @@ function clean(ast) {
         : { kind: ast.kind, level };
     }
   }
+  // Normalize numbers
   if (ast.kind === "number") {
     ast.value = util.printNumber(ast.value);
 
     return ast;
+  }
+  // All magic constant should be upper case
+  if (ast.kind === "magic") {
+    return ast.value.toUpperCase();
+  }
+
+  // All reserved words should be lowercase case
+  if (ast.kind === "identifier" && typeof ast.name === "string") {
+    const lowerCasedName = ast.name.toLowerCase();
+    const isLowerCase =
+      [
+        "int",
+        "float",
+        "bool",
+        "string",
+        "null",
+        "void",
+        "iterable",
+        "object",
+        "self"
+      ].indexOf(lowerCasedName) !== -1;
+
+    return isLowerCase ? lowerCasedName : ast.name;
   }
 }
 
--- a/src/printer.js
+++ b/src/printer.js
@@ -258,7 +258,8 @@ function printExpression(path, options, print) {
       case "inline":
         return node.value;
       case "magic":
-        return node.value;
+        // for magic constant we prefer upper case
+        return node.value.toUpperCase();
       case "nowdoc":
         return concat([
           "<<<'",
@@ -1072,14 +1073,31 @@ function printNode(path, options, print) {
     return printStatement(path, options, print);
   }
   switch (node.kind) {
-    case "identifier":
+    case "identifier": {
       // this is a hack until https://github.com/glayzzle/php-parser/issues/113 is resolved
+      // for reserved words we prefer lowercase case
       if (node.name === "\\array") {
         return "array";
       } else if (node.name === "\\callable") {
         return "callable";
       }
-      return node.name;
+
+      const lowerCasedName = node.name.toLowerCase();
+      const isLowerCase =
+        [
+          "int",
+          "float",
+          "bool",
+          "string",
+          "null",
+          "void",
+          "iterable",
+          "object",
+          "self"
+        ].indexOf(lowerCasedName) !== -1;
+
+      return isLowerCase ? lowerCasedName : node.name;
+    }
     case "case":
       return concat([
         node.test
--- a/tests/case/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/case/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,488 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`case.php 1`] = `
+<?php
+DECLARE(ticks=1);
+
+NAMESPACE Vendor\\Package;
+
+USE FooClass;
+USE BarClass AS Bar;
+
+INTERFACE iTemplate {}
+
+ABSTRACT CLASS AbstractClass
+{
+    ABSTRACT PROTECTED FUNCTION getValue();
+    ABSTRACT PROTECTED FUNCTION prefixValue($prefix);
+
+    PUBLIC function printOut() {
+        PRINT $this->getValue();
+    }
+}
+
+TRAIT ezcReflectionReturnInfo {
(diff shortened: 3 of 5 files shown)


src/printer.js
```diff
@@ -1053,6 +1053,8 @@ function printNode(path, options, print) {
         return "array";
       } else if (node.name === "\\callable") {
         return "callable";
+      } else if (node.name.toLowerCase() === "null") {
```
evilebottnawi: By PSR-2

src/printer.js
```diff
@@ -166,14 +167,25 @@ function printExpression(path, options, print) {
...
+      case "bin": {
+        let type = node.type;
+
+        if (type.toLowerCase() === "and") {
+          type = "&&";
+        }
+
+        if (type.toLowerCase() === "or") {
+          type = "||";
+        }
```
evilebottnawi: More pretty

czosel: Does this also handle the difference between `||` and `or`? (https://stackoverflow.com/questions/17710924/logical-operators-or-vs-double-pipe-in-php)

tests/operations/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -100,8 +100,8 @@ $a <= $b;
 $a >= $b;
 $a <=> $b;
 
-$a and $b;
-$a or $b;
+$a && $b;
+$a || $b;
```
czosel: `or` and `||` are not the same - we can't make this substitution.

src/printer.js
czosel: :+1:

evilebottnawi:
Seems we don't account short from :confused: Let's do this

tests/switch/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -21,6 +25,10 @@ switch (2) {
   case 2:
     $test = 'second';
     break;
+  case 10:
+  case 20:
```
evilebottnawi: Pretty :+1:

tests/case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,457 @@
...
+  {
+    echo self::CONSTANT;
+  }
+
+  protected function MyProtected()
+  {
+    require ('somefile.php');
+    require_once ('somefile.php');
+    include ('somefile.php');
+    include_once ('somefile.php');
```
evilebottnawi: `require ('somefile.php');` => `require 'somefile.php';` 

All popular framework use this form
https://github.com/laravel/framework/search?l=PHP&q=require&type=&utf8=%E2%9C%93
https://github.com/symfony/symfony/search?utf8=%E2%9C%93&q=require&type=
https://github.com/yiisoft/yii2/search?l=PHP&q=require&type=&utf8=%E2%9C%93

evilebottnawi: Let's do this in other PR better :smile:

tests/exit/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -4,14 +4,13 @@ exports[`exit.php 1`] = `
 <?php
 exit("exit info");
 exit(1);
-__halt_compiler();
-echo "this is not parsed anymore!";
+die("die info");
```
evilebottnawi: Move test to `half` directory

tests/exit/exit.php
```diff
@@ -1,5 +1,5 @@
 <?php
 exit("exit info");
 exit(1);
-__halt_compiler();
-echo "this is not parsed anymore!";
+die("die info");
```
evilebottnawi: Already reported in php parser

tests/if/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -24,6 +26,8 @@ if (true) {
   return 1;
 } elseif (false) {
   return 2;
+} elseif (null) {
```
evilebottnawi: :+1:

evilebottnawi:
/cc @czosel @mgrip ready for review

In next PR i will fix handling mixed content:
```php
<div>Test</div>
<?php echo 'test'; ?>
<div>Foo</div>

**Comment by czosel** [comment]
Puh, this one got quite big - would you mind splitting out the parts that have nothing to do with case sensitivity into separate PRs?

**Comment by czosel** [comment]
Concerning mixed content: I think proper support for PHP in HTML is still blocked by proper HTML support in prettier, right? (See https://github.com/prettier/prettier/pull/3521)

**Comment by evilebottnawi** [comment]
@czosel no, we break code with mixed content right now, also not all mixed content is HTML, we should print mixed content as is to allow don't break code.

**Comment by evilebottnawi** [comment]
@czosel it is will difficult separate this PR on part some parts depend on other part. We can merge this as is and add this to changelog. Also right now we don't release any version and can merge big PR after review, it is normal. We will do same in prettier between release.

**Comment by evilebottnawi** [comment]
@mgrip what do you think about above?

**Comment by mgrip** [comment]
Sorry haven't had a chance to look through yet but I will today!

File: src/printer.js (hunk 1 above)
Comment (mgrip) [reply src/printer.js]: can you add a comment explaining we prefer lowercase?

File: src/printer.js [hunk 9]
```diff
@@ -1030,9 +1050,9 @@ function printStatement(path, options, print) {
     case "silent":
       return concat(["@", path.call(print, "expr")]);
     case "halt":
-      return concat(["__halt_compiler();", node.after]);
-    //@TODO: leaving eval until we figure out encapsed https://github.com/prettier/prettier-php/pull/2
+      return concat(["__halt_compiler();", node.after.replace(/\n$/, "")]);
```
mgrip: add comment explaining what the `.replace()` is for

src/clean.js
```diff
@@ -14,6 +14,14 @@ function clean(ast) {
         : { kind: ast.kind, level };
     }
   }
+
+  if (ast.kind === "magic") {
```
mgrip: can you add a comment for both of these explaining we change case in the printer?

evilebottnawi:
@mgrip i think comments about lowercase a bit overkill, by default we always prefer lowercase

evilebottnawi:
@czosel @mgrip after merge separate PR i rebase this PR and we can merge :+1:

mgrip:
@evilebottnawi thanks for separating everything! I think at this point everything else got merged so probably good to rebase now! Might even be easier to throw this out and start a new MR for the case stuff, up to you

evilebottnawi:
@mgrip need review https://github.com/prettier/plugin-php/pull/92, also i can do other PR to tests `else if` and need also review

Issue #31, opened 2018-02-14
Trailing commas in arrays


aboyton:
I couldn't find anything in the various PSR, but it seems acceptable to add trailing commas to arrays. Thoughts about making this the default?

czosel:
Found this: https://framework.zend.com/manual/2.4/en/ref/coding.standard.html#arrays
I could imagine that having trailing commas in multiline arrays, but no trailing comma in one-liners, would work well?

aboyton:
Ah yes, I was assuming multiline arrays when asking for this. Basically the same as in JavaScript if you ask for `comma=es5|all`.

PR #78, opened 2018-02-20
feat: trailing commas in array

evilebottnawi:
:star:


tests/trailing_commas/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,65 @@
...
+$foo = ['foo', 'bar',];
+$foo = [
+  'very-very-very-very-very-very-very-very-very-long-item-1',
+  'very-very-very-very-very-very-very-very-very-long-item-2',
+];
+$foo = [
+  'very-very-very-very-very-very-very-very-very-long-item-1' => 'foo',
+  'very-very-very-very-very-very-very-very-very-long-item-2' => 'bar',
+];
+$foo = array('foo', 'bar',);
```
czosel: In https://github.com/prettier/plugin-php/issues/31 we said that if the array fits on one line, we never want trailing commas.

evilebottnawi: @czosel oh yes, stupid error :smile:

evilebottnawi:
@czosel Done :+1:

czosel:
I'm not sure if we discussed already if we want to make this configurable with an option already? What would be the default?
/cc @mgrip

src/printer.js
```diff
@@ -16,6 +17,17 @@ function includes(array, val) {
...
 }
 
+function shouldPrintComma(options) {
+  switch (options.trailingComma) {
+    case "all":
+      return true;
+    // fallthrough
+    case "none":
+    default:
+      return false;
```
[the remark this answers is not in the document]
evilebottnawi: @czosel in future i think we can add option `php7`, `php7.2`, `php7.3` for avoid break code and allow insert trailing commd

evilebottnawi:
@czosel laravel already use trailing comma in array, they don't adapt prettier if we ignore this PR

czosel:
My gut feeling also says that this might be an option that makes sense. Do you have any intuition on a proper default value?

evilebottnawi:
@czosel default value as in `prettier`, i.e. in `prettier@1` - `none`, in `prettier@2` - `all` (prettier change default value to `es5` in next version), maybe be should discussion about this.

/cc @vjeux @azz

vjeux:
Trailing commas in PHP is a relatively recent addition (it was first added by Facebook in Hack and then copied by PHP if my memory serves me well) so it may not make sense to make it the default.

My gut feeling is that having no trailing comma the default and respect trailing-comma=all option would make sense. If that becomes a big issue in the future, then you can reevaluate.

czosel:
👍 sounds good!

czosel:
Awesome! We should slowly start thinking about documenting supported options :wink:

PR #79, opened 2018-02-20
feat: improve number output

evilebottnawi:
:star:


mgrip:
I'm honestly not sure this is something we would want to do? Does JS prettier do this?

(also if we start w/ an issue on the repo we can have this discussion before going through the effort of implementing 😉)

evilebottnawi:
@mgrip yes prettier do this, also we were just lucky that php and js have very similar processing of numbers, for other languages this should be very other, so I do not think that they will do this function shared. Also it is very pretty :+1:

mgrip:
👌 if JS prettier is doing the exact same thing I guess I'm ok with it

PR #83, opened 2018-02-20
tests: cast

evilebottnawi:
Parser problem:
https://github.com/glayzzle/php-parser/issues/119

In future we can discussion about `int` vs `integer` and etc.

### DIFF
--- a/tests/cast/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/cast/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,35 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`cast.php 1`] = `
+<?php
+
+$bar = (int) $foo;
+$bar = (integer) $foo;
+$bar = (bool) $foo;
+$bar = (boolean) $foo;
+$bar = (float) $foo;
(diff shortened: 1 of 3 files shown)


czosel:
In the past we agreed to always use short forms, since PSR-12 seems to go that way - see https://github.com/prettier/plugin-php/pull/5#discussion_r158664155
That would also mean that we don't need any change in the parser. What do you think?

evilebottnawi:
@czosel we can't change logic in our code (we use short forms), but parser should be return all data, it is related to parser. What about `(float)` vs `(double)` vs `(real)`

czosel:
@evilebottnawi i meant just the short forms: Here i think we should always print `(int)`, also if it was `(integer)`.

evilebottnawi:
@czosel right now we print always `int`, but `float` when you use `double` or `real`, should we fix this in parser?

czosel:
Actually i think the argument in https://github.com/prettier/plugin-php/pull/5#discussion_r158664155 means that we should always print `float`, because the aliases `double` and `real` are not supported for type hinting. I'd say it makes sense to have one consistent form for casts and type hints.

evilebottnawi:
@czosel agree :+1:

PR #93, opened 2018-02-21
tests: exit

evilebottnawi:
Separate https://github.com/prettier/plugin-php/pull/73

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -885,6 +885,8 @@ function printStatement(path, options, print) {
         ])
       );
     case "exit":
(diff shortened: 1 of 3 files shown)


src/printer.js
```diff
@@ -885,6 +885,8 @@ function printStatement(path, options, print) {
         ])
       );
     case "exit":
+      // Todo use `node.useDie` for determining `exit` or `die`
+      // after https://github.com/glayzzle/php-parser/commit/b22ea4863be9125c381951dc008820b68fc3d135 shipped
```
mgrip: ❤️

PR #94, opened 2018-02-21
fix: don't output extra newline with single open php tag

evilebottnawi:


czosel:
Is this specified by a styleguide?

evilebottnawi:
@czosel it is ugly, right now we have  (two empty newline)
```php
$a = 1;


```

After (one empty newline), prettier always output one empty line in end of file:
```php
$a = 1;

```

czosel:
Ah, that makes sense - thanks for the clarification! :+1:

src/printer.js
```diff
@@ -368,10 +368,14 @@ function printStatement(path, options, print) {
         return concat([
           "<?php",
           hardline,
-          path.call(childrenPath => {
-            return printLines(childrenPath, options, print);
-          }, "children"),
-          hardline
```
czosel: Why not just do it with a one-line change?
```
...,
path.call(childrenPath => {
  return printLines(childrenPath, options, print);
}, "children"),
node.children.length ? hardline : ''
```

evilebottnawi: @czosel avoid extra `call`, improve perf

czosel: Hmm, that sounds like a microoptimization to me - the effect is (a) probably very small and (b) only present when you're formatting an empty php file. I'd prefer to optimize for readability.

evilebottnawi: @czosel let's do this

evilebottnawi:
@czosel done

src/printer.js
```diff
@@ -368,10 +368,12 @@ function printStatement(path, options, print) {
...
+          concat([
```
czosel: I think you don't even need the extra call to `concat`

evilebottnawi:
@czosel done :+1:

PR #95, opened 2018-02-21
[WIP] tests: bom

evilebottnawi:
Should we remove BOM?

/cc @vjeux 

In PSR https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-1-basic-coding-standard.md#1-overview

Problem in prettier: https://github.com/prettier/prettier/issues/4050


vjeux:
I would mirror what prettier js does in this case, we have special handling for BOM (but I didn’t write it so I’m not sure what we did exactly)

evilebottnawi:
@vjeux prettier don't remove BOM, but PSR disallow using BOM, we can solve this inside our `parser.js`. Question: should we remove or not :confused:

vjeux:
Then get rid of it. I don’t really care either way :)

czosel:
I'll close this since it has become stale.

PR #97, opened 2018-02-21
fix: newline before trait body brace

evilebottnawi:


tests/traits/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -41,14 +42,19 @@ trait A {
   }
 }
 
-trait B extends otherTrait {
+trait B
+  extends otherTrait
```
czosel: Shouldn't the line break come after `extends`?

evilebottnawi: @czosel seems no, good catch :+1:

evilebottnawi:
@czosel done :+1:

PR #98, opened 2018-02-21
fix: newline before interface body brace

evilebottnawi:


tests/interface/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -13,12 +13,15 @@ interface test2 extends testInterface {
...
-interface testInterface {
+interface testInterface
+{
   public function setVariable($test);
   public function method(array $array): iterable;
 }
 
-interface test2 extends testInterface {
+interface test2
+  extends testInterface
```
czosel: Same issue as with the traits

evilebottnawi:
@czosel done

PR #102, opened 2018-02-21
fix: semicolon after new

evilebottnawi:


mgrip:
I think tests broke in another MR - gonna merge this and fix

PR #100, opened 2018-02-21
Feat newline after usegroup

evilebottnawi:
Also refactor newline after `namespace`

PR #103, opened 2018-02-21
feat: newline before file level doc block

evilebottnawi:
Ref: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#3-declare-statements-namespace-and-import-statements

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -77,11 +77,24 @@ function isLastStatement(path) {
   return body && body[body.length - 1] === node;
 }
 
+function isFileLevelDocBlock(path) {
+  const parent = path.getParentNode();
+  if (!parent || parent.kind !== "program") {
+    return false;
+  }
+  const node = path.getValue();
+  const body = parent.children;
+  return node.kind === "doc" && node.isDoc && body && body.indexOf(node) === 0;
+}
+
 function printLines(path, options, print) {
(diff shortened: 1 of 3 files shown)


mgrip:
all of these PR's ( #97, #98, #100) seem a little hacky to me - all trying to accomplish the spacing standards you linked above. I'm honestly not 100% sold that this is something thats totally worth the time - but if we do decide to implement it I feel like there might be a better way to go about it. Maybe 1 function to handle the spacing of the header in a file? it could just inspect the top level ast and look for what components exist out of this list and add spaces where necessary or something

> The header of a PHP file may consist of a number of different blocks. If present, each of the blocks below MUST be separated by a single blank line, and MUST NOT contain a blank line. Each block MUST be in the order listed below, although blocks that are not relevant may be omitted.
> 
> - Opening <?php tag.
> - File-level docblock.
> - One or more declare statements.
> - The namespace declaration of the file.
> - One or more class-based use import statements.
> - One or more function-based use import statements.
> - One or more constant-based use import statements.
> - The remainder of the code in the file.

evilebottnawi:
@mgrip let's do this as is, add when we can move logic to function as `shouldNeedSemicolon and etc`

evilebottnawi:
Let's do it in other PR

PR #104, opened 2018-02-21
feat: inline

evilebottnawi:
Feedback welcome :+1: 

Bug: https://github.com/glayzzle/php-parser/issues/120

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -40,6 +40,23 @@ function getLast(arr) {
   return null;
 }
 
+function getParentNodeListProperty(path) {
+  const parent = path.getParentNode();
+  if (!parent) {
+    return null;
+  }
+  const body = parent.children || parent.body || parent.adaptations;
+  return Array.isArray(body) ? body : null;
+}
+
+function getNodeIndex(path) {
+  const body = getParentNodeListProperty(path);
+  if (!body) {
+    return -1;
+  }
+  return body.indexOf(path.getValue());
+}
+
 function genericPrint(path, options, print) {
   const n = path.getValue();
   if (!n) {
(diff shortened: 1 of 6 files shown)


src/printer.js
```diff
@@ -244,8 +244,25 @@ function printExpression(path, options, print) {
           ),
           getEncapsedQuotes(node, { opening: false })
         ]);
-      case "inline":
-        return node.value;
+      case "inline": {
+        const parentNode = path.getParentNode();
+        const isFirstNode =
```
mgrip: maybe helpful to add these `isFirstNode(path)` and `isLastNode(path)` to that util file you added?

mgrip:
Yessss this is awesome

evilebottnawi:
It is really not easy, but i find solution, today i will release first working version :+1:

src/printer.js
```diff
@@ -767,7 +801,7 @@ function printStatement(path, options, print) {
       case "parameter": {
         const name = concat([
           node.nullable ? "?" : "",
-          node.type ? path.call(print, "type") + " " : "",
+          node.type ? concat([path.call(print, "type"), " "]) : "",
```
evilebottnawi: I try 1 hour on catch and fix it :imp:  When we use `path.call` it can return not string, also `object`. Don't concatenate string (i.e. `+ "string"`) - use `concat([..args])` :smile:

evilebottnawi:
@mgrip @czosel ready for first release

evilebottnawi:
In some places ugly, but it is better than broken code. I think we can merge this as is. And after problem in `php` parser was fixed, i update tests.

src/printer.js
```diff
@@ -40,14 +40,55 @@ function getLast(arr) {
   return null;
 }
 
+function getParentNodeBody(path) {
```
mgrip: do you think these fns would be better suited in `util.js`?

mgrip: also seems more generic than just `body` - maybe name something like `getParentNodeListProperty`? Naming things is hard haha

src/printer.js
```diff
@@ -119,15 +159,15 @@ function printLines(path, options, print) {
     if (lineShouldEndWithSemicolon(stmtPath)) {
       parts.push(";");
     }
-    if (
-      sharedUtil.isNextLineEmpty(text, stmt, options) &&
-      !isLastStatement(stmtPath)
-    ) {
+    if (!isLastStatement(stmtPath)) {
```
mgrip: why split out?

src/printer.js
evilebottnawi: @mgrip i think all function what working with ast should be here. But we can discussion about this and move all other function to utils. Maybe best create issue and do this in other PR :+1:

src/printer.js
evilebottnawi: @mgrip it is more logic and for future to allow `<?php echo 'test; ?>` in one line/

evilebottnawi:
@mgrip `funciton` name changes :+1: After this PR in theory we don't break code, but i have a lot of project and code base, after merge this i run and fix other problem

mgrip:
🎉

Issue #107, opened 2018-02-22
Option for opening brace


mgrip:
1st off I want to be clear - I think the default should still be to follow PSR2-like formatting standards

With that in mind, from my own experience and now starting to look through other open-source php projects, it seems like one of the biggest differences in formatting standards is putting the opening brace for functions, classes, etc on a new line vs same line. I'm a little worried that not having this option might scare people off from using prettier/plugin-php.

I was thinking of starting to work on a PR for adding an option to do this (defaulting to the PSR2 format of putting it on a new line), but wanted to open for discussion first.

```php
openingBraceNewLine = true (default)

class Test
{
  public function testMethod()
  {
  }
}

function testFunction()
{
}

openingBraceNewLine = false


class Test {
  public function testMethod() {
  }
}

function testFunction() {
}
```

mgrip:
cc @evilebottnawi @czosel

czosel:
For me, that line break definitely is the ugliest aspect of PSR-2, so from a user perspective i'd support this 100%. From a maintainance perspective, I think it depends on the implementation complexity - if this boils down to a couple of ternarys printing hardlines or not (which it probably will), then I'd say let's go for it!
I guess we should leave the default to be PSR2-compatible, right?

mgrip:
Totally agree. I'm taking a look now and I think it should definitely be doable - I think it just boils down to classes, functions, methods, and traits - which all extend the `statement` type which should help. And yes definitely agree we should leave the default to be PSR2-compatible.

Ok cool I'll try and put together a PR and we can see how much complexity it adds

mgrip:
Added in #108

mgrip:
Reopening for more discussion - @evilebottnawi

evilebottnawi:
@mgrip thanks, be good if we revert option and keep PR open

mgrip:
I'll put up an MR to remove it - it's just removing 1 loc of code from the printer. How do you propose we get more insight/opinions?

evilebottnawi:
@mgrip twitter + issue
- :+1: Add
- :-1: Don't add

We have other many problem right now, new option should be added only after stable release

mgrip:
[...]

Issue #54, opened 2018-02-18
move brace to new line for functions


mgrip:
From https://github.com/prettier/plugin-php/issues/45#issuecomment-366111694
> This came to mind when seeing the output:
> 
> The opening brace of a function should be on its own line according to PSR-12 (same as class methods).
> 
> (but the opening brace of a closure goes on the same line as the closing parens of the parameter list)

aboyton:
I know we want to go with PSR-12 and we don't want to add options for everything, but man is that an ugly style.

mgrip:
This got fixed a while ago - please reopen if you see anything else!

PR #108, opened 2018-02-22
Creating option for moving opening declaration brace to new line

mgrip:
#107 #54

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,6 +3,7 @@
 const parse = require("./parser");
 const print = require("./printer");
 const clean = require("./clean");
+const options = require("./options");
 
 const languages = [
   {
@@ -54,5 +55,6 @@ const printers = {
 module.exports = {
   languages,
   printers,
-  parsers
+  parsers,
+  options
 };
--- a/src/options.js
+++ b/src/options.js
@@ -1,0 +1,10 @@
+"use strict";
+
+module.exports = {
+  openingBraceNewLine: {
+    type: "boolean",
+    category: "Global",
+    default: true,
+    description: "Move open brace for code blocks onto new line."
+  }
+};
--- a/src/printer.js
+++ b/src/printer.js
@@ -69,6 +69,12 @@ function lineShouldEndWithSemicolon(path) {
   if (node.kind === "method" && node.isAbstract) {
     return true;
   }
+  if (node.kind === "method") {
+    const parent = path.getParentNode();
+    if (parent && parent.kind === "interface") {
+      return true;
+    }
+  }
   return includes(semiColonWhitelist, node.kind);
 }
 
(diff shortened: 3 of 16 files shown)


mgrip:
So I mostly solved for #54 with this PR - one thing we still need to address though is _not_ breaking between `) {` if the function breaks. ie

```php
function reallyLongFunctionName(
  $argument,
  $argument2
)
{
```
should actually be
```php
function reallyLongFunctionName(
  $argument,
  $argument2
) {
```

It seems like thats gonna be complicated though so once this gets merged I'll open a new issue for it

mgrip:
🎉

evilebottnawi:
/cc @vjeux 

we need remove new option

evilebottnawi:
@mgrip @czosel don't do this, it is bad, very bad

mgrip:
why?

evilebottnawi:
@mgrip no option, it is prettier, please read https://prettier.io/docs/en/option-philosophy.html

mgrip:
In my mind this is similar to

> --jsx-bracket-same-line was needed for a big company with a huge code base (Facebook), which backed the project when it got started, to be able to adopt Prettier at all.

I did post an issue on this first and the only feedback on it was 2 👍 . I think my argument from the original issue posting still stands, but happy to continue that discussion if you'd l like

evilebottnawi:
@mgrip let's revert option and open issue with vote

mgrip:
I'd rather not revert it - the actual option is 1 loc. If we really think we should remove it I will remove that loc but I'd like some other opinions first

czosel:
I think this example from prettier for JS fits even better with our situation:
> --arrow-parens was added after huge demand. Prettier has to strike a balance between ideal goals and listening to the community.

evilebottnawi:
@mgrip @czosel we should have discussion about option, no merge without issue and vote. Each option increase complex and add new problem in future

mgrip:
There was an issue, I got no opposition to it before it was merged - I just reopened it if you want to continue the discussion there #107

evilebottnawi:
@mgrip we merge this without my vote :trollface:

mgrip:
Tbh I didn't think this would be controversial - for reasons stated above and in the issue

evilebottnawi:
@mgrip @czosel  I'm not against the option or not specifically this option, the only problem is that adding one option other will want more options, and each will complicate the maintenance of the project and deliver only problems, our goal is to make the code beautiful, not optional, for this we have linter
In js `prettier + easlint`
In css `prettier + stylelint`
In php `prettier + PHP_CodeSniffer`

mgrip:
I definitely agree - it just seems like this is probably the largest difference between formatting standards across php code. Like I mentioned originally I'm worried not having this option will prevent people from using prettier for php

mgrip:
Also lets keep discussion on the open issue pls 😄 https://github.com/prettier/plugin-php/issues/107

evilebottnawi:
@mgrip we need revert :smile:

nonoroazoro:
This is why I'll never use prettier.

PR #115, opened 2018-02-23
Remove open brace option

mgrip:
Temporarily removing until we get more feedback on https://github.com/prettier/plugin-php/issues/107


mgrip:
👍 Now lets go get more feedback! haha

PR #123, opened 2018-02-25
Function call wrapping

czosel:


tests/functions/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -170,4 +190,27 @@ function gen(): iterable
...
+  $evenLongerParameter
+]);
+
+$app->get('/hello/{name}', function ($name) use ($app) {
+  return 'Hello ' . $app->escape($name);
+});
+
+$this->something->method($argument, $this->more->stuff($this->even->more->
+  things->
+  complicatedMethod()));
```
czosel: I tried using a `conditionalGroup` to wrap function calls just like the JS printer does - which works generally quite well, but I'm still having issues with this nested call example. Maybe @azz can point me in the right direction?

mgrip: @czosel do you know where the JS printer does this? I'm mostly just curious

czosel: @mgrip https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L3195

mgrip: 🙏

src/printer.js
```diff
@@ -108,6 +116,30 @@ function printLines(path, options, print) {
   return join(hardline, printed);
 }
 
+function printArgumentsList(path, options, print) {
```
mgrip: mind just adding a comment outlining the overall strategy here?

tests/functions/__snapshots__/jsfmt.spec.js.snap
mgrip: JS actually just does the long form in that example. I'm beginning to wonder if we're actually going to be able to print everything in the way PSR12 outlines

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMAWBLAzgOixAWznQygHMcj0IATACgEMAnMgVyNgBoACE3AiEzh4YrAGZi6AHSjdemXHABuCSoOElyuSAQAOAGwxgG8GgFliaWnQCUMmzYDcITiAi6YGaFmShmTCAB3AAVmBB8UBiUIDBoXEAYsGGQxBn0sOFcAIyYGMABrYgBlXTzSMmQYJlZMkBoIMBS0jNdSDKYYYNyyAgYm9NqAKywADwAhXILihiIAGVI4fpa3VhhdVYAmJdrSpnbkECyGLIBPfWh43SZSGAB1WPRkAA4ABlcriAzb3N0Dq7h2ip4kIAI6sDBCLoMHp9JCpAauDIEDCVaq1LDlfRwACKrAg8G2rhgx3uNEeSA2RNyGEM5AAwoRegcoNBFq5WBkACrHCLwjIAX35QA

czosel:
Removing `mediumForm` actually also yields quite good results. I'll now look into using `mediumForm` only for arrayExpressions. If you want we can merge this state already, since it's already much better than the current one.

czosel:
@mgrip @evilebottnawi I think this is ready for review now 🎉

mgrip:
This is great! Nice work @czosel 💯

harmenjanssen:
Hi! Sorry to be leaving a comment on such an old PR, but I started using this plugin today, and noticed the following format being rendered by Prettier:

```php
array_map(function ($entry) {
    return $entry * 2;
}, $arr);
```

Through the snapshots I was able to find PR https://github.com/prettier/plugin-php/pull/23, in which a change is pushed that changes the above to this:

```php
array_map(
  function ($entry) {
    return $entry * 2;
  },
  $arr
);
```

However, about a month after that PR @czosel, the author of both PRs, pushes the changes in _this PR_ reverting that beautiful format to the old.

I'm wondering: is this deliberate? If not, would you accept a PR that changes the format to that in the second snippet.
I can take a stab at it, but maybe there's a reasoning why the above snippet is preferred.

czosel:
Hi @harmenjanssen!

I think this was not deliberate. In #23 we optimized the printing of long function call expressions, but made the printing of expressions like the one you posted above a little worse, which was an acceptable trade-off at the time. If you find a clever way to optimize the example above that also works for long call expressions, I'd be happy to merge your PR :+1:

PR #124, opened 2018-02-26
Change tab width to 4

czosel:
This time i'll wait with updating snapshots to avoid conflicts


czosel:
This should be ready now :tada:

evilebottnawi:
@czosel great!

Issue #134, opened 2018-02-26
Strip loc when running `--debug-check`


mgrip:
When running prettier with the `--debug-check` option (for testing), I get a whole lot of ast differences around [node locations](https://github.com/glayzzle/php-parser/blob/master/docs/AST.md#location). I _think_ all this means is that we have to strip out node locations in our ast-clean function? https://github.com/prettier/plugin-php/blob/master/src/clean.js#L5

mgrip:
Actually looks like it should be happening here - https://github.com/prettier/plugin-php/blob/master/tests_config/run_spec.js#L84. Not sure why thats not working

mgrip:
Fixed in #135

PR #135, opened 2018-02-26
fix: clean-ast function

mgrip:
https://github.com/prettier/plugin-php/issues/134


czosel:
Out of curiosity: How does the second argument to `clean` work? What does it represent?

mgrip:
It's a copy of the original node, that core prettier will use when comparing ast's - [here's where the function gets called](https://github.com/prettier/prettier/blob/84c603623da590487fa13bc6ae3fce4e80f465af/src/common/clean-ast.js#L50). And here's an example in other language plugins ([css](https://github.com/prettier/prettier/blob/84c603623da590487fa13bc6ae3fce4e80f465af/src/language-css/clean.js#L3))

PR #136, opened 2018-03-01
fix: strings

evilebottnawi:
All buggy things are commented.

Feel free to feedback.

Issue about syntax: https://github.com/prettier/plugin-php/issues/138.

Let's review this and merge as is. It is maximum how we can handle strings right now.


tests/string/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,124 @@
...
+// $string = "This is the value of the var named by the return value of \\$object->getName(): {\${$object->getName()}}";
+$string = "I'd like an {\${beers::softdrink}}\\n";
+$string = "I'd like an {\${beers::$ale}}\\n";
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+$string = 'this is a simple string';
+// $string = 'You can also have embedded\\n newlines in
+// strings this way as it is
+// okay to do';
+$string = 'Arnold once said: "I\\'ll be back"';
```
mgrip: Same thing - i think an extra `\` got added

tests/string/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,124 @@
...
+$string = "This is the value of the var named $name: {\${$name}}";
+$string = "This is the value of the var named by the return value of getName(): {\${getName()}}";
+$string = "This is the value of the var named by the return value of getName(): {\${getName($arg1, $arg2, $arg3)}}";
+// $string = "This is the value of the var named by the return value of \\$object->getName(): {\${$object->getName()}}";
+$string = "I'd like an {\${beers::softdrink}}\\n";
+$string = "I'd like an {\${beers::$ale}}\\n";
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+$string = 'this is a simple string';
+// $string = 'You can also have embedded\\n newlines in
```
mgrip: this should remain `'You can also have embedded\n newlines in`, right?

tests/string/__snapshots__/jsfmt.spec.js.snap
mgrip: maybe add a test `$string = 'This is an escaped quote \' test';` - I'm curious if that works

tests/string/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: @mgrip yes

tests/string/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: @mgrip yep let's add test, but here all fine with `\`, jest always escaped special characters

mgrip: just realized that - thanks! looks good to me!

Issue #139, opened 2018-03-01
Multi-line conditions


mgrip:
input:
```php
if (reallyReallyReallyLongFunction() && $this->reallyreallyreallyreallyLongMethodName()) {
  return true;
}

if ($this || $that && $theOtherThing && $someReallyReallyReallyLongBooleanVariable) {
  return true;
}
```
output:
```php
if (
reallyReallyReallyLongFunction() && $this->
    reallyreallyreallyreallyLongMethodName()
) {
    return true;
}

if (
$this || $that && $theOtherThing && $someReallyReallyReallyLongBooleanVariable
) {
    return true;
}
```

PSR12 [specifies](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#51-if-elseif-else):

> Expressions in parentheses MAY be split across multiple lines, where each subsequent line is indented at least once. When doing so, the first condition MUST be on the next line. The closing parenthesis and opening brace MUST be placed together on their own line with one space between them. Boolean operators between conditions MUST always be at the beginning or at the end of the line, not a mix of both.

mgrip:
interestingly prettier doesn't necessarily put each case on its own line if smaller ones fit https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAZgAgBQCc4CGANkQJ4BKhJFVZAMtAOYBiArlGDKtNgJSYAyAZgAkMABaoAzgDp8xMvOpLFtUgyiMAsnAkQAJgDkCAWzh9+wADpRMmfDFa5bMXKzgBuGwF8bNjDhiklKYAD6hohIEMILCQXAA8hJwuAAqkpqxolIQZpQKNAX51BqMAEIQEESEUABqBLioBABG1ZY2dg5OLm6ePiAANCAQAA5c0FLIoA24EADuAAoNCJMoBABuEKj6gyAEUjDI6MRScEPNuARgANa6AMojV6iayK7uQ-oQYEcnZ2hQp1wMAWl0YJgIPyIpyGACspAAPMqXG73UxwOjPOCQ6HDVgwEZ4gBM2L+j1wgOQIGaLVIRGguxGjVgAHVthJkAAOAAMQ0ZEFOzMuI0pjLggPWWKG+AAjqxUPgQQQwRCkMcoX9TiZUK9ekMpM9GNUAIqsCDwElDGAtVn6dlIQmWy6oIgGgDCuXBlKg0ElIFYp1SLVWatO3m8QA

[...]

PR #140, opened 2018-03-02
fix: multi-line conditions

mgrip:
https://github.com/prettier/plugin-php/issues/139

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -191,14 +191,50 @@ function printExpression(path, options, print) {
         return concat([node.type + node.type, path.call(print, "what")]);
       case "post":
         return concat([path.call(print, "what"), node.type + node.type]);
-      case "bin":
-        return concat([
-          path.call(print, "left"),
-          " ",
-          node.type,
-          " ",
-          path.call(print, "right")
-        ]);
+      case "bin": {
+        // general idea here is that we are continually checking nested
+        // binary expressions to see if we should be starting a new group
+        // or not (based on operator precedence)
+        const printBinaryExpression = (
+          path,
+          print,
+          options,
+          isNested = false
+        ) => {
+          const node = path.getValue();
+          if (node.kind !== "bin") {
+            return path.call(print);
+          }
+
+          const shouldGroupLeft = !(
+            node.left.kind === "bin" &&
+            util.shouldFlatten(node.type, node.left.type)
+          );
+          const printedLeft = path.call(
+            left => printBinaryExpression(left, print, options, true),
+            "left"
+          );
+
+          const shouldGroupRight = !(
+            node.right.kind === "bin" &&
+            util.shouldFlatten(node.type, node.right.type)
+          );
+          const printedRight = path.call(
+            right => printBinaryExpression(right, print, options, true),
+            "right"
+          );
+
+          const printedExpression = concat([
+            shouldGroupLeft ? group(printedLeft) : printedLeft,
+            " ",
+            node.type,
+            line,
+            shouldGroupRight ? group(printedRight) : printedRight
+          ]);
+          return isNested ? printedExpression : group(printedExpression);
+        };
+        return printBinaryExpression(path, print, options);
+      }
       case "parenthesis":
         return group(
           concat(["(", softline, path.call(print, "inner"), softline, ")"])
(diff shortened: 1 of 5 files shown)


czosel:
This looks awesome! I have two comments:

* Is the JS printer doing the same thing?
* Should we try to add some more tests for all the cases where `shouldFlatten` returns false?

evilebottnawi:
Tests always welcome :+1:

mgrip:
Yup! I stole the general idea from the JS printer - https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L4706. I'll add some more tests - feel free to add more after this gets merged too! I'm guessing this won't cover edge case but I think its a good starting point that we can start to open up more specific issues

mgrip:
new test cases are a little strange but thats how [JS prints them](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAZgAgBQBIYAWqAzpgD5mb4ECGMmAZA1YXAPKsBOAKkVAOaYAPFWIQAtnABKcGgBs5ATxnylKhYoAy0fgCEIEObKgA1Gp1Q0ARkYCUmYAB0omTJzgwArpxcxOnuABuZwBfZ2cMHGoSckpqOkZmanYuXlQBYVEJaVkNdTVcpW0BfUNjMwtrI0wAXhqqOABHT3kAYQI4MABreycXNw9vX38g0JAAGhAIAAcYVGhiZFBzTggAdwAFcwRFlBoANwhUABMJkBpiGGR0eWI4SatOGm6PAGVp5-T+ZD8AyeOIGBrrd7mgoHdODANk9+OIaMC5HdJgArYgAD10TxeMFeNEkmnScARSKmnhg0zJACZiaCPpwIcgQFZrIo5NAztMLLAAOonQjIAAcAAZJpyIHduU9pozOXAIfsiZN3M1UO5oTRYfCkDdEaC7uJUD8RpNiF8jABFTwQeA0yYway8478pCUu1PVByL6tCRwxlQaCKkCeO7cay7HV3EIhIA)

evilebottnawi:
@mgrip let's add tests around `(` and `)`. Example:
```php
if (($this || $that) && $theOtherThing < $someReallyReallyReallyLongBooleanVariable == $equalCheck) {
  return true;
}
```

And we can merge :+1:

mgrip:
confirmed [same as JS](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAZgAgBTYCQwAWqAzpgD7mYGECGMAlJgGTPVFwDyHATgCrEoAc0wAeaiQgBbOACU4tADaKAnvKWr1ylQBloQgEIQIihVABqtHqloAjU5gC8j6nACOAVyUBhQnDAA1kzAADpQmJg8cDAePOEwPB5wANxhAL4gADQgEAAOMKjQJMigVjwQAO4AClYIxSi0AG4QqAAmWSC0JDDI6EokcNm2PLSB0QDKuaOowsgJSdmtEGC9-YNoUAM8MFUjQlK0q4oD2QBWJAAeBiNjMOO0MjozcEcnOR4wuR8ATK-rUzwtsgQLY7CpFNAOrlrLAAOptIjIAAcAAZstCIANYSNcsDoXAto0Xtkop5UFFdrR9ockH1jusBlJUHNEgyZkJTABFDwQeB-bIwOzw1qIpDfAUjVCKdneaQHYFQaDEkAeAZ8Oz1OkDNJpIA) 👍

czosel:
:100:

upwebdesign:
This is not valid PSR-2 according to `phpcs`. 

```php
if (
    $e instanceof NotFoundHttpException ||
    $e instanceof ModelNotFoundException ||
    $e instanceof RelationNotFoundException
) {
    return response()->json(['message' => 'Resource not found :: ' . $e->getMessage()], 404);
}
```

It should be:

```php
if ($e instanceof NotFoundHttpException ||
    $e instanceof ModelNotFoundException ||
    $e instanceof RelationNotFoundException
) {
    return response()->json(['message' => 'Resource not found :: ' . $e->getMessage()], 404);
}
```

Thoughts?

nicoder:
see section 5.1 of [PSR-12](https://www.php-fig.org/psr/psr-12/) :

> Expressions in parentheses MAY be split across multiple lines, where each subsequent line is indented at least once. When doing so, the first condition MUST be on the next line.

upwebdesign:
@nicoder thanks for the clarification. I was under the assumption that PSR-2 was being used. I updated PHPCS and will see how it goes.

PR #145, opened 2018-03-02
fix: `semicolon` in control structure without brackets

### LINKED ISSUE #141: missing semicolon when transforming `if` without brackets
if (rand() > 0.5) return true;

becomes

    if (rand() > 0.5) {
        return true
    }

evilebottnawi:
fixes #141 

it is not solve in `printLines` because it is part of statement


evilebottnawi:
What a problem with travis today :angry:

src/clean.js
```diff
@@ -51,6 +51,22 @@ function clean(node, newObj) {
   if (["Location", "Position"].includes(node.constructor.name)) {
     newObj = "locationNode";
   }
+
+  if (["foreach", "for", "if", "while", "do"].indexOf(node.kind) !== -1) {
```
nicoder: instead of `.indexOf` we could use `.includes`, like in the previous `if` statement?

ichiriac: I'm not sure about nodejs internal optimisations, but in principle the fact to build an array at each time the if statement is executed should cost more than defining a const array :

```js
// in header, outside function
const statements = ["foreach", "for", "if", "while", "do"];
// into a function/loop repeated many times 
if (statements.includes(node.kind) {
```

nicoder: it would not surprise me if v8 optimised this away (it should be able to see that the array does not change), but I have no idea either.

evilebottnawi: let's do it thanks :+1:

PR #146, opened 2018-03-02
tests: long `foreach`

evilebottnawi:
@czosel @mgrip @vjeux @azz That do you think about this output? Should we print `as` always newline when group breaks?

### DIFF
--- a/tests/foreach/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/foreach/__snapshots__/jsfmt.spec.js.snap
@@ -20,6 +20,9 @@ foreach ( $test as $i ) {
 
 foreach ($test as $i)
     $test2 = $i;
+
+foreach (['foo' => ['very-very-very-long-value'], 'bar' => ['very-very-very-long-value']] as $veryVeryVeryVeryVeryVeryVeryVeryLongKey) {}
+foreach (['foo' => ['very-very-very-long-value'], 'bar' => ['very-very-very-long-value']] as $veryVeryVeryVeryVeryVeryVeryVeryLongKey => $veryVeryVeryVeryVeryVeryVeryVeryLongValue) {}
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
 foreach ($test as $i) {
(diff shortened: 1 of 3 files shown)


tests/foreach/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -38,4 +54,16 @@ foreach ($test as $i) {
     $test2 = $i;
 }
 
+foreach (
+    [
+        'foo' => ['very-very-very-long-value'],
+        'bar' => ['very-very-very-long-value']
+    ] as
+    $key => $aliases
+) {
```
vjeux: This is a tricky one. The way it is printed above wouldn't pass my "a human would have written it like this" test. I can see the following two as acceptable:

```js
foreach (
    [
        'foo' => ['very-very-very-long-value'],
        'bar' => ['very-very-very-long-value']
    ] as $key => $aliases
) {
```

```js
foreach (
    [
        'foo' => ['very-very-very-long-value'],
        'bar' => ['very-very-very-long-value']
    ]
    as $key => $aliases
) {
```

Not sure how hard it is going to be to handle all the cases though, I would recommend getting more concrete use cases to see all the possible things that people write in practice.

evilebottnawi: @vjeux how about long variable names? For me looks good:
```php
foreach (
    [
        'foo' => ['very-very-very-long-value'],
        'bar' => ['very-very-very-long-value']
    ]
    as 
    $keylonglonglolonglonglonglonglonglonglonglonglong => 
        $keylonglonglolonglonglonglonglonglonglonglonglong
) {
```

vjeux: Yeah, this looks good to me. The following would also look acceptable. We sometime ignore 80 columns limitation when there's no obvious breaks with prettier.

```js
foreach (
    [
        'foo' => ['very-very-very-long-value'],
        'bar' => ['very-very-very-long-value']
    ]
    as $keylonglonglolonglonglonglonglonglonglonglonglong => 
        $keylonglonglolonglonglonglonglonglonglonglonglong
) {
```

evilebottnawi: @vjeux i agree, limitation don't impose code to format so that it is impossible to read

czosel: @evilebottnawi I'd also be in favor of removing the linebreak after `as`.

evilebottnawi: @czosel :+1:

PR #147, opened 2018-03-02
fix: empty `for`

evilebottnawi:
Based on `prettier` https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L1481


tests/for/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -14,13 +14,28 @@ for ($i = 1, $j = 0; $i <= 10; $j += $i, print $i, $i++);
 
 for ($really_really_long_variable_name = 0, $and_something_else = 0; $really_really_long_variable_name <= 5; $really_really_long_variable_name++, print $really_really_long_variable_name, $even_more_stuff++);
 
-for (;;):
+for ($i = 0; $i <= 5; $i++):
```
evilebottnawi: Just move tests below

PR #156, opened 2018-03-05
chore: move prettier in `peerDependencies`

evilebottnawi:
@azz @vjeux I think we should use `prettier` as `peerDependencies`, it is allow update `prettier` without upgrading plugin (cli/api/options fixes and features). `ESlint`, `stylelint`, `webpack` plugins/loaders and other pluggable systems use `peerDependencies`. 

But in `prettier` we don't have `prettier/src/common/clean-ast` and some other utils/functions, maybe create bundle for this?


package.json
```diff
@@ -10,18 +10,21 @@
...
     "eslint-config-prettier": "2.9.0",
     "eslint-plugin-import": "2.6.1",
     "eslint-plugin-jest": "21.5.0",
     "eslint-plugin-prettier": "2.4.0",
+    "prettier": "1.11.1",
     "jest": "21.1.0",
     "jest-runner-eslint": "0.3.0"
   },
+  "peerDependencies": {
+    "prettier": "^1.11.1"
```
evilebottnawi: Maybe better only `patch` version :question:

evilebottnawi:
@vjeux i don't familiar with internal prettier api, maybe we can to ping somebody and she/he help us with this :smile:

vjeux:
@azz? :)

azz:
For cleanAST we could add it to `__debug`:  https://github.com/prettier/prettier/blob/master/index.js#L430

evilebottnawi:
@azz will be great :+1:

czosel:
Alright, given that it would be nice that this lands before our first release: What is still needed? Do these two functions still have to be added to `__debug`?
```
tests_config/run_spec.js:const { massageAST } = require("prettier/src/common/clean-ast");
tests_config/run_spec.js:const { normalize } = require("prettier/src/main/options");
```

evilebottnawi:
@czosel yep, only this :smile:

czosel:
Ok, does this still need to be fixed in prettier itself?

vjeux:
clean-ast is only used for tests fwiw

czosel:
Right, and the same holds for `normalize`. That means we could just ignore it, right?

czosel:
Looks like tests are passing now :tada: @mgrip

PR #157, opened 2018-03-06
tests: long function with type declaration

evilebottnawi:
From PSR-12: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#45-method-and-function-arguments

### DIFF
--- a/tests/class/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/class/__snapshots__/jsfmt.spec.js.snap
@@ -178,6 +178,14 @@ abstract class ReallyReallyReallyLongClassName extends AbstractModelFactoryResou
   {
     return new SplQueue();
   }
+
+    public function longLongAnotherFunction(
+        string $foo,
+        string $bar,
+        int $baz
+    ): string {
+        return 'foo';
+    }
 }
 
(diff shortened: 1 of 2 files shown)


tests/class/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -341,6 +349,15 @@ abstract class ReallyReallyReallyLongClassName
     {
         return new SplQueue();
     }
+
+    public function longLongAnotherFunction(
+        string $foo,
+        string $bar,
+        int $baz
+    ): string
+    {
```
czosel: This is acutally a bug, isn't it?

evilebottnawi: @czosel yep, we have todo around this, let's do this in other PR?

czosel: Alright! Can you open an issue for this? (Or does one exist already?)

evilebottnawi: @czosel Done https://github.com/prettier/plugin-php/issues/171 :+1:

czosel: Thanks :)

PR #158, opened 2018-03-06
fix: indent in long `switch` test

evilebottnawi:
PSR-12: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#52-switch-case

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1040,8 +1040,7 @@ function printStatement(path, options, print) {
         group(
           concat([
             "switch (",
-            softline,
-            path.call(print, "test"),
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Also i think we should open issue about option for PSR compatibility:
```php
if (
    $var1 && 
    $var2
)
```
vs
```php
if (
    $var1 
    && $var2
)
```

czosel:
>  Boolean operators between conditions MUST always be at the beginning or at the end of the line, not a mix of both.

That means we're compatible right now, aren't we?

evilebottnawi:
@czosel right now no, but in near future we will :smile:

PR #165, opened 2018-03-06
refactor: tests

evilebottnawi:
@czosel Let's move all `call` tests in `call` directory. `function` for `function` declaration


czosel:
:+1: Good idea

PR #169, opened 2018-03-06
tests: spaces in function and method declaration

evilebottnawi:
PSR-12:

> Method and function names MUST NOT be declared with a space after the method name. The opening brace MUST go on its own line, and the closing brace MUST go on the next line following the body. There MUST NOT be a space after the opening parenthesis, and there MUST NOT be a space before the closing parenthesis.

### DIFF
--- a/tests/class/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/class/__snapshots__/jsfmt.spec.js.snap
@@ -199,6 +199,11 @@ implements \\ArrayAccess, \\Countable
 }
 
 class FooBar { public $property; public $property2; public function method() {} public function method2() {} }
+
+class FooBarFoo
+{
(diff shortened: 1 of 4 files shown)


tests/class/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -355,4 +360,12 @@ class A
 
 $someVar = new ReaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaalyLongClassName();
 
+class FooBarFoo
+{
+    public function fooBarBaz($arg1, &$arg2, $arg3 = [])
+    {
+        // Comment
```
evilebottnawi: Problem with comments, but not related to this PR

Issue #152, opened 2018-03-04
String concatenation


czosel:
Input:
```php
            $this->view->textareaContentId = 'a' . $activationId . 'n' . $noticeTypeId;
```
Output:
```php
            $this->view->textareaContentId = 'a' .
            $activationId .
            'n' .
            $noticeTypeId;
```
I'd suggest
```php
            $this->view->textareaContentId = 
                 'a' . $activationId . 'n' . $noticeTypeId;
```
or (for longer cases)
```php
            $this->view->textareaContentId = 'a' .
                $activationId .
                'n' .
                $noticeTypeId;
```

czosel:
It seems that this needs to be fixed inside `printBinaryExpression` - @mgrip, could you take a look?

mgrip:
Will do!

mgrip:
So I think this is actually related to how we print `assign` nodes - here's how the JS printer handles it: https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L4807

mgrip:
Fixed in #177

PR #177, opened 2018-03-07
fix: breaking assignments for binary expressions onto newline

mgrip:
#152


mgrip:
If I had to guess there are probably more edge cases we'll run into (the JS one looks more complicated haha) - but I think this covers the basics for now!

Issue #176, opened 2018-03-07
Escape forward slash


mgrip:
input:
```php
$testString = '\\';
```
output:
```php
$testString = '\';
```

nicoder:
'forward' slash?

mgrip:
fixed in #180

PR #180, opened 2018-03-08
fix: string escape for fwd slash

mgrip:
#176


tests/string/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -85,14 +87,14 @@ $string = "foo\\\\\\\\bar";
...
 $string = "foo'bar";
-$string = "foo\\377bar";
-$string = "foo\\3";
-$string = "foo\\xFFbar";
-$string = "foo\\xr";
-$string = "foo\\x";
-$string = "foo\\u{ff}bar";
-$string = "foo\\ubar";
-$string = "foo\\u";
+$string = "foo\\\\377bar";
```
evilebottnawi: Broke :disappointed:

mgrip: Yup my bad, didn't realize what these were - just put back!

tests/string/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -57,15 +57,17 @@ $string = "This is the value of the var named by the return value of getName():
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
 $string = 'this is a simple string';
 // $string = 'You can also have embedded\\n newlines in
 // strings this way as it is
 // okay to do';
 $string = 'Arnold once said: "I\\'ll be back"';
-$string = 'You deleted C:\\*.*?';
-$string = 'You deleted C:\\*.*?';
+$string = 'You deleted C:\\\\*.*?';
```
evilebottnawi: How about `\\\'`?

tests/string/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: @mgrip do not need to admit mistakes, it's okay search better solution

tests/string/__snapshots__/jsfmt.spec.js.snap
mgrip: Ya im confused - the test is

```php
$string = 'You deleted C:\\*.*?';
$string = 'You deleted C:\*.*?';
```

So I'm not sure why these are the same (in either case)... The snapshot escaping just makes this so much more confusing haha

evilebottnawi: and `\`, `\\`, `\\\` just for verification tests

mgrip: will do 👍

evilebottnawi: $string = 'You deleted C:\\*.*?'; - just two slash
$string = 'You deleted C:\*.*?'; - just one slash
In snapshot each slash should be escaped

mgrip: err actually `\` and `\\\` are invalid because they escape the closing quote

evilebottnawi: @mgrip when I think that something is wrong or wrong, i just create `test.php` with content and `echo $string;` and look how it is output :smile:

mgrip: right right - so snapshot should be
```php
$string = 'You deleted C:\\\\*.*?';
$string = 'You deleted C:\\*.*?';
```
right?

I'm just not sure why its incorrect atm. They're both the same

evilebottnawi: @mgrip this examples was got from `php.net` :smile:

evilebottnawi: yep it is right, first with two slashes, second with one slash

mgrip: Ok its a parser error - I'll open a new issue

image

mgrip:
@evilebottnawi ignoring the parser issue I think this should be good now!

Issue #112, opened 2018-02-23
Remove brackets when composing attribute/methods calls


mathieutu:
```diff
-        return $this->quantity * $this->{self::PRICE_ATTRIBUTE};
+        return $this->quantity * $this->self::PRICE_ATTRIBUTE;
```

czosel:
Also:
```php
$this->view->bar(
    $this->view->form->{$data['data-some-looooooooooooooooooooong-reference']}->getValue()
);
```
becomes
```php
$this->view->bar(
    $this->view->form->$data['data-some-looooooooooooooooooooong-reference'
    ]->getValue()
);
```

czosel:
Fixed in #183

PR #183, opened 2018-03-08
Fix dynamic propertylookup

czosel:
There are probably many cases where this still breaks, but it might be an improvement anyway :smile: 
#112

### DIFF
--- a/tests/encapsed/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/encapsed/__snapshots__/jsfmt.spec.js.snap
@@ -23,6 +23,6 @@ $encapsHereDoc2 = <<<STR
 Hello foo $a bar
 new line
 STR;
-$encapsedOffset = $a->bar$b;
+$encapsedOffset = $a->{bar$b};
 
 `;
--- a/tests/encapsed/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/encapsed/__snapshots__/jsfmt.spec.js.snap
@@ -11,7 +11,6 @@ $encapsHereDoc2 = <<<STR
 Hello foo $a bar
 new line
 STR;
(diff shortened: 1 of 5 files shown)


mgrip:
is it worth considering doing this from the parent level instead of the child?

ie when we print the property lookup, and `node.offset.kind !== "constref"`, then wrap in curly's? Might be simpler than trying to go the other way and trying to catch every edge case

czosel:
@mgrip Of course, i can't believe i didn't notice that :sweat_smile:

tests/encapsed/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -23,6 +23,6 @@ $encapsHereDoc2 = <<<STR
 Hello foo $a bar
 new line
 STR;
-$encapsedOffset = $a->bar$b;
+$encapsedOffset = $a->{bar$b};
```
czosel: Is `$a->bar$b` valid PHP?

nicoder: I don't think so, it seems to be a parse error in PHP7

```
php > $a = new stdClass;
php > $a->foobar = 1;
php > $b = 'bar';
php > echo $a->{"foo$b"};
1
php > echo $a->foo$b;

Parse error: syntax error, unexpected '$b' (T_VARIABLE), expecting ',' or ';' in php shell code on line 1
php > echo $a->{foo$b};

Parse error: syntax error, unexpected '$b' (T_VARIABLE) in php shell code on line 1
```

czosel: :+1: i'll remove it from the tests

Issue #182, opened 2018-03-08
Nested if inside function argument


mgrip:
input:
```php
$this->sort(function ($a, $b) use ($key, $ascending) {
  if (
    $a->$key == $b->$key
  ) {
    return 0;
  } else {
    if ($ascending) {
      return $a->$key < $b->$key ? -1 : 1;
    } else {
      return $a->$key < $b->$key ? 1 : -1;
    }
  }
});
```
output:
```php
$this->sort(function ($a, $b) use ($key, $ascending) {
    if (
        $a->$key == $b->$key
    ) { return 0; } else { if ($ascending) { return $a->$key < $b->$key ? -1 : 1; } else { return $a->$key < $b->$key ? 1 : -1; } }
});
```

mgrip:
Fixed in #184

PR #184, opened 2018-03-08
fix: nested if formatting in function parameter

mgrip:
#182

### DIFF
--- a/tests/closure/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/closure/__snapshots__/jsfmt.spec.js.snap
@@ -108,11 +108,9 @@ $longArgs_shortVars = function (
     // body
 };
 
-$shortArgs_longVars = function ($arg) use (
-    $longVar1,
-    $longerVar2,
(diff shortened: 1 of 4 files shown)


tests/closure/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -108,11 +108,9 @@ $longArgs_shortVars = function (
     // body
 };
 
-$shortArgs_longVars = function ($arg) use (
-    $longVar1,
-    $longerVar2,
-    $muchLongerVar3
-) {
+$shortArgs_longVars = function (
```
mgrip: I'm unsure on this one - but I think the larger problem is more important. If we think the original formatting is better, we'll probably have to set up a `conditionalGroup` for functions, figuring out the order of when to break the function arguments vs the group arguments

czosel: Totally agree. :+1:

PR #186, opened 2018-03-09
feat: improve long offset lookup output

evilebottnawi:
:star: 

Based on prettier output

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -367,16 +367,28 @@ function printExpression(path, options, print) {
           "::",
           path.call(print, "offset")
         ]);
-      case "offsetlookup":
+      case "offsetlookup": {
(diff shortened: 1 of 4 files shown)


src/printer.js
```diff
@@ -367,16 +367,28 @@ function printExpression(path, options, print) {
...
+        const isOffsetNumberNode = node.offset && node.offset.kind === "number";
+
         return group(
           concat([
             path.call(print, "what"),
             "[",
-            group(indent(concat([softline, path.call(print, "offset")]))),
-            softline,
+            group(
+              concat([
```
mgrip: is the extra ` concat()` necessary?

evilebottnawi: @mgrip need, for `isOffsetNumberNode ? "" : softline`

mgrip: 👍

PR #189, opened 2018-03-09
Fix curly wrapping of propertyLookup in chained expressions

czosel:


src/printer.js
```diff
@@ -312,6 +315,13 @@ function printArgumentsList(path, options, print, argumentsKey = "arguments") {
   return conditionalGroup(formsToConsider);
 }
 
+function wrapPropertyLookup(node, doc) {
+  const addCurly =
```
mgrip: should probably check that node is the right kind just to be safe, `node.offset.kind` could fatal if this is called incorrectly

czosel: good point!

PR #192, opened 2018-03-10
Function call wrapping, part 2

### LINKED ISSUE #164: Uglify arguments output with closure
Input:
```php
$foo->bar(
    $arg1,
    function ($arg2) use ($var1) {
        // body
    },
    $arg3
);
```
Output:
```php
$foo->bar($arg1, function ($arg2) use ($var1) {
    // body
}, $arg3);
```

I think we should always break lines if we have closure between

czosel:
Fixes #164

### DIFF
--- a/tests/closure/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/closure/__snapshots__/jsfmt.spec.js.snap
@@ -108,9 +108,11 @@ $longArgs_shortVars = function (
     // body
 };
 
-$shortArgs_longVars = function (
-    $arg
-) use ($longVar1, $longerVar2, $muchLongerVar3) {
+$shortArgs_longVars = function ($arg) use (
+    $longVar1,
+    $longerVar2,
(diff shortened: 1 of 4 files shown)


tests/closure/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -108,9 +108,11 @@ $longArgs_shortVars = function (
...
 };
 
-$shortArgs_longVars = function (
-    $arg
-) use ($longVar1, $longerVar2, $muchLongerVar3) {
+$shortArgs_longVars = function ($arg) use (
+    $longVar1,
+    $longerVar2,
+    $muchLongerVar3
+) {
```
czosel: This change was not intentional, but i'm not sure if it's actually for the better?

mgrip: 👍 agreed!

PR #194, opened 2018-03-11
tests: newlines

evilebottnawi:
> All PHP files MUST use the Unix LF (linefeed) line ending only.

> All PHP files MUST end with a non-blank line, terminated with a single LF.

Just tests


evilebottnawi:
Ready for review :+1:

PR #197, opened 2018-03-11
fix: don't break identifier

evilebottnawi:
We can't change case or other changes in `identifier`, because it is can break code `$array[oBjEcT]` where `object` is constant

### DIFF
--- a/tests/case/case.php
+++ b/tests/case/case.php
@@ -231,6 +231,21 @@ FINAL CLASS MyClass EXTENDS Bar IMPLEMENTS iTemplate
     {
         // Nothing
     }
+
+    public function parentTest()
+    {
+        $array = [];
+        echo $array[iNt];
+        echo $array[sElF];
+        echo $array[pArEnT];
+        echo $array[aRaRay];
+        echo $array[oBjEcT];
+
+        SELF::who();
+        STATIC $a = 'test';
+
+        return PARENT::who();
+    }
 }
 
 __HALT_COMPILER();
--- a/src/clean.js
+++ b/src/clean.js
@@ -30,22 +30,54 @@ function clean(node, newObj) {
   }
 
   // All reserved words should be lowercase case
-  if (node.kind === "identifier" && typeof node.name === "string") {
-    const lowerCasedName = node.name.toLowerCase();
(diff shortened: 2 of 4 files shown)


tests/case/case.php
```diff
@@ -231,6 +231,21 @@
...
+        echo $array[iNt];
+        echo $array[sElF];
+        echo $array[pArEnT];
+        echo $array[aRaRay];
+        echo $array[oBjEcT];
+
+        SELF::who();
+        STATIC $a = 'test';
+
+        return PARENT::who();
```
evilebottnawi: All this example works in php

tests/case/case.php
```diff
@@ -231,6 +231,21 @@
...
         // Nothing
     }
+
+    public function parentTest()
+    {
+        $array = [];
+        echo $array[iNt];
+        echo $array[sElF];
+        echo $array[pArEnT];
+        echo $array[aRaRay];
```
nicoder: is `aRaRay` meant to be `array` with a different case?

nicoder: sure, but since the other constants (`iNt`, `sElF`, `pArEnT`) looked like keywords (`int`, `self`, `parent`), I wondered if `aRaRay` was also meant to

evilebottnawi: @nicoder it is also constant 
http://php.net/manual/en/reserved.keywords.php vs http://php.net/manual/en/reserved.other-reserved-words.php :trollface: php very interesting language

aboyton: I think the point was that `array` is spelt wrong.

evilebottnawi: @aboyton just run:
```
define('aRaRay', 'foo');

$array = [aRaRay => 1];
echo $array[aRaRay];
```

mgrip:
This is adding a decent amount of complexity... whats the problem we're trying to solve exactly? Is there a corresponding issue open?

evilebottnawi:
@mgrip i found this problem in own code base, i use contant `OBJECT` as key array and it is break code. But php allow this contant

evilebottnawi:
Closed in favor https://github.com/prettier/plugin-php/pull/632

PR #202, opened 2018-03-12
fix: don't loose type of `useitem`

evilebottnawi:
Bug :smile:

### DIFF
--- a/tests/use/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/use/__snapshots__/jsfmt.spec.js.snap
@@ -20,6 +20,16 @@ use Vendor\\Package\\SomeNamespace\\{
     SubnamespaceTwo\\ClassY,
     ClassZ
 };
+use Mizo\\Web\\{
+    Php\\WebSite,
+    Php\\KeyWord,
+    Php\\UnicodePrint,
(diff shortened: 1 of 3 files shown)


tests/use/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -29,6 +39,14 @@ use MyVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeyLon
...
 use function some\\Full\\{fn_a, fn_b, fn_c};
 use const some\\Full\\{ConstA, ConstB, ConstC};
+use Mizo\\Web\\{Php\\WebSite,
+    Php\\KeyWord,
+    Php\\UnicodePrint,
+    JS\\JavaScript,
+    function JS\\printTotal,
+    function JS\\printList,
+    const JS\\BUAIKUM,
+    const JS\\MAUTAM};
```
mgrip: i think this might have gotten solved in the other PR, but line break for these braces

evilebottnawi: @mgrip yep, already solved, need rebase :+1:

PR #208, opened 2018-03-12
feat: improve `inline` output

evilebottnawi:
It is not easy :+1:  Why remove `printLines`?
1. One place where we put `;` and `newlines`.
2. Easy reading.
3. For future: here we can add/remove unnecessary nodes(example - `parens` around arithmetic operation how it do `prettier` or better output `inline`).

After add `\n` in `php-parser` (already added, need just update) output will be more better :+1:

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -17,12 +17,23 @@ const hardline = docBuilders.hardline;
 const softline = docBuilders.softline;
 const willBreak = docUtils.willBreak;
 
-const makeString = sharedUtil.makeString;
+const includes = util.includes;
+const getNodeListProperty = util.getNodeListProperty;
+const getLast = util.getLast;
+const isFirstNodeInParentProgramNode = util.isFirstNodeInParentProgramNode;
+const isFirstNodeInParentNode = util.isFirstNodeInParentNode;
+const isLastNodeInParentNode = util.isLastNodeInParentNode;
+const isPrevNodeInline = util.isPrevNodeInline;
+const isNextNodeInline = util.isNextNodeInline;
+const lineShouldHaveStartPHPTag = util.lineShouldHaveStartPHPTag;
+const lineShouldEndWithSemicolon = util.lineShouldEndWithSemicolon;
+const lineShouldEndWithHardline = util.lineShouldEndWithHardline;
+const fileShouldEndWithHardline = util.fileShouldEndWithHardline;
+const lineShouldHaveEndPHPTag = util.lineShouldHaveEndPHPTag;
+const shouldRemoveLines = util.shouldRemoveLines;
+const removeNewlines = util.removeNewlines;
 
-// polyfill for node 4
-function includes(array, val) {
-  return array.indexOf(val) !== -1;
-}
+const makeString = sharedUtil.makeString;
 
 function shouldPrintComma(options) {
   switch (options.trailingComma) {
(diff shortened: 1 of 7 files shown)


evilebottnawi:
Very difficult PR, feel free to feedback

src/printer.js
```diff
@@ -152,24 +129,134 @@ function isLastStatement(path) {
...
+    lineShouldHaveEndPHPTag(path)
+      ? concat([
+          isFirstNodeInParentNode(path) || !isPrevNodeInline(path)
+            ? hardline
+            : " ",
+          "?>"
+        ])
+      : "",
+    fileShouldEndWithHardline(path) ? hardline : ""
+  ]);
```
evilebottnawi: Should be easy to read and maintenance :+1:

src/printer.js
```diff
@@ -42,13 +43,17 @@ function getLast(arr) {
   return null;
 }
 
+function getNodeListProperty(node) {
```
mgrip: i think @czosel ended up adding like a `getBody` function that does something similar?

src/printer.js
```diff
@@ -59,41 +64,14 @@ function getNodeIndex(path) {
...
-  if (!n) {
-    return "";
-  } else if (typeof n === "string") {
-    return n;
-  }
-  if (n.kind === "program" && n.children.length === 0) {
-    return concat(["<?php", hardline]);
-  }
-  const printed = printNode(path, options, print);
+function lineShouldEndWithSemicolon(path) {
```
mgrip: github is doing a terrible job displaying this diff ☹️

src/printer.js
```diff
@@ -152,24 +130,138 @@ function isLastStatement(path) {
...
+    "namespace",
+    "class",
+    "interface",
+    "trait",
+    "traituse",
+    "declare"
+  ].includes(node.kind);
+}
+
+function isFirstNodeInParentProgramNode(path) {
```
mgrip: would all of these functions make sense to go in the util file? the printer is getting crazy long

src/printer.js
```diff
@@ -152,24 +130,138 @@ function isLastStatement(path) {
...
+}
+
+function isNextNodeInline(path) {
+  const parentNodeBody = getParentNodeListProperty(path);
+  const nodeIndex = getNodeIndex(path);
+  const nextNode = nodeIndex !== -1 && parentNodeBody[nodeIndex + 1];
+  return nextNode && nextNode.kind === "inline";
+}
+
+function shouldRemoveRemoveLines(path) {
```
mgrip: this seems weird to me - is this fn specifically used for inline stuff only?

src/printer.js
```diff
@@ -152,24 +130,138 @@ function isLastStatement(path) {
...
+
+function lineShouldHaveStartPHPTag(path) {
+  return isFirstNodeInParentProgramNode(path) || isPrevNodeInline(path);
+}
+
+function lineShouldHaveEndPHPTag(path) {
+  return isNextNodeInline(path);
+}
+
+function fileShouldEndWithHardline(path) {
```
mgrip: is there a reason this should be separate from `lineShouldEndWithHardline`?

src/printer.js
```diff
@@ -152,24 +130,138 @@ function isLastStatement(path) {
...
+  const node = path.getValue();
+  const isProgramNode = node.kind === "program";
+  return (
+    isProgramNode && node.children[node.children.length - 1].kind !== "inline"
+  );
+}
+
+function genericPrint(path, options, print) {
```
mgrip: theres a lot going on here - a comment explaining the basic idea behind the logic would be good

src/printer.js
mgrip: agreed looks great 🎉

src/printer.js

src/printer.js
evilebottnawi: @mgrip `lineShouldEndWithHardline` for statement, `fileShouldEndWithHardline` for `program` node - i.e. end of file

src/printer.js
evilebottnawi: @mgrip agree :disappointed:

src/printer.js
evilebottnawi: @mgrip we can rename to `getBody`, do it?

evilebottnawi:
@mgrip i agree, it is hard to handle `inline`, but we should do it :star:

src/printer.js

evilebottnawi:
@mgrip basic refactor done :+1: Feel free to feedback

mgrip:
we should probably have @czosel take a quick look as well since this is a big one but looks great to me! im good with merging 👍

evilebottnawi:
@mgrip great! It is first part of imrove `inline`. I think i can do all pretty in `3-5` parts :smile:

evilebottnawi:
Next PR will be less :+1:

PR #215, opened 2018-03-14
feat: method declaration PSR compat

### LINKED ISSUE #171: Long function declaration ugly
Input:
```php
public function longLongAnotherFunction(string $foo, string $bar, int $baz): string 
{
    return 'foo';
}
```

Output:
```php
public function longLongAnotherFunction(
    string $foo,
    string $bar,
    int $baz
): string
{
    return 'foo';
}
```

Should be:
```php
public function longLongAnotherFunction(
    string $foo,
    string $bar,
    int $baz
): string {
    return 'foo';
}
```

evilebottnawi:
fixes #171 
Some hack, if somebody have time to do PR in prettier - welcome! We need `invertLine`, logic as line, but versa vice, i.e. when group breaks we should insert `" "`, if does not break - insert `hardline`

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -758,17 +758,15 @@ function printStatement(path, options, print) {
       returnTypeContents = "",
       bodyContents = ""
     }) {
-      return concat([
-        concat([
-          group(
+      const isClassLikeNode = ["class", "interface", "trait"].includes(
+        node.kind
+      );
+      const printedDeclaration = group(declaration);
+      const printedSignature = !isClassLikeNode
+        ? group(
             concat([
-              declaration,
-              !["class", "interface", "trait"].includes(node.kind) ? "(" : ""
-            ])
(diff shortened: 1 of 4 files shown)


src/printer.js
```diff
@@ -758,45 +758,62 @@ function printStatement(path, options, print) {
...
+                : "",
+              ")",
+              returnTypeContents ? concat([": ", returnTypeContents]) : ""
             ])
           )
-        ]),
+        : "";
+      const printedBody = bodyContents
+        ? concat([
+            // see https://github.com/prettier/plugin-php/issues/107
```
mgrip: this should move down to ~804 i think?

evilebottnawi: @mgrip oh yep

evilebottnawi:
@mgrip done, also remove todo

Issue #191, opened 2018-03-09
Inline comments are moved to next line


nikolasmatt:
Encountered this issue on old file.
```
use Forbidden\Classes\Secret_Class; // @codingStandardsIgnoreLine because I like to sweep problems under the rug
```
turns to
```
use Forbidden\Classes\Secret_Class; 
// @codingStandardsIgnoreLine because I like to sweep problems under the rug
```
While this is a non-breaking change in php, the [PHP_Codesniffer](https://github.com/squizlabs/PHP_CodeSniffer) tag no longer points to the right line.

evilebottnawi:
@nikolasmatt i think we should have don't move `php_codesniffer` comments

mgrip:
I don't think it makes a difference that its a `php_codesniffer` comment or not, we should just keep inline comments as is

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMAWBLAzgOjAQwBtCAxAV1g2gAoBKAbgAIB6ZxrCAWzkck+9gAdKMOHMAVOOGMI6OACdeXATGnTxzUSKitG8LKp1t9h9NhwAzCmBhUoAYSKFqdRgF4AfI2DT2XOADycvIAKphQAOZ0TLoYUIRxcMIAvrQgADQgEAAOttBYyKD48vIQAO4ACsUIBSj4AG4QGAAmGSD4BsgWRFhwmQBG8vhgANZwMADK2cNxEcgw8mR9IM0QYF09y3G98jAVQxGc+BuEvZkAVlgAHgBCQ6PjE-jcADKJJ2dZZDDZ3wBMH2W03kO2QIH6+H6AE9CNA2tl5HEYAB1FroZAADgADJkERBesihtkwQi4Dt6nA2vI4ABHMgYan7fCHY5IbqnZa9TgYeaLTmzQhwACKZFklLZm0yMEhqOa6KQfylQwwCUi9mUrJAUGglMyZF6IUhtXZvWSySAA

evilebottnawi:
@mgrip yes you are right :+1:

cmancone:
I believe that I am seeing this same issue.  It doesn't break anything at all for me, but when I run `prettier --debug-test` on some files with inline comments it says that it broke when it didn't actually break.  If I run without the debug-test flag it prettifies it just fine, and the results are syntactically identical, but inline comments have indeed been moved to the next line (I'm guessing that is why `debug-test` thinks it broke?).

Sample code:
```php
<?php
[...]

Issue #190, opened 2018-03-09
Massage comments


czosel:
Running comments through `--debug-check` currently fails. Example Input:
```php
/*
 * This file is part of Composer.
 *
 * (c) Nils Adermann <naderman@naderman.de>
 *     Jordi Boggiano <j.boggiano@seld.be>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Composer\Autoload;
```

```diff
[error] ../ClassLoader.php: prettier(input) !== prettier(prettier(input))
[error] Index: 
[error] ===================================================================
[error] --- 
[error] +++ 
[error] @@ -1,10 +1,12 @@
[error]  <?php
[error] -/** This file is part of Composer.
[error] +/**
[error] + * This file is part of Composer.
[error]   *
[error]   * (c) Nils Adermann <naderman@naderman.de>
[error] - *     Jordi Boggiano <j.boggiano@seld.be>
[error] + * Jordi Boggiano <j.boggiano@seld.be>
[error]   *
[error]   * For the full copyright and license information, please view the LICENSE
[error] - * file that was distributed with this source code. */
[error] + * file that was distributed with this source code.
[error] + */
[error]  
[error]  namespace Composer\Autoload;
[error]
```

In this particular example there seem to be two issues:

* the added extra lines at the begging and end should be massaged away
* the removed indent in `Jordi Boggiano <j.boggiano@seld.be>` is unwanted (right?)

PS: I think this is the last kind of breaking AST change in my codebase! :tada:

mgrip:
will be fixed with https://github.com/prettier/plugin-php/pull/217

andreiglingeanu:
Just stumbled on the fact that the parser removes indentation from my comments, which I very carefully wrote, for some detailed illustration of a difficult concept. I think it would be saner to not touch the comments & perform such destructive changes to them.

evilebottnawi:
@andreiglingeanu it is bug, already fixed in master branch in `parser`

andreiglingeanu:
Actually it is not fixed in `master`, but #217 does fix it. But, #217 introduces another comment-related breaking change.

evilebottnawi:
@andreiglingeanu in master in `parser` (here https://github.com/glayzzle/php-parser) :smile: Yep PR will fix this, near future I'll finish it

andreiglingeanu:
Oh, got it :) Thanks for the awesome work!

[...]

PR #217, opened 2018-03-14
Feat update php parser

evilebottnawi:
Prepare to new release. Feel free to feedback/fix/commit in this branch


evilebottnawi:
@azz can you help us with this? Why comments inside `class`?

mgrip:
@evilebottnawi can you create an issue for this worth some more detail on what we're trying to accomplish?

evilebottnawi:
@mgrip it is just PoC, attach comments from `comments`, which in future version will be in `program.comments` (all good parsers do this)

evilebottnawi:
@mgrip you can see what all comments now printed, some on incorrect place

mgrip:
I still think it would be good to have an issue open where we can discuss this more - I'm unclear what benefits we'd get from attaching comments from `program.comments`

mgrip:
also are there other benefits we'd get from updating the parser?

evilebottnawi:
@mgrip Examples:
```php
$var /* comment */ = /* comment */ 'string';
```
Right now is is impossible. Also add newlines between statement which have leading comments. Also in future version we don't have `doc` (i.e. comment) node in ast.

evilebottnawi:
@mgrip Fixes with `public`, `exit` and `die`, `shortForm` echo - `<?= ?>` and etc

evilebottnawi:
@mgrip i update PR, you can see what we don't have `doc` node anymore (in CI or run locally)

src/index.js
```diff
@@ -48,7 +48,36 @@ const parsers = {
...
   php: {
     print,
-    massageAstNode: clean
+    massageAstNode: clean,
+    getCommentChildNodes(node) {
+      let children = node.children || node.body || node.traits;
+      if (!children) {
+        return [];
+      }
+      if (children.kind && children.kind === "block") {
```
nicoder: I guess this would be equivalent:`if (children.kind === "block") {`

do you prefer the longer version?

evilebottnawi: @nicoder It is PoC, but you are right :+1:

mgrip:
@evilebottnawi would this fix https://github.com/prettier/plugin-php/issues/191 and/or https://github.com/prettier/plugin-php/issues/190?

evilebottnawi:
@mgrip Yes and Yes

evilebottnawi:
@mgrip yes, we can fix and implement new features using latest master version from git (it is allow early bug reporting and easy migrate on 3 version). Help welcome :+1: Need solve problem with comments some comments on incorrect places, need investigate why and if it is bug we will report about this.

mgrip:
I am so confused haha. I have the branch checked out and am trying to debug through how this works but I feel like I'm missing something. How do the comments actually get printed now? For example the class node - it seems like there's a new `class.comments` and `class.leadingComments` attribute, how does `printComment()` know how to print those?

mgrip:
Believe this is related https://github.com/glayzzle/php-parser/issues/135

evilebottnawi:
@mgrip using offset in position :smile: Better debug logic and look how it is work :smile:

andreiglingeanu:
Here's a quick bug that I discovered (in this branch):

Before:

```php
<?php

/**
 * test comment
 */

class Test
{
	public function testme($param)
	{
	    return null;
	}

}
```

After:

```php
/**
 * test comment
 */

<?php
class Test
{
	public function testme($param)
	{
		return null;
	}
}
```

When we start the file with a comment, it moves the `<?php` tag right before the `class` declaration, leaving the comment at the top. I believe this is a breaking change.

mgrip:
thanks for the bug report @andreiglingeanu!

@evilebottnawi I'm making some good progress here, will push an update to the `feat-update-php-parser` branch at some point today. One thing I'm worried about though is that a lot of the built-in logic for comments seems specific to the estree ast https://github.com/prettier/prettier/blob/master/src/main/comments.js#L794 - not sure how big of a problem it will end up being, but what do you think?

evilebottnawi:
@mgrip let's push here all and create issue in `prettier` around problem, seems something should be redefined inside `parser` option

mgrip:
@evilebottnawi just updated - still one failing test, but let me know what you think so far!

src/printer.js
```diff
@@ -3,6 +3,8 @@
 const docBuilders = require("prettier").doc.builders;
 const docUtils = require("prettier").doc.utils;
 const sharedUtil = require("prettier").util;
+//@TODO: super hacky, remove once comment functions are exported for plugin use
+const comments = require("../node_modules/prettier/src/main/comments");
```
mgrip: https://github.com/prettier/prettier/issues/4177

evilebottnawi:
@mgrip great work :+1: I'll see what I can fix

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -106,17 +106,25 @@ class Foo
...
- * Some long comment *
- *********************/
-// This is a one-line c++ style comment
-/* This is a multi line comment
-   yet another line of comment */
-/* This is a multiple-lines comment block
-that spans over multiple
-lines */
-// This is a one-line shell-style comment
+    /**
```
evilebottnawi: I think we should't change comment output, print as is

evilebottnawi: It is very bad practice

evilebottnawi: you can tests how prettier handle comment using `playground`

mgrip: 👍 I agree. The printComments function needs some more work, was just trying to get something to cover the most common cases for now but agreed its currently too opinionated

evilebottnawi:
/cc @mgrip @czosel 
Tests passed (except `AST_COMPARE=1`). Roadmap (asap - we should solve this problem first before merge anything in master):
1. Normalize only `isDoc` comment, comments what are not `isDoc` print as is (example look prettier logic).
2. Some comments have incorrect places, need investigate and solve.
3. Update `cleaner`.
4. Add more tests around comments.

evilebottnawi:
We can't continue work on this until `prettier` don't solve problem around api :disappointed:

mgrip:
As far as step 2. goes, I did some debugging and I believe its because some of the logic to attach comment nodes in the right place is ast-specific (https://github.com/prettier/prettier/issues/4177#issuecomment-374392660). `decorateComment()` is working as expected (precending, following, and enclosing nodes get populated correctly), but then we need to have access to the "attach" part of the process to handle different node types and situations for comments

mgrip:
I think this PR is pretty important for us to get over the line to get to a v1 - do we think we should focus on contributing to prettier and php-parser to get these dependencies solved before solving more issues here?

evilebottnawi:
@mgrip yes, it will be great, but some issue can take many times before will be merged and release :disappointed:

czosel:
I'd suggest we try to fix the things that block us in our dependencies. Doesn't really matter when something is released, we can just put the right commit in package.json (like we do right now with our prettier dependency).

czosel:
@mgrip Awesome! Just in case you didn't find the right file in prettier yet - this is the util which is available inside plugins: https://github.com/prettier/prettier/blob/master/src/common/util-shared.js

mgrip:
Just pushed a patch based on https://github.com/prettier/prettier/pull/4182 - let me know what you guys think. We can always revert or change if things change on the prettier end. Mostly wanted to verify this would actually work

mgrip:
cc @evilebottnawi @czosel

mgrip:
tests are passing! 🎉

czosel:
@mgrip looks like deleting the travis cache did the job! :smile: 
Unfortunately, I don't have time to take a closer look at this right now - but it certainly looks great on first glance and I'm truly excited for our first release :wink:

evilebottnawi:
@czosel we can merge this master and when fix after report users, we can handle right now all possible comments location :+1:

tests/case/case.php
```diff
@@ -1,236 +1,4 @@
...
-
-ABSTRACT CLASS AbstractClass
-{
-    ABSTRACT PROTECTED FUNCTION getValue();
-    ABSTRACT PROTECTED FUNCTION prefixValue($prefix);
-
-    PUBLIC function printOut() {
-        PRINT $this->getValue();
-    }
+FOR ($i = 0; $i < 10; $i++) {
```
evilebottnawi: It seems I accidentally deleted tests, need return /cc @mgrip

mgrip: looks like there are a couple types of comments not getting printed when i add it back. ill get it updated and fixed, then I think we'll be good to merge 👍

mgrip:
@evilebottnawi added back `case` test, updated 👍

evilebottnawi:
@mgrip great work! Ready for review and merge?

mgrip:
I think so! 🎉

src/printer.js
```diff
@@ -1484,12 +1534,12 @@ function printNode(path, options, print) {
       }
       return concat(parts);
     }
-    case "doc": {
+    /*case "doc": {
```
evilebottnawi: Let's remove this node :+1:

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -106,17 +106,100 @@ class Foo
         // there should be no docblock here
         $this->title = $title;
     }
+
+    /**
```
evilebottnawi: Let's check `isDoc` comment (maybe use regex or package), if yes - pretty output, if no - print as is :+1:

mgrip:
@evilebottnawi all set!

evilebottnawi:
@mgrip approve :+1: let's wait what says @czosel and feel free to merge. Tomorrow i will test on big code base with comments and send PR with fixes :+1:

src/comments.js
```diff
@@ -0,0 +1,129 @@
...
+const handleOwnLineComment = function(comment) {
+  if (
+    handleClass(comment) ||
+    handleFunction(comment) ||
+    handleForLoop(comment) ||
+    handleTryCatch(comment)
+  ) {
+    return true;
+  }
+  return false;
```
czosel: Why not just `return handleClass(comment)  || ...` (no extra `if`)?

mgrip: good point 😄 i'll update now

mgrip: all set 👍

src/printer.js
czosel: Isn't this fixed in prettier core now?

czosel: Apart from this, this looks all reaaally good! 💯

mgrip:
I copied the `printDanglingComments` function from prettier since technically it is part of the printer, I think it makes sense to be plugin-specific. Think we should be all set!

mgrip:
thanks everyone, this was a big one! 🎉

PR #220, opened 2018-03-15
fix: `declare` statement and `inline`

evilebottnawi:
fix :star:

### DIFF
--- a/tests/inline/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/inline/__snapshots__/jsfmt.spec.js.snap
@@ -1,5 +1,24 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`declare.php 1`] = `
+<?php declare(strict_types=1); ?>
+<html>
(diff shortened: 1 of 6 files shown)


tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -271,6 +290,37 @@ $a = 1;
...
+exports[`mixed-3.php 1`] = `
+<?php if ($expression == true): ?>
+    <p>Test.</p>
+<?php else: ?>
+    <p>Test.</p>
+<?php endif; ?>
+<div>Hello world</div>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+if ($expression == true): ?>    <p>Test.</p>
```
evilebottnawi: In some next PR i will fix it :+1:

PR #222, opened 2018-03-15
fix: echo indent

evilebottnawi:
Improve `echo` and `print` :star:

### DIFF
--- a/tests/echo/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/echo/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,184 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`echo.php 1`] = `
+<?php
+
+echo "Hello World";
+
+echo "This spans
+multiple lines. The newlines will be
+output as well";
+
+echo "This spans\\nmultiple lines. The newlines will be\\noutput as well.";
+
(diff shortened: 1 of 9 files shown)


evilebottnawi:
@czosel @mgrip ready for review. Why print and echo in one PR, because they have same logic and many places.

evilebottnawi:
Also remove parens from arguments, all Laravel/Symfony/Yii2/PSR don't use their.

tests/echo/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,184 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`echo.php 1`] = `
+<?php
+
+echo "Hello World";
+
+echo "This spans
+multiple lines. The newlines will be
+output as well";
```
nicoder: (not the topic of this PR but) the line feeds inside this string are transformed into `\n`

this code:

```php
echo "This spans
multiple lines. The newlines will be
output as well";
```

becomes:

```php
echo "This spans\\nmultiple lines. The newlines will be\\noutput as well";
```

is this what we want?

evilebottnawi: @nicoder it is bug in parser, will be fixed in new version :+1: When we update php parser we will update all snapshots :+1:

PR #229, opened 2018-03-16
feat: improve `switch` output

evilebottnawi:
Based on https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L1631. Also all tests were adapted.

P.S. We don't break anything in switch statement :star:

### DIFF
--- a/tests/switch/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/switch/__snapshots__/jsfmt.spec.js.snap
@@ -1,5 +1,170 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`empty_lines.php 1`] = `
+<?php
+
+switch (foo) {
+    case "bar":
+        doSomething();
+
+    case "baz":
+        doOtherThing();
+}
+
+switch (foo) {
+
+    case "bar":
+        doSomething();
+
+    case "baz":
+        doOtherThing();
+}
+
+switch (foo) {
+    case "bar":
+        doSomething();
+
+    case "baz":
+        doOtherThing();
+
+}
+
+switch (foo) {
+    case "bar":
+        doSomething();
+
+
+
+    case "baz":
+        doOtherThing();
+}
+
+switch (x) {
+    case y:
+        call();
+
+        break;
+
+    case z:
+        call();
+
+        break;
+}
+
+switch (a) {
+    case b:
+        if (1) {};
+        c();
+}
+
+switch (a) {
+    case x:
+    case y:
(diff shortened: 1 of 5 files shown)


tests/switch/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -107,15 +301,17 @@ switch ($expr) {
...
-) {
+switch ($longLongLongLongLongLongLongLongLongLongLongLongLongLongLongLongVariable) {
 }
 
-switch (
-    $longLongLongLongLongLongLongVariable &&
-    $longLongLongLongLongLongLongVariable
-) {
+switch ($longLongLongLongLongLongLongVariable &&
+    $longLongLongLongLongLongLongVariable) {
```
evilebottnawi: [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBnA7gSxmAFgAgAoASAG2gHMAZSmqa2x+uigNQEMAnTdgI1Lj4AZEIA6UfGSYNm0liw7c+A4WJj4Nk8rJ0y98yop784ASnzAAvvhAAaEBAAOMTNFTJQXThHQAFLgjuKOwAbhCYACZ2IOyoMMgAZuykqHD2vJzsYADWcDAAyo5ZmPTIMJwArmkgERBgicmp9iWpnDC+mRQAtuwNKdUAVqgAHgBCmTl5+exdcFQlcH1NDhUwjqsATEvVRZytyCC8fACe2tGO3LAA6pEwuMgAHAAM9hcQqVeZjgcXcK0hi3snDgAEcKphgR12N1ekgkv17KkupgypVqqgShQBABFCoQeDbewwPg3CJ3ZAbImZTCkTEAYQgXR6Byg0EBIAqqQAKnwgvDUpZLEA).

Should we do as `js`?

evilebottnawi: @vjeux it is bug in `prettier` or expected behavior? Don't found issues around this.

vjeux: Looks like we're not applying the same treatment for switch as for `if/while/...`. This is not intended. `switch` is very rarely used in js so it probably never come up in real code.

vjeux: If you can send a fix to prettier that would be very welcome :)

evilebottnawi: @vjeux yep, i will send PR :+1:

PR #235, opened 2018-03-19
tests: `yield` node

evilebottnawi:
Very ugly, any ideas how we can print this better?


evilebottnawi:
Merge as is, let's do better output in future. Feel free to feedback.

PR #243, opened 2018-03-20
feat: strip leading slash from `use` node

evilebottnawi:
From php.net:
> Note that for namespaced names (fully qualified namespace names containing namespace separator, such as Foo\Bar as opposed to global names that do not, such as FooBar), the leading backslash is unnecessary and not recommended, as import names must be fully qualified, and are not processed relative to the current namespace.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -352,6 +352,14 @@ function fileShouldEndWithHardline(path) {
   return true;
 }
 
+function maybeStripLeadingSlashFromUse(name) {
+  const nameWithoutLeadingSlash = name.replace(/^\\/, "");
+  if (nameWithoutLeadingSlash.indexOf("\\") !== -1) {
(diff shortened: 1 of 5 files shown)


src/util.js
```diff
@@ -352,6 +352,14 @@ function fileShouldEndWithHardline(path) {
   return true;
 }
 
+function maybeStripLeadingSlashFromUse(name) {
+  const nameWithoutLeadingSlash = name.replace(/^\\/, "");
```
czosel: Wouldn't it be simpler to just always return the replaced string?

evilebottnawi: @czosel `use Exception;` and `use \Exception` with namespace have difference effect

czosel: Oh right, you're just removing leading backslashes if there's another backslash remaining :+1:

PR #255, opened 2018-03-26
fix: print comments in array

evilebottnawi:
Some comment will be printed better (as do prettier). We need implement `printComments` function as in prettier or move in shared utils. Let's merge this as is and in next PR fix comment places.

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -553,11 +553,12 @@ function printExpression(path, options, print) {
           node.shortForm ? "[" : "array(",
           indent(
             concat([
-              softline,
+              node.comments && node.comments.length > 0 ? hardline : softline,
               join(concat([",", line]), path.map(print, "items"))
             ])
           ),
           ifBreak(shouldPrintComma(options) ? "," : ""),
+          comments.printDanglingComments(path, options, /* sameIndent */ true),
           softline,
           node.shortForm ? "]" : ")"
         ])
@@ -1540,11 +1541,20 @@ function printNode(path, options, print) {
       }
       return concat(parts);
     }
-    case "entry":
+    case "entry": {
+      const dangling = comments.printDanglingComments(
+        path,
+        options,
+        /* sameLine */ true
+      );
+      const printedComments = dangling ? concat([hardline, dangling]) : "";
+
       return concat([
         node.key ? concat([path.call(print, "key"), " => "]) : "",
-        path.call(print, "value")
+        path.call(print, "value"),
+        printedComments
       ]);
+    }
     case "traituse":
       return group(
         concat([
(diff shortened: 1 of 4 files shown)


src/printer.js
```diff
@@ -553,11 +553,12 @@ function printExpression(path, options, print) {
           node.shortForm ? "[" : "array(",
           indent(
             concat([
-              softline,
+              node.comments && node.comments.length > 0 ? hardline : softline,
```
mgrip: do we actually want to do this? seems like as long as its not more than the line length we shouldn't be enforcing new lines just because there are comments

mgrip: https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAE8OYLyYDaAjADSYD0AVJgJZQA29cmkAtmwllRZgEwBdEKRAQADjFrpkoAIYAneRADuABQUI0yELIBuEWgBNhOjMgBmshmjgiARvNlgA1nBgBlMU-oBzZDHkAV1sQQwgwCysbEXobeRhVRx82WUjrEIArNAAPACFHFzd3WU4AGWY06NFAmDEavkqQr3k47TtZOwBPBmgTMXl6GAB1IxgAC2QADgAGEX6IGyHHMW1+uDjdOBN5OABHQNodxNlk1KRLdJEbNlp-IJC0XwY4AEVAiHhGkRgOkcNx5B8b6OWhMKA+ADCEA4ZxAUGgWxEgRsABUOlpzlE4ABfbFAA

evilebottnawi: @mgrip looks we should break only we have leading or dangling comments :confused: But improve output comment i think we should do in other PR. First goal is print all comments without break code.

evilebottnawi: Also related https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAE8OYLyYDaA9MZgIwA6UmFANNbQEwM2YDM1AugNwh0gIABxgBLdMlABDAE4yIAdwAKshGmQgpANwiiAJv00ZkAMykAbNHAEAjGVLABrODADKQh6KgBzZDBkArtYgehBgphZWAl5WMjBK9t4AtlIRlsEAVmgAHgBC9k4urlJJcAAyXnBpUYIBMEJ1TNXBHjKxGjZSNgCe5tCGQjJeMADq+jAAFsgAHAAMAoMQViP2QhqDcLFaVQIycACOAaJ7CVLJqUhm6QJWSaJ+gcFoXt7mcACKARDwzQIwXWM9JNkCwQP4pKJzC8AMIQJIpDRQaA7EABKwAFS66kukTgAF88UA

mgrip: i agree, im just questioning if this line should change at all or just remain as `softline` always

evilebottnawi: @mgrip softline break when code
```php
$array=[ // test
 1
];
```
Output:
```php
$array=[// test 1 ];
```

mgrip:
Fixes https://github.com/prettier/plugin-php/issues/258

src/printer.js
mgrip: ahh ok got it. ok ya agree lets just merge this to get code working and circle back on the formatting if we need to

PR #259, opened 2018-03-28
tests: adding tests for destructure

mgrip:
* adding test for array desctructuring
* fixing bug in ast clean function


mgrip:
also updating snapshot (for test currently failing in master)

Issue #253, opened 2018-03-25
Don't change embedded newlines into \n


aboyton:
Related to #29 (although the opposite), I think it would be nice if we also didn't change embedded newlines into inline `\n`s.

```php
$query =
    'SELECT *
     FROM users
     WHERE u.id IS NULL';
```
becomes
```php
$query = 'SELECT *\n     FROM users\n     WHERE u.id IS NULL';
```

aboyton:
Depending how high priority this is I can try take a look next week.

PR #263, opened 2018-03-29
fix: add string escape for encapsed

mgrip:
https://github.com/prettier/plugin-php/issues/29

Unfortunately I don't think we can fix https://github.com/prettier/plugin-php/issues/253 right now. Maybe we can utilize the `raw` attribute from the parser? But `node.value` is exactly the same whether its a literal line break or escaped

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -477,7 +477,10 @@ function printExpression(path, options, print) {
         // @TODO: for now just reusing double/single quote preference from doc. could eventually
         // use setting from options. need to figure out how this works w/ complex strings and interpolation
         // also need to figure out splitting long strings
-        const quote = node.isDoubleQuote ? '"' : "'";
+        let quote = node.isDoubleQuote ? '"' : "'";
+        if (path.getParentNode().kind === "encapsed") {
+          quote = "";
+        }
         return makeString(stringEscape(node.value), quote, false);
       }
       case "number":
@@ -494,7 +497,9 @@ function printExpression(path, options, print) {
             path.map(valuePath => {
               const node = valuePath.getValue();
               if (node.kind === "string") {
-                return node.value;
+                return valuePath.getParentNode().type === "heredoc"
+                  ? node.value
+                  : print(valuePath);
               } else if (node.kind === "variable") {
                 if (typeof node.name === "object") {
                   return concat(["${", path.call(print, "name"), "}"]);
--- a/src/printer.js
+++ b/src/printer.js
@@ -33,7 +33,6 @@ const {
   printNumber,
   shouldFlatten,
   shouldRemoveLines,
-  stringEscape,
   removeNewlines,
   maybeStripLeadingSlashFromUse,
   fileShouldEndWithHardline
@@ -478,10 +477,19 @@ function printExpression(path, options, print) {
         // use setting from options. need to figure out how this works w/ complex strings and interpolation
         // also need to figure out splitting long strings
         let quote = node.isDoubleQuote ? '"' : "'";
+        let stringValue = node.raw;
         if (path.getParentNode().kind === "encapsed") {
           quote = "";
+        } else {
+          // if this is not part of an encapsed node, we need to strip out the quotes from the raw value
+          if (['"', "'"].includes(stringValue[0])) {
+            stringValue = stringValue.substr(1);
+          }
+          if (['"', "'"].includes(stringValue[stringValue.length - 1])) {
+            stringValue = stringValue.substr(0, stringValue.length - 1);
+          }
         }
-        return makeString(stringEscape(node.value), quote, false);
+        return makeString(stringValue, quote, false);
       }
       case "number":
         return printNumber(node.value);
--- a/src/util.js
+++ b/src/util.js
@@ -20,6 +20,7 @@ function printNumber(rawNumber) {
 }
 
 function stringEscape(str) {
+  debugger;
   return str.replace(/[\n\r\t\v\f\u001b\\]/g, (character, index) => {
     switch (character) {
       case "\n":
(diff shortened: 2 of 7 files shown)


czosel:
Hmm, are you sure?

PHP Script :
```php
<?php
'You can also have embedded\n newlines in 
strings this way as it is
okay to do';
```

AST Structure :
```json
{
  "kind": "program",
  "children": [
    {
      "kind": "string",
      "value": "You can also have embedded\\n newlines in \nstrings this way as it is\nokay to do",
      "raw": "'You can also have embedded\\n newlines in \nstrings this way as it is\nokay to do'",
      "isDoubleQuote": false
    }
  ],
  "errors": []
}
```
https://github.com/glayzzle/php-parser/issues/123

mgrip:
Ya I saw the same thing when using the parser playground, but when I actually debugged in the code `node.value` was exactly the same

czosel:
Oh, ok - maybe this changed in the meantime. Using `raw` definitely sounds like a plan.

src/printer.js
```diff
@@ -477,7 +477,10 @@ function printExpression(path, options, print) {
         // @TODO: for now just reusing double/single quote preference from doc. could eventually
         // use setting from options. need to figure out how this works w/ complex strings and interpolation
         // also need to figure out splitting long strings
-        const quote = node.isDoubleQuote ? '"' : "'";
+        let quote = node.isDoubleQuote ? '"' : "'";
+        if (path.getParentNode().kind === "encapsed") {
+          quote = "";
```
czosel: How is this working? I've just looked at the implementation of `makeString`, but I don't see how passing an empty quote is affecting the escaping ...

mgrip: I don't think it would affect the escaping - but when the string is part of an encapsed node, that parent node is what's responsible for printing the actual quote. so if you had `$test = "test ${string}";` - that actually gets broken up into multiple strings. the parent node is an encapsed node (which is responsible for printing the quote), and the children are `test ` and `${string}` i think - which we wouldn't want to re-print quotes for

czosel: Right - and is the change affecting the tests in any way?

evilebottnawi:
Why `raw` sounds like a pain?

src/printer.js
mgrip: I don't think so? Other than keeping escaped line breaks escaped (which we're not currently doing)

mgrip:
@czosel @evilebottnawi switched to use `node.raw` - let me know what you think. Would now fix https://github.com/prettier/plugin-php/issues/253 as well

src/printer.js
```diff
@@ -477,8 +476,20 @@ function printExpression(path, options, print) {
...
+        } else {
+          // if this is not part of an encapsed node, we need to strip out the quotes from the raw value
+          if (['"', "'"].includes(stringValue[0])) {
+            stringValue = stringValue.substr(1);
+          }
+          if (['"', "'"].includes(stringValue[stringValue.length - 1])) {
+            stringValue = stringValue.substr(0, stringValue.length - 1);
+          }
```
czosel: I'm usually not too much into using Regexes, but here it might be shorter:
```js
if (['"', "'"].includes(stringValue[0])) {
  stringValue = str.replace(/^["'](.*)["']$/, '$1');
}
```

mgrip: 🙏 thank you - im terrible with regex haha

czosel: Me too, i basically copied this from StackOverflow :joy:

mgrip: hrm that doesn't seem to be working for me though

czosel: Hmm, looks ok to me:
```
strip = str => str.replace(/^["'](.*)["']$/, '$1')
strip("\"foo\"")
"foo"
strip("'foo'")
"foo"
strip("'foo' a")
"'foo' a"
```

src/util.js
```diff
@@ -19,6 +19,7 @@ function printNumber(rawNumber) {
   );
 }
 
+// @TODO: if we're using the "raw" value from the parser, do we need this?
```
czosel: Maybe @evilebottnawi can answer that :smile:

tests/string/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -120,11 +124,14 @@ $string = "This is the value of the var named by the return value of getName():
...
-";
+$string = "I'd like an \${beers::softdrink}\\n";
+$string = "I'd like an \${beers::$ale}\\n";
 $test = "\\\\";
 $test = '\\\\';
+printf("Query run in $queryTimeTaken seconds.\\n");
+$test =
+    "You can also have embedded\\n newlines in
+strings this way as it is
+okay to do";
```
czosel: Indentation is strange here, right? (Has not much to do with this PR though)

src/printer.js
mgrip: for whatever reason this test is failing
```php
echo "This spans
multiple lines. The newlines will be
output as well";
```

tests/string/__snapshots__/jsfmt.spec.js.snap
mgrip: agreed. i think thats just the way our `assign` node breaks though if we go over the line limit. maybe we need to think about an exception for strings or something? not sure best way to handle it

src/printer.js
czosel: Oh, it was lacking multiline support. Try `^["']([\s\S]*)["']$`

czosel: I think we're slowly reaching the point where the Regex becomes so cryptic that it's better to keep the longer, non-regex version :wink:

tests/string/__snapshots__/jsfmt.spec.js.snap
czosel: Yep, that might be an idea. Let's open a new issue about this, it's probably not very urgent.

Issue #113, opened 2018-02-23
Remove comment


mathieutu:
I think this one can be prettier hard, but Prettier should not remove comment.

```diff
-    protected $policies = [
-        // 'App\Model' => 'App\Policies\ModelPolicy',
-    ];
+    protected $policies = [];
```

evilebottnawi:
@mathieutu known issue, right now parser don't respect comments in array, you can help with this in https://github.com/glayzzle/php-parser/issues/97

mgrip:
Fixed with #265

Issue #264, opened 2018-03-29
Comments on arrays declared as const not printed


czosel:
Input:
```php
const FOO = [
    // foo
    1
];
```

Output: `Comment "// foo" was not printed. Please report this error!`

mgrip:
wow thats weird.

```php
$foo = [
    // foo
    1
];
```
works. but
```php
const FOO = [
    // foo
    1
];
```
doesn't

czosel:
Yep - took a while until I realized what the problem was 😆

PR #265, opened 2018-03-29
fix: comments for constants

mgrip:
https://github.com/prettier/plugin-php/issues/264


mgrip:
Also think this would fix https://github.com/prettier/plugin-php/issues/113

Issue #273, opened 2018-03-30
Comments in nested array definitions move and drop commas


cmancone:
You guys are going to murder me for my nested arrays :)

Input:
```php
<?php
$array = [
    'type'      => [],
    'config'    => [
        // comments in nested arrays break things
        'key' => 1,
    ],
    'accepted'  => [],
];
```
Output:
```php
<?php
$array = [
    'type' => [],
    'config' => ['key' => 1]
    // comments in nested arrays break things,
    'accepted' => []
];
```
Expected Output (probably?):
```php
<?php
$array = [
    'type' => [],
    'config' => [
        // comments in nested arrays break things
        'key' => 1
    ],
    'accepted' => []
];
```

Notice two problems: the comment moved, and also a separating comma moved to the end of the comment.  The latter of course causes a syntax error.

evilebottnawi:
Done in https://github.com/prettier/plugin-php/pull/276

PR #276, opened 2018-03-30
fix: nested array comments

mgrip:
https://github.com/prettier/plugin-php/issues/273

Also refactoring the getCommentChildNodes to make it more scalable

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -39,7 +39,15 @@ return [
   // foo
   "bar"
 ];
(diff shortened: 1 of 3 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -69,13 +77,11 @@ $array = [
     /* comment */
     /* comment */
     /* comment */
-    "foo" => "bar"
-    /* comment */
-    /* comment */, /* comment */
+    "foo" => /* comment */ /* comment */
```
mgrip: This doesn't look fantastic but I think its worth the trade-off for now of not breaking the new example

evilebottnawi: Yep, let's fix it in future :+1:

PR #287, opened 2018-03-31
Retif comments support

### LINKED ISSUE #281: Comments in anonymous function inside ternary
Gotta love those edge cases:
```php
abstract class Kernel
{
    protected function initializeContainer()
    {
        $previousHandler = previousHandler ?: set_error_handler(function () {
            $foo = "bar";
            // Clean the trace by removing first frames added by the error handler itself.
        });
    }
}
```
`[error] ../symfony/src/Symfony/Component/HttpKernel/Kernel.php: Error: Comment "// Clean the trace by removing first frames added by the error handler itself." was not printed. Please report this error!`

czosel:
Fixes #281

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -134,6 +134,10 @@ const printers = {
         {
           kinds: ["parameter"],
           children: { listNodes: [], nodes: ["value"] }
+        },
+        {
+          kinds: ["retif"],
+          children: { listNodes: [], nodes: ["trueExpr", "falseExpr", "test"] }
         }
       ];
       const children = nodeMap.reduce((currentChildren, map) => {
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -134,6 +134,10 @@ const printers = {
         {
           kinds: ["parameter"],
           children: { listNodes: [], nodes: ["value"] }
+        },
+        {
+          kinds: ["retif"],
+          children: { listNodes: [], nodes: ["trueExpr", "falseExpr"] }
```
mgrip: should probably add [test](https://github.com/glayzzle/php-parser/blob/master/docs/AST.md#retif) as well

czosel: Good catch! done.

PR #288, opened 2018-03-31
Support comments in "new" node

czosel:
Too lazy to create an issue for this :sweat_smile:


mgrip:
I don't blame you haha

PR #299, opened 2018-04-02
fix: print semicolon fi statement inside `parens`

### LINKED ISSUE #292: Semicolon missing after inline `new`
Input:
```php
(new Translator($container, new MessageFormatter(), 'en', array(), array('foo' => 'bar')));
```

Output:
```php
        (new Translator(
            $container,
            new MessageFormatter(),
            'en',
            array(),
            array('foo' => 'bar')
        ))
```

This is taken from [symfony's repo](https://github.com/symfony/symfony/blob/96bec9de77bfedadcb638a39b9b1c7505d0ae786/src/Symfony/Bundle/FrameworkBundle/Tests/Translation/TranslatorTest.php#L142).

evilebottnawi:
fixes #292

### DIFF
--- a/tests/parens/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/parens/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,31 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`new.php 1`] = `
+<?php
+(new Translator(
+    $container,
+    new MessageFormatter(),
+    'en',
+    array(),
+    array('foo' => 'bar')
+))
+?>
+<?php
+(((new Translator(
+    $container,
+    new MessageFormatter(),
+    'en',
+    array(),
+    array('foo' => 'bar')
+))))
+?>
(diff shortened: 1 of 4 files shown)


tests/parens/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,31 @@
...
+))))
+?>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+(new Translator($container, new MessageFormatter(), 'en', array(), array(
+    'foo' => 'bar'
+)));
+(new Translator($container, new MessageFormatter(), 'en', array(), array(
+        'foo' => 'bar'
+    )));
```
evilebottnawi: Something wrong with indent, let' fix it in other PR :+1:

czosel:
This is much better indeed! Does it make sense to leave out the semicolon in the test though? Usually I’d say we should test for valid PHP.

evilebottnawi:
@czosel I think it all depends on the situation, when we tests semicolon we should add tests without semicolon. in normal tests we should write semicolon always - to better reading tests

PR #301, opened 2018-04-02
fix: print comments inside incide node

evilebottnawi:
We have extra indent, let's fix in other PR

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -142,6 +142,14 @@ const printers = {
         {
           kinds: ["echo"],
           children: { listNodes: ["arguments"], nodes: [] }
+        },
+        {
+          kinds: ["parenthesis"],
+          children: { listNodes: [], nodes: ["inner"] }
+        },
+        {
+          kinds: ["include"],
+          children: { listNodes: [], nodes: ["target"] }
         }
       ];
       const children = nodeMap.reduce((currentChildren, map) => {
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -148,7 +152,12 @@ const printers = {
               return childValue ? [...accumulator, ...childValue] : accumulator;
             }, []),
             ...map.children.nodes.reduce((accumulator, childNode) => {
-              const childValue = node[childNode];
+              let childValue = node[childNode];
+              if (node.kind === "include") {
+                while (childValue && childValue.kind === "parenthesis") {
```
mgrip: this seems fishy to me - why not just add `parenthesis` to the map above?

evilebottnawi: @mgrip `parenthesis` can have other `parenthesis` and they also have `parenthesis`

mgrip: right... but as long as `parenthesis` is in the map once it should work - don't think it matters that its recursive

evilebottnawi: @mgrip i get error about comments not printed :confused:

mgrip: hrm I'm trying to check out your branch to see but I can't seem to find it...

mgrip: anyways just tried it myself and it worked

```javascript
{
  kinds: ["parenthesis"],
  children: { listNodes: [], nodes: ["inner"] }
},
{
  kinds: ["include"],
  children: { listNodes: [], nodes: ["target"] }
}
```

evilebottnawi: @mgrip your solution is better :+1: I will do this in near future :+1:

evilebottnawi:
@mgrip done :+1:

PR #307, opened 2018-04-02
fix: only format printing of comments for docblocks

mgrip:
fixes #303 fixes #268

Pretty much re-using the same logic from the js printer https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L5383

PR #308, opened 2018-04-02
fix: adding dangling comments for parens

mgrip:
see title 😝


mgrip:
will probably need to fix conflicts after https://github.com/prettier/plugin-php/pull/307

PR #316, opened 2018-04-03
fix: don't add extra newline in `offsetlookup` and add newline in `assign`

### LINKED ISSUE #312: Extra newline in array push
This
```php
<?php
$someArray[] = $this->something->long->that->makes->it->break->onto->two->lines();
```
becomes
```php
<?php
$someArray[

] = $this->something->long->that->makes->it->break->onto->two->lines();
```

which has a really strange blank line in it. I'm not sure the best way to break this. Probably something like this?
```php
<?php
$someArray[] =
  $this->something->long->that->makes->it->break->onto->two->lines();
```

evilebottnawi:
fixes #312


evilebottnawi:
WIP on newline after `assign`

evilebottnawi:
Looks like [prettier](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEASAzhAtnAggJ3wEMBPAbQEYBdAAgF4bUYALAS3QDpMcXWoBzDgBtoglkRgcsRANZxOrSQCN8cWR2gwIHGAHdtKtTI6H1QvvIAUASgDcIADQgIABxito6ZKCKEIugAVfBC8UIgA3CFYAE0cQInQYZAAzIiF0OCcVIjA5GABlFxy+fmQYfABXTJBoiDAUtIynPgz8GADifmkG9OqAK3QADwAhYly4AqIcABkLHqbnCpgXJYAmeeqi-FbkECUiJRIRKDiXfD4YAHUYlmQADgAGJzOIDMviF12z+Th8cLg4qoAI4VViqDpELpEDZODJYVhlSrVdAlIRwACKFQg8BhIBgB2u0VuSFWTnKRFY5gEAGFsN0UFBoACnBUMgAVA6hVK9AC+PKAA) don't add newline after `=`. Should we do this or not?

evilebottnawi:
We can merge this as it to avoid ugly newline inside `[]` and open issue for discussion about newline

src/printer.js
```diff
@@ -330,23 +330,26 @@ function printExpression(path, options, print) {
           path.call(print, "offset")
         ]);
       case "offsetlookup": {
-        const isOffsetNumberNode = node.offset && node.offset.kind === "number";
+        const hasOffset = node.offset;
```
czosel: If it's called `hasOffset`, it should be `Boolean(node.offset)` instead. Maybe you could just drop the extra variable and use node.offset directly.

evilebottnawi:
@czosel done :+1:

PR #339, opened 2018-04-06
fix: output comments in `print` node

evilebottnawi:
Ugly, let's fix it in other PR


evilebottnawi:
Merge as is, just couple of lines and tests. Feel free to feedback :+1:

PR #341, opened 2018-04-06
fix: output comments in `global` node

evilebottnawi:


evilebottnawi:
Merge as is, just couple of lines and tests. Feel free to feedback :+1:

Issue #347, opened 2018-04-06
Potential nodes print without comments.


evilebottnawi:
Need add tests for each
- [x] interface
- [x] exit
- [x] goto
- [x] silent
- [x] halt
- [x] break
- [x] continue
- [x] traituse
- [x] traitprecedence
- [x] traitalias
- [x] label

For many nodes i already send PR. I think list above is last. When we should improve ugly output :+1:

evilebottnawi:
Cross issue https://github.com/prettier/prettier/issues/4276

evilebottnawi:
Information: strange with
```php
exit(/* Comment */);
```

mgrip:
🎉 nice work @evilebottnawi

evilebottnawi:
Looks like all nodes print comments, but i will add more tests, example around empty arguments in function  and double check every node.
@mgrip Thanks!

PR #349, opened 2018-04-06
fix: output comments in trait stuff

evilebottnawi:


evilebottnawi:
partial fix https://github.com/prettier/plugin-php/issues/347

PR #350, opened 2018-04-06
Fix escaping in encapsed nodes

### LINKED ISSUE #298: Quote escaping removed
Input:
```php
$test = "\"$commandName\"";
```
Output:
```php
$test = ""$commandName"";
```

czosel:
Fixes #298

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -495,18 +495,14 @@ function printExpression(path, options, print) {
         // @TODO: for now just reusing double/single quote preference from doc. could eventually
         // use setting from options. need to figure out how this works w/ complex strings and interpolation
         // also need to figure out splitting long strings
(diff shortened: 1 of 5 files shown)


src/printer.js
```diff
@@ -524,9 +524,7 @@ function printExpression(path, options, print) {
             path.map(valuePath => {
               const node = valuePath.getValue();
               if (node.kind === "string") {
-                return valuePath.getParentNode().type === "heredoc"
```
mgrip: you can probably get rid of some extra logic under "string" now that we aren't printing this

czosel: Good catch! Please check again 😄

mgrip:
🎉

PR #353, opened 2018-04-07
fix: output comments in `silent` node

evilebottnawi:


evilebottnawi:
Feel free to feedback

PR #355, opened 2018-04-07
fix: output comments in `break` and `continue` node

evilebottnawi:
:star:


evilebottnawi:
Merge as is, feel free to feedback :+1:

PR #148, opened 2018-03-03
Basic chaining support

czosel:
This is basically a simplified copy of what is done in the JS printer (see [here](https://github.com/prettier/prettier/blob/2e59f178aa174b35989d6e1a168032a48626dfc6/src/language-js/printer-estree.js#L3886))

Fixes #109 
Fixes #116

PR #372, opened 2018-04-10
fix: newlines in lookup nodes

### LINKED ISSUE #206: Chaining doesn't work always
This should be picked up by the chaining printer:
```php
$this->loooooooooooong->lookup = (int) $this->getRequest()->getParam(
   'some-param'
);
```
but it seems that it's printed the same way as shown.
(This came up in #193)

evilebottnawi:
fixes #206

Notes:
1. It is port logic from `prettier`
2. Parser have problem with `break` node in `propertylookup` https://github.com/glayzzle/php-parser/issues/148
3. Related issue around `propertylookup` and `offsetlookup` https://github.com/prettier/plugin-php/issues/373
4. Multiple function arguments in call https://github.com/prettier/plugin-php/issues/374
5. Move `chain` test from `functions` to `member-chain` test
6. Issue around parens https://github.com/prettier/plugin-php/issues/375


tests/yield/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -42,8 +42,10 @@ function gen_one_to_three()
 {
     for ($i = 1; $i <= 3; $i++) {
         yield $veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongValues;
-        yield $veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongObjName->veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongMethod();
-        yield from $veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongObjName->veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongMethod();
+        yield $veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongObjName->
```
evilebottnawi: Looks better :+1:

tests/offsetlookup/offsetlookup.php
```diff
@@ -7,4 +7,4 @@
 $variable[1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111][1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111] = html_entity_decode(wp_get_document_title());
 $variable['veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongString'] = $variable->registered['veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongString'];
 $variable = get_taxonomy_function(get_queried_object_function()->taxonomy)->object_type[0];
-$someArray[] = $this->something->long->that->makes->it->break->onto->two->break->break->lines();
+$someArray[] = $this->something->long->that->makes->it->break->onto->two->bbreak->bbreak->lines();
```
evilebottnawi: Comment around issue above

tests/member_chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,389 @@
...
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+function call($store)
+{
+    return callApi($endpoint, $schema)->then(function ($response) {
+        return next(
+            actionWith(['response' => $response, 'type' => $successType])
+        );
+    }, function ($error) {
```
evilebottnawi: Should we start with newlines when we have two function? But not related to this issue. Feel free create issue around this,

tests/member_chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,389 @@
...
+
+($a ? $b : $c)->d()->e();
+
+($a ? $b : $c)
+    ->d()
+    ->e()
+    ->f();
+
+(
+    $valid
```
evilebottnawi: WIP

tests/member_chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,389 @@
...
+    ->find('SomewhatLongNodeName')
+    ->prop('longPropFunctionName')
+    ->then(function () {
+        doSomething();
+    });
+
+$wrapper
+    ->find('SomewhatLongNodeName')
+    ->
+        prop(
```
evilebottnawi: WIP

tests/new/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -53,13 +53,14 @@ abstract class A
     }
 }
 
-$class = (new Foo())->veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongMethod();
+$class = (new Foo())->
```
evilebottnawi: Should we handle `new`, `clone` as `$this`?

tests/propertylookup/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -37,16 +51,29 @@ $this->loooooooooooong->loooooooooooong->loooooooooooong->lookup =
...
 
 `;
 
 exports[`single.php 1`] = `
 <?php
-$obj->veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongProperty;
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+$obj->veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongProperty;~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
 $obj->
```
evilebottnawi: Should be printed in one line. WIP

tests/new/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: Also feel free create new issue around this problem. PR already is big, let's do it in other PR :+1:

tests/propertylookup/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: Not related to `call` node, let's do this in other PR

evilebottnawi: Issue: https://github.com/prettier/plugin-php/issues/373

tests/member_chain/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: https://github.com/prettier/plugin-php/issues/374

tests/functions/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -33,7 +33,8 @@ one()
     ->two->three()
     ->four->five();
 
-Route::prefix('api')
+Route
+    ::prefix('api')
```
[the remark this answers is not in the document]
vjeux: Js doesn’t have :: so the question didn’t happen. It does look a bit weird, but I haven’t written any php in a while so i’m not really sure what’s the way people format code like this

tests/member_chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,420 @@
...
+$a->b->a();
+$a
+    ->b()
+    ->c()
+    ->d();
+$a->b->c->d;
+
+$this->loooooooooooong->lookup = (int) $this->getRequest()->getParam(
+    'some-param'
+);
```
evilebottnawi: WIP

tests/functions/__snapshots__/jsfmt.spec.js.snap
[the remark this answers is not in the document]
evilebottnawi: @vjeux babylon parse this as `BindExpression`, it is proposed spec. I think better print as we do. And maybe in future fix it in `js` (if spec will be accept).

tests/member_chain/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: Prettier have difference logic for `this` and regural `variable`. Should we do this as `prettier` or no?

tests/functions/__snapshots__/jsfmt.spec.js.snap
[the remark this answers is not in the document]
vjeux: First, no one uses it yet so it hasn't really been optimized. But more important, it doesn't have the same semantics as in PHP. In PHP you access static property of an object, in JS the left part is the context the right part is going to be called with. They have drastically different meanings so are likely going to be used in different contexts and should likely be printed differently.

tests/member_chain/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: Not related to this PR. Issue https://github.com/prettier/plugin-php/issues/375

src/comments.js
```diff
@@ -183,9 +183,19 @@ function printDanglingComments(path, options, sameIndent, filter) {
   return indent(concat([hardline, join(hardline, parts)]));
 }
 
+function hasLeadingComment(node) {
+  return node.comments && node.comments.some(comment => comment.leading);
+}
+
+function hasTrailingComment(node) {
```
mgrip: i think we had a "hadDanglingComments" in `comments.js` - should we move it here instead?

src/printer.js
```diff
@@ -111,12 +118,52 @@ function genericPrint(path, options, print) {
...
+      : "",
+    "]"
+  ]);
+}
+
+// We detect calls on member expressions specially to format a
+// common pattern better. The pattern we are looking for is this:
+//
+// arr
+//   .map(x => x + 1)
```
mgrip: this should be php?

src/printer.js
```diff
@@ -111,12 +118,52 @@ function genericPrint(path, options, print) {
...
+//   .filter(x => x > 10)
+//   .some(x => x % 2)
+//
+// The way it is structured in the AST is via a nested sequence of
+// MemberExpression and CallExpression. We need to traverse the AST
```
mgrip: change to match our ast node kinds

src/printer.js
```diff
@@ -133,31 +180,74 @@ function printMemberChain(path, options, print) {
...
+        originalText,
+        nextCharIndex + 1,
+        options
+      );
+    }
+
+    return isNextLineEmpty(originalText, node, options);
+  }
+
+  function rec(path) {
```
mgrip: use a more descriptive name

src/printer.js
```diff
@@ -133,31 +180,74 @@ function printMemberChain(path, options, print) {
   //    PropertyLookup d, Call]
   const printedNodes = [];
 
-  function traverse(path) {
+  // Here we try to retain one typed empty line after each call expression or
```
mgrip: im not really sure what this function is doing from this comment - can you explain a little more? maybe an example would be helpful

src/printer.js
```diff
@@ -111,12 +118,52 @@ function genericPrint(path, options, print) {
...
+//
+// The way it is structured in the AST is via a nested sequence of
+// MemberExpression and CallExpression. We need to traverse the AST
+// and make groups out of it to print it in the desired way.
 function printMemberChain(path, options, print) {
-  // First step: Linearize and reorder the AST.
+  // The first phase is to linearize the AST by traversing it down.
   //
   // Example:
   //   a()->b->c()->d()
```
mgrip: change the comment below to use our ast kinds

src/printer.js
```diff
@@ -111,12 +118,52 @@ function genericPrint(path, options, print) {
   ]);
 }
 
-// special layout for "chained" function calls, i.e.
-// $foo = a()
-//     ->b()
-//     ->c();
+function printPropertyLookup(path, options, print) {
```
mgrip: should these be moved inside `printMemberChain`?

src/printer.js
```diff
@@ -133,31 +180,74 @@ function printMemberChain(path, options, print) {
...
+      path.call(what => rec(what), "what");
     } else {
       printedNodes.unshift({
         node,
         printed: path.call(print)
       });
     }
   }
-  traverse(path);
+  // Note: the comments of the root node have already been printed, so we
```
mgrip: this comment doesn't really make sense - can you clarify?

src/printer.js
```diff
@@ -173,26 +263,92 @@ function printMemberChain(path, options, print) {
...
-  let currentGroup = [];
-  let hasSeenCall = false;
-  printedNodes.forEach(printedNode => {
-    if (hasSeenCall && printedNode.node.kind === "propertylookup") {
-      groups.push(currentGroup);
-      currentGroup = [printedNode];
-      hasSeenCall = false;
+  let currentGroup = [printedNodes[0]];
+  let i = 1;
+  for (; i < printedNodes.length; ++i) {
```
mgrip: what was wrong with foreach? (you could still just use the index from the forEach)

src/printer.js
```diff
@@ -173,26 +263,92 @@ function printMemberChain(path, options, print) {
...
     } else {
-      currentGroup.push(printedNode);
+      break;
     }
-
-    if (printedNode.node.kind === "call") {
-      hasSeenCall = true;
+  }
+  if (printedNodes[0].node.kind !== "call") {
+    for (; i + 1 < printedNodes.length; ++i) {
```
mgrip: forEach? the iterator here is confusing

src/printer.js
```diff
@@ -173,26 +263,92 @@ function printMemberChain(path, options, print) {
...
+      }
     }
-  });
+  }
   groups.push(currentGroup);
+  currentGroup = [];
+
+  // Then, each following group is a sequence of propertylookup followed by
+  // a sequence of call. To compute it, we keep adding things to the
+  // group until we has seen a call in the past and reach a
```
mgrip: have seen

src/printer.js
```diff
@@ -173,26 +263,92 @@ function printMemberChain(path, options, print) {
...
+  // propertylookup
+  let hasSeenCallExpression = false;
+  for (; i < printedNodes.length; ++i) {
```
mgrip: forEach

src/printer.js
```diff
@@ -173,26 +263,92 @@ function printMemberChain(path, options, print) {
...
+    }
+
+    if (printedNodes[i].node.kind === "call") {
+      hasSeenCallExpression = true;
+    }
+    currentGroup.push(printedNodes[i]);
+
+    if (
+      printedNodes[i].node.comments &&
+      printedNodes[i].node.comments.some(comment => comment.trailing)
```
mgrip: can you use the `hasTrailingComment` helper you added?

src/printer.js
```diff
@@ -173,26 +263,92 @@ function printMemberChain(path, options, print) {
...
+      groups.push(currentGroup);
+      currentGroup = [];
+      hasSeenCallExpression = false;
+    }
+  }
+  if (currentGroup.length > 0) {
+    groups.push(currentGroup);
+  }
+
+  const shouldMerge =
```
mgrip: this is a lot haha - can we add a comment outlining what rules we're defining?

src/printer.js
```diff
@@ -203,15 +359,68 @@ function printMemberChain(path, options, print) {
   }
 
   const printedGroups = groups.map(printGroup);
-  if (groups.length <= 2) {
-    return group(concat(printedGroups));
+  const oneLine = concat(printedGroups);
+
+  const cutoff = shouldMerge ? 3 : 2;
```
mgrip: not sure what this is doing - can you add a comment explaining?

src/printer.js
```diff
@@ -203,15 +359,68 @@ function printMemberChain(path, options, print) {
...
+  const flatGroups = groups
+    .slice(0, cutoff)
+    .reduce((res, group) => res.concat(group), []);
+
+  const hasComment =
+    flatGroups
+      .slice(1, -1)
+      .some(node => comments.hasLeadingComment(node.node)) ||
```
mgrip: can you use the helpers you added?

src/printer.js
```diff
@@ -203,15 +359,68 @@ function printMemberChain(path, options, print) {
...
+    flatGroups
+      .slice(0, -1)
+      .some(node => comments.hasTrailingComment(node.node)) ||
+    (groups[cutoff] && comments.hasLeadingComment(groups[cutoff][0].node));
+
+  // If we only have a single `.`, we shouldn't do anything fancy and just
```
mgrip: "->" ?

src/printer.js
```diff
@@ -133,31 +180,74 @@ function printMemberChain(path, options, print) {
...
+          shouldInsertEmptyLineAfter(node) ? hardline : ""
+        ])
       });
-      path.call(what => traverse(what), "what");
-    } else if (node.kind === "propertylookup") {
+      path.call(what => rec(what), "what");
+    } else if (isMemberish(node)) {
+      let printedMemberish = null;
+      if (node.kind === "propertylookup") {
+        printedMemberish = printPropertyLookup(path, options, print);
```
mgrip: maybe add a comment explaining why we do this instead of just calling `print` - im assuming its to avoid grouping?

src/printer.js
czosel: I'm having a deja-vu :laughing: 
https://github.com/prettier/plugin-php/pull/148#discussion_r172655593

src/comments.js
evilebottnawi: @mgrip I think all related to comments better have in `comments.js` :smile:

src/printer.js
evilebottnawi: @mgrip :+1: let's fix it :+1:

src/printer.js
evilebottnawi: let's use `traverse` :+1:

src/printer.js
evilebottnawi: @mgrip hm, for me everything is clear, whatever you comment added?

src/comments.js
mgrip: fine by me! should we move the `hasDanglingComments` helper in here then?

src/printer.js
mgrip: `"propertylookup", "staticlookup"`, etc are handled by `print` already - so whats the benefit to calling `printPropertyLookup` instead of just `path.call(print, "offset")`?

evilebottnawi: @mgrip better output method/property chains :smile:

src/printer.js
evilebottnawi: @mgrip `forEach` is not optimize for this situation, we don't start with `0` value and it is create more memory and cpu usage, also i do not see any problems with reading this

src/printer.js
mgrip: ya so maybe a comment saying how its different and why its better 🤷‍♂️

src/printer.js
mgrip: I thought `forEach(callback(item, index) {})` just uses the index, which would start at `0`? you're not resetting `i` from the previous loop, wouldn't that cause a problem?

src/printer.js
evilebottnawi: @mgrip forEach don't have `break`, need use `for` for this case.

src/printer.js
evilebottnawi: @mgrip it is expected behaviour, we break grouping when it is not call (i.e. `$a[1][2][3]->property` should be in one line, reseting `i` break property on newline and it is ugly).

src/printer.js
evilebottnawi: @mgrip no, we use this function inside in `printMemberChain` and output `propertylookup`

src/printer.js
evilebottnawi: @mgrip right now we don't print comments on 100% as prettier, it is for future, i remove this comment from this PR and add this in future again with example

src/printer.js
evilebottnawi: @mgrip expected behavior, i will write about above

tests/member_chain/__snapshots__/jsfmt.spec.js.snap
czosel: Hmm, my first intuition is that in PHP `$this` is not as special as `this` in JS. Do you have a concrete case that would call for different logic?

evilebottnawi:
@mgrip @czosel can we merge this and continue to improve? this PR blocked some improvements, to avoid merge

mgrip:
looks good to me! This is a pretty big one so @czosel you might wanna at least glance through if you haven't already

czosel:
Unfortunately I don't have time for a proper review - but I did skim over the changes, and it looks really good! :+1: 

Basically, what we had before was also a port of the JS printer, but I probably simplified it a bit too much. From the diff it's quite hard to see what was missing exactly in the previous implementation, but the tests seem to show that we're making a step forward with this, so I totally support merging it! :smile:

Issue #390, opened 2018-04-13
Formating causes wrong indent


sseypt:
Hello.

Input:
```php
<?php
if ($parity != array_reduce($r,function ($acc, $tmp) {return $acc ^ $tmp;},0)) {
    return [];
}

```

Output:
```php
<?php
if ($parity != array_reduce(
        $r,
        function ($acc, $tmp) {
            return $acc ^ $tmp;
        },
        0
    )) {
    return [];
}

```

Expected:
```php
<?php
if ($parity != array_reduce(
    $r,
    function ($acc, $tmp) {
        return $acc ^ $tmp;
    },
    0
)) {
    return [];
}

```

Best regards,
Steve

sseypt:
Another example:

Input:
```php
<?php
implode(
    ', ',
    array_map(
        function ($f) {
            return $f;
        },
        array_merge($arr1, $arr2, $arr3)
    )
);

```

Output:
```php
<?php
implode(', ', array_map(function ($f) {
    return $f;
}, array_merge($arr1, $arr2, $arr3)));

```

Expected:
```php
<?php
implode(
    ', ',
    array_map(function ($f) {
        return $f;
    }, array_merge($arr1, $arr2, $arr3))
);

```

evilebottnawi:
1. Should be https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAZgAgBQBIAOAhgE6owCemAhALyYnGHkD6xcAJgK5hzYA6UTEOFDcxADQCRI9JyhgYqaDlyEwYcZlwwAtvgCUmYFOmm2MTsUGr1mAHpbd+ANwnTAX0mDTmAAxvMfUNjbyFzS0EAbQBdVyh3EHEQCHxFaABnZFAGCAB3AAUSBEyUQgA3CFR2RJBCdJhkdEIAG3S4JIAjRjAAazgYAGUiMFQoAHNkGGJOdpB2CDBGlrak0bbiGHzGMZ1CJdbZgCt0gA8AIW6+wcIdOAAZUbh9leTOGHw3gCZn2aJideQIA6hA65Ga0Bq+FIsAA6lUYAALZAADl8SShEDaMMY+EBULg6zKTySbAAjpxUGwtoQdnskE0Dkk2jpUJNprN0qMxs04ABFTgQeA-JIwEFw9iI5CfEWMVDNLkAYQgOl2gKg0GJIE4bQAKiCSgy2u53EA
[...]

PR #398, opened 2018-04-16
fix: printArgumentsList

mgrip:
https://github.com/prettier/plugin-php/issues/390

pretty much copying logic from prettier - I checked all of the updated snapshots and they matched prettierjs output

### DIFF
--- a/tests/call/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/call/__snapshots__/jsfmt.spec.js.snap
@@ -108,6 +108,20 @@ $packages = array_merge(
     idx($composer, 'require', []),
     idx($composer, 'require-dev', [])
 );
+
+if ($parity != array_reduce($r,function ($acc, $tmp) {return $acc ^ $tmp;},0)) {
+    return [];
+}
+
+implode(
+    ', ',
+    array_map(
+        function ($f) {
+            return $f;
+        },
+        array_merge($arr1, $arr2, $arr3)
+    )
+);
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
 func();
@@ -128,14 +142,17 @@ $db->Execute($sql, [
     $somewhatLongParameterXYZ
 ]);
 
-$db->Execute([
-    $foo,
-    $bar,
-    $foobar,
-    $somewhatLongParameter,
-    $somewhatLongParameterX,
-    $somewhatLongParameterXYZ
-], $sql);
+$db->Execute(
+    [
+        $foo,
+        $bar,
+        $foobar,
+        $somewhatLongParameter,
+        $somewhatLongParameterX,
+        $somewhatLongParameterXYZ
+    ],
+    $sql
+);
 
 $app->get('/hello/{name}', function ($name) use ($app) {
     return 'Hello ' . $app->escape($name);
@@ -178,15 +195,10 @@ $this->filter([
     new MergeValue(['Abigail', 'Lydia'])
 ]);
 
-$this->assertEquals([
-    'First',
-    'Second',
-    'Taylor',
-    'Mohamed',
-    'Jeffrey',
-    'Abigail',
-    'Lydia'
-], $results);
+$this->assertEquals(
+    ['First', 'Second', 'Taylor', 'Mohamed', 'Jeffrey', 'Abigail', 'Lydia'],
+    $results
+);
 
 $this->assertEquals(['First'], $results);
 
@@ -218,6 +230,26 @@ $foo->bar(
 $packages = array_merge(
     idx($composer, 'require', []),
     idx($composer, 'require-dev', [])
+);
+
+if (
+    $parity !=
+    array_reduce(
+        $r,
+        function ($acc, $tmp) {
+            return $acc ^ $tmp;
+        },
+        0
+    )
+) {
+    return [];
+}
+
+implode(
+    ', ',
+    array_map(function ($f) {
+        return $f;
+    }, array_merge($arr1, $arr2, $arr3))
 );
 
 `;
--- a/tests/class/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/class/__snapshots__/jsfmt.spec.js.snap
@@ -319,8 +319,8 @@ abstract class ReallyReallyReallyLongClassName
     {
         parent::__construct($test_int ?: 1);
         $this->other = $test_string;
-        $this->current_version =
-            $current_version ?: new Content_Version_Model();
+        $this->current_version = $current_version
+            ?: new Content_Version_Model();
         self::$staticTest = $test_int;
     }
 
--- a/tests/class/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/class/__snapshots__/jsfmt.spec.js.snap
@@ -47,13 +47,15 @@ class Application
 }
 
 $app = new Application();
-$app->setLogger(new class implements Logger
-{
-    public function log(string $msg)
+$app->setLogger(
+    new class implements Logger
     {
-        echo $msg;
+        public function log(string $msg)
+        {
+            echo $msg;
+        }
     }
-});
+);
 
 `;
 
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -217,10 +217,14 @@ exports[`call.php 1`] = `
 foo(/* Comment */ $a /* Comment */, /* Comment */ [] /* Comment */, /* Comment */ ($a + 2) /* Comment */);
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
-foo(/* Comment */
-$a, /* Comment */ /* Comment */
-[], /* Comment */ /* Comment */
-($a + 2)); /* Comment */
+foo(
+    /* Comment */
+    $a, /* Comment */
+    /* Comment */
+    [], /* Comment */
+    /* Comment */
+    ($a + 2) /* Comment */
+);
 
 `;
 
@@ -626,7 +630,9 @@ do {
     // some comment
 } while (true);
 
-$test = foo ?: bar(function () {
+$test =
+    foo
+    ?: bar(function () {
         $foo = "bar";
         // test
     });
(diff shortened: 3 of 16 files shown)


mgrip:
For reference: https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L3180

tests/parens/new.php
```diff
@@ -8,11 +8,20 @@
 ))
 ?>
 <?php
-(((new Translator(
```
evilebottnawi: Let's leave this test as is with multiple parens

tests/encapsed/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -117,9 +117,11 @@ $code = <<<EOF
 \\$kernel = unserialize('$kernel');
 \\$request = unserialize('$request');
 EOF;
-$content = sprintf(<<<EOF
+$content = sprintf(
+    <<<EOF
```
evilebottnawi: Break code?

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -626,7 +630,9 @@ do {
     // some comment
 } while (true);
 
-$test = foo ?: bar(function () {
+$test =
+    foo
+    ?: bar(function () {
```
evilebottnawi: Ugly? Need link how prettier do this :smile:

tests/call/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -128,14 +142,17 @@ $db->Execute($sql, [
...
+    [
+        $foo,
+        $bar,
+        $foobar,
+        $somewhatLongParameter,
+        $somewhatLongParameterX,
+        $somewhatLongParameterXYZ
+    ],
+    $sql
+);
```
evilebottnawi: Looks PSR-12 say what we can don't write each argument on newline. Or i am wrong?

tests/encapsed/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -117,9 +117,11 @@ $code = <<<EOF
...
 foo
 EOF
+
```
czosel: This newline is unexpected though, right?

tests/comments/__snapshots__/jsfmt.spec.js.snap
mgrip: `?:` isn't valid JS sytax, but [this](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEASeBnGACAvNgMwgmwH5sYAnAVzmyWwCMBDSgCgOqjBgEtpsbAJTZgAHSjYp2VERL4xIFpUUBuCdOwB6LRThYJAXyGqQAGhAQADn2gZkoVpQgB3AAqsE9lMwBuEXgATcxBmLGQCZgAbDDgLRkpmMABrOBgAZSsk3igAc2QqWgtAiDAI6NiLHNjKGDdE3IBbZnKYuJAAKwwADwAhRJS09OZGuAAZHLhWystqGCs5gCZp9qzKGuQlZkYATyjoEKtKHJgAdSCYAAtkAA4ABgsjiFjTxKtNo-04Sl8pi0o4ABHai8AH1ZhNFpISJtCyxRq8Ao0doYHK5KJwACK1Ag8BWFhg23OgSuyEWBMSvCiaIAwhBGs1NlBoH8QNRYgAVbbeGGxQyGIA) is the closest I think - I can look to see why `foo` isn't on the 1st line

tests/call/__snapshots__/jsfmt.spec.js.snap
mgrip: im not sure, I couldn't find anything... do you know where it says that? This is how prettierjs outputs it

czosel: > Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line.

> When the argument list is split across multiple lines, the closing parenthesis and opening brace MUST be placed together on their own line with one space between them.

https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#45-method-and-function-arguments

Seems fine to me.

tests/encapsed/__snapshots__/jsfmt.spec.js.snap
mgrip: correct - not sure why its happening though... I can try and investigate

czosel: There is some special logic that adds linebreaks after heredoc/nowdoc nodes, because before we had issues with code like `EOF;` or `EOF]` which would break.

czosel:
All in all this looks great! :100:

mgrip:
also fixing `retif` formatting

tests/comments/__snapshots__/jsfmt.spec.js.snap
mgrip: updated - let me know what you think. Basically stripped out some extra logic we had for the "null if" case, I _think_ this looks better

czosel: :ok_hand: Let's go with that!

mgrip:
@evilebottnawi @czosel ready for review again 👍

tests/encapsed/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -121,8 +121,7 @@ $content = sprintf(
     <<<EOF
 foo
 EOF
-
-);
+    );
```
czosel: Now the indent seems off 🙈

mgrip: ugh 😕 thought this might be how js does it but apparently [not](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEASSsEwAQF5tgCGANsQCpwDOMAFADpTZNMAGAyhALZzbzXYQAZg2LQA5tmoAnAJawWASgDcIADQgIABxgzolZKEJSpEAO4AFIwn0pCANwgyAJmpCFqyQSUpx1AIylCMABrOBg2TSC5MWQYKQBXXxAnCDBPbyS5HykYc0CxTkJ04h91ACtKAA8AIUCQsLZCbgAZOThi0o14mE1ugCYOpMipbOQQP0I-AE9RKFdNWVgAdWcYAAtkAA4ABnUFiB8lwM0xhao4KTt29Sk4AEd4mVu8wgKipC8SpJ9OGViE77RYhwACK8Qg8EG6hgkxWTnWyD60MCMmI0QAwlxCmMoNBriB4j4yJMbJ8fABfclAA)

mgrip: ahh ok that makes sense - its because the newline is getting added in the heredoc node rather than here. I think we need to invert that logic, i'll try poking around some more

mgrip: 👌 should be good now

tests/call/__snapshots__/jsfmt.spec.js.snap
nicoder: maybe @evilebottnawi was thinking of this sentence: 

> A single argument being split across multiple lines (as might be the case with an anonymous function or array) does not constitute splitting the argument list itself.

in:

https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#47-method-and-function-calls

czosel: You're right @nicoder - I think this is at least part of the reason our first implementation was a bit different than the one in JS.

tests/class/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -317,8 +319,8 @@ abstract class ReallyReallyReallyLongClassName
     {
         parent::__construct($test_int ?: 1);
         $this->other = $test_string;
-        $this->current_version =
-            $current_version ?: new Content_Version_Model();
+        $this->current_version = $current_version
+            ?: new Content_Version_Model();
```
evilebottnawi: Can we add test when in it on one line?

tests/retif/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -39,29 +39,26 @@ return $test ? doSomething() : doSomethingElse();
...
 
 $var = $someOtherReallyReallyLongVariable
-? $someOtherReallyReallyLongVariable
-? $someOtherReallyReallyLongVariable
-: $someOtherReallyReallyLongVariable
-: $someOtherReallyReallyLongVariable
+    ? $someOtherReallyReallyLongVariable
+        ? $someOtherReallyReallyLongVariable
+        : $someOtherReallyReallyLongVariable
+    : $someOtherReallyReallyLongVariable
```
evilebottnawi: Can we add test with parens, to ensure all print pretty?

tests/call/__snapshots__/jsfmt.spec.js.snap
czosel: @mgrip What'

mgrip: i interpret that as you _can_ split an argument across multiple lines while keeping the others on a single line, but also

> Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line

From reading the psr12 spec im not seeing that there's any rhyme or reason to when you split vs keep the arguments on the same line (which is good because imo we should follow the prettierjs logic)

tests/class/__snapshots__/jsfmt.spec.js.snap
mgrip: I think we already have a test where its on 1 line? https://github.com/prettier/plugin-php/blob/master/tests/retif/__snapshots__/jsfmt.spec.js.snap#L29

tests/call/__snapshots__/jsfmt.spec.js.snap
mgrip: ie by psr12 standards i think both

```php
$db->Execute([
    $foo,
    $bar,
    $foobar,
    $somewhatLongParameter,
    $somewhatLongParameterX,
    $somewhatLongParameterXYZ
], $sql, $someOtherArg);
```

_and_

```php
$db->Execute(
  [
      $foo,
      $bar,
      $foobar,
      $somewhatLongParameter,
      $somewhatLongParameterX,
      $somewhatLongParameterXYZ
  ],
  $sql,
  $someOtherArg
);
```
 would be acceptable

evilebottnawi: @mgrip can we open issue about this for discussion?

czosel: Some of the examples mentioned in the style guides seem to prefer the "shorter" version, but I agree that the "longer" style doesn't seem to be explicitly forbidden. Probably it's the easiest to stay close to the JS implementation and move forward like this?

Issue #323, opened 2018-04-03
Comment position shuffling in function parameters


cmancone:
I saw some related issues, so apologies if this is a duplicate, but it looks a bit different so I figured I'd make sure you have everything.  Probably the best solution is that if code is on a line by itself and has a trailing comment, it should probably be forced to remain on a line by itself.

Note that the generated code is still the same - it is just debug-check that is complaining because of moving comments.

Input
```php
<?php
class add_vendor extends request{
    protected function request(){
        $qb_vendor->setVendorAddress(
            $vendor->address_1, // address 1
            $vendor->address_2, // address 2
            '', // address 3
            '', // address 4
            '', // address 5
            $vendor->city, // city
            $vendor->state, // state
            $vendor->zip // state
        );
    }
}
```
Output:
```php
<?php
class add_vendor extends request
{
    protected function request()
    {
        $qb_vendor->setVendorAddress($vendor->
            address_1, $vendor->address_2, // address 2 // address 1
        '', // address 3
        '', // address 4
        '', $vendor->city, $vendor->state, $vendor->zip); // address 5 // city // state // state
    }
}
```
It actually has less trouble when long lines force it to leave code as-is.  This works fine:

Input:
```php
<?php
class add_vendor extends request{
    protected function request(){
        $qb_vendor->setVendorAddress(
            'a really long line that will further confuse the issue', // address 1
            'also another really long line just because :)', // address 2
            '', // address 3
            '', // address 4
            '', // address 5
            $vendor->city, // city
            $vendor->state, // state
            $vendor->zip // state
        );
    }
}
```
Output:
```php
<?php
class add_vendor extends request
{
    protected function request()
    {
        $qb_vendor->setVendorAddress(
            'a really long line that will further confuse the issue',
            // address 1
            'also another really long line just because :)',
            // address 2
            '',
            // address 3
            '',
            // address 4
            '', // address 5
            $vendor->city, // city
            $vendor->state, // state
            $vendor->zip // state
        );
    }
}
```

mgrip:
fixed in https://github.com/prettier/plugin-php/pull/405

PR #405, opened 2018-04-26
fix: inline comments

mgrip:
https://github.com/prettier/plugin-php/issues/323

For reference this is a workaround to temporarily solve https://github.com/glayzzle/php-parser/issues/155

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -13,7 +13,17 @@ function parse(text) {
     }
   });
 
-  return parser.parseCode(text);
+  const ast = parser.parseCode(text);
+  // https://github.com/glayzzle/php-parser/issues/155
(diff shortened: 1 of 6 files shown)


src/parser.js
```diff
@@ -13,7 +13,17 @@ function parse(text) {
     }
   });
 
-  return parser.parseCode(text);
+  const ast = parser.parseCode(text);
+  // https://github.com/glayzzle/php-parser/issues/155
+  // currently inline comments include the line break at the end, we need to
+  // strip those out and update the end location for each comment manually
+  ast.comments.forEach(comment => {
+    if (comment.value[comment.value.length - 1] === "\n") {
```
czosel: Does node 6 already support `String.endsWith`?

Issue #394, opened 2018-04-14
Respect indent of opening <?php tag for multiline expressions


adamroyle:
First of all this plugin is amazing! Thank you. Works great for our standard php class files.

Couple of weird behaviors for our template files (we use CakePHP).

Input:
```php
<section class="<?= $sectionClass ?>">
  <header class="<?= $headerClass ?>">
    <?php

      echo $this->element('collection/item-title', [
        'class' => $itemTitleClass,
        'small_title' => $item->small_title,
        'title' => $item->title,
        'url' => $url,
        'color' => $item->title_color ?: 'primary',
      ]);
    
    ?>
  </header>
</section>
```

Output:
```php
<section class="<?php echo $sectionClass; ?>">
  <header class="<?php echo $headerClass; ?>">
    <?php echo $this->element(
  'collection/item-title',
  [
    'class' => $itemTitleClass,
    'small_title' => $item->small_title,
    'title' => $item->title,
    'url' => $url,
    'color' => $item->title_color ?: 'primary',
  ]
); ?>
  </header>
</section>
```

Expected (Option 1):
- indent multiline code block to match `<?php` open tag
- remove unnecessary breaks for function arguments (works as expected in plain php)
```php
<section class="<?php echo $sectionClass; ?>">
  <header class="<?php echo $headerClass; ?>">
    <?php echo $this->element('collection/item-title', [
      'class' => $itemTitleClass,
      'small_title' => $item->small_title,
      'title' => $item->title,
      'url' => $url,
      'color' => $item->title_color ?: 'primary',
    ]); ?>
  </header>
</section>
```

Expected (Option 2):
- same as option 1, BUT put `<?php` and `?>` on their own line, separated by an empty line 
```php
<section class="<?php echo $sectionClass; ?>">
  <header class="<?php echo $headerClass; ?>">
    <?php

    echo $this->element('collection/item-title', [
      'class' => $itemTitleClass,
      'small_title' => $item->small_title,
      'title' => $item->title,
      'url' => $url,
      'color' => $item->title_color ?: 'primary',
    ]);

    ?>
  </header>
</section>
```

Expected (Option 3):
- same as option 1, BUT use short open tag for single statement echo calls (breaks PHP < 5.4 if short open tag disabled)
```php
<section class="<?= $sectionClass ?>">
  <header class="<?= $headerClass ?>">
    <?= $this->element('collection/item-title', [
      'class' => $itemTitleClass,
      'small_title' => $item->small_title,
[...]

PR #407, opened 2018-04-27
[WIP] fix: inline alignment formatting

mgrip:
https://github.com/prettier/plugin-php/issues/394

### DIFF
--- a/src/clean.js
+++ b/src/clean.js
@@ -119,6 +119,11 @@ function clean(node, newObj) {
     });
   }
 
+  // for inline code, we always use the long form <?php
+  if (node.kind === "echo") {
+    newObj.shortForm = false;
+  }
+
   // Ignore `parenthesis` for `include`
   if (node.kind === "include" && node.target.kind === "parenthesis") {
     newObj.target = newObj.target.inner;
--- a/src/printer.js
+++ b/src/printer.js
@@ -13,7 +13,8 @@ const {
   dedent,
   ifBreak,
   hardline,
-  softline
+  softline,
+  align
 } = require("prettier").doc.builders;
 const { willBreak } = require("prettier").doc.utils;
 const {
@@ -76,7 +77,29 @@ function genericPrint(path, options, print) {
       comments.printDanglingComments(path, options, /* sameIndent */ true)
     ]);
   }
-  const printed = printNode(path, options, print);
+  // if this is embedded code, lets try to match the indentation of the open <?php tag
+  let printed;
+  if (isPrevNodeInline(path)) {
+    const tagOpenIndex = options.originalText.lastIndexOf(
+      "<",
+      options.locStart(node)
+    );
+    let alignment = options.originalText.substring(0, tagOpenIndex);
+    if (options.originalText.lastIndexOf("\n", tagOpenIndex) > -1) {
+      alignment = options.originalText.substring(
+        options.originalText.lastIndexOf("\n", tagOpenIndex) + 1,
+        tagOpenIndex
+      );
+    }
+    printed = printNode(path, options, print);
+    printed = willBreak(printed) ? concat([hardline, printed]) : printed;
+    printed = align(
+      new Array(alignment.length + 1).join(" "),
+      printNode(path, options, print)
+    );
+  } else {
+    printed = printNode(path, options, print);
+  }
   if (node.kind === "inline") {
     const parentNode = path.getParentNode();
     return concat([
--- a/tests/inline/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/inline/__snapshots__/jsfmt.spec.js.snap
@@ -360,6 +360,84 @@ exports[`mixed-4.php 1`] = `
 
 `;
 
+exports[`mixed-5.php 1`] = `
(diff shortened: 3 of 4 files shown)


src/clean.js
```diff
@@ -119,6 +119,11 @@ function clean(node, newObj) {
     });
   }
 
+  // for inline code, we always use the long form <?php
+  if (node.kind === "echo") {
+    newObj.shortForm = false;
+  }
```
evilebottnawi: It is not right, prettier should not change ast, many people use difference `echo` form in difference situation. We need output echo as is.

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -360,6 +360,84 @@ exports[`mixed-4.php 1`] = `
...
+    <?php echo $this->element(
+        'collection/item-title',
+        [
+            'class' => $itemTitleClass,
+            'small_title' => $item->small_title,
+            'title' => $item->title,
+            'url' => $url,
+            'color' => $item->title_color ?: 'primary'
+        ]
+    ); ?>
```
evilebottnawi: We should use there `group`.

I.E. if it is allow be in one line print one line:
```php
<?php echo 1; ?>
```

If something break - break all (add newline after `<?php` and before `?>`) (as we do in prettier :smile: ):
```php
<?php
echo $this->veryVeryVeryVeryVeryVeryVeryLongFunction(
    'something',
   [
      'class' => $itemTitleClass
      // etc
   ]
);
?>
```

src/clean.js
mgrip: fine by me 👍

mgrip:
thanks for the feedback @evilebottnawi! i can try and take another look again next week, but feel free to make updates on the branch if you'd like in the meantime!

mgrip:
@evilebottnawi any preference on a gameplan for this? should we continue this PR or are you still planning on doing a separate one?

evilebottnawi:
@mgrip don't have time right now, sorry, regular work, can we separate this pr on three parts as i wrote above?

mgrip:
no worries! just checking. and what do you mean exactly by separating? I was at least trying to only address `Support indent in <?php and ?> (issue #394)`

evilebottnawi:
@mgrip hm, let's fix all problems which i describe above :+1:

tests/inline/__snapshots__/jsfmt.spec.js.snap
mgrip: @evilebottnawi I've been digging into this and I'm unsure if its going to be possible...

Inline nodes and regular php nodes are stored next to each other in a flat list, there's no top level node to group. ie if you had something like this

```php
<div>Foo</div>
<?php
$a = 1;
$some_otherThing = 1;
?>
<span>test</span>
```

the node list ends up being

```
[
  {type: "inline" ...},
  {type: "assign" ...},
  {type: "assign" ...},
  {type: "inline" ...}
]
```

So if we wanted to group the php open/close tags with the assign nodes, there's no parent node to actually group them together. Any ideas?

mgrip: We could maybe do something hacky our parser file to look through the ast we get from `php-parser` and create a faux top-level node for anything surrounded by inline nodes? Thats the only think I can think of atm

evilebottnawi: @mgrip Let's solve this in `3. Improve group for <?php and ?>`. Right now let's left this as is

mgrip: sounds good 👍

src/printer.js
```diff
@@ -76,7 +77,29 @@ function genericPrint(path, options, print) {
       comments.printDanglingComments(path, options, /* sameIndent */ true)
     ]);
   }
-  const printed = printNode(path, options, print);
+  // if this is embedded code, lets try to match the indentation of the open <?php tag
+  let printed;
+  if (isPrevNodeInline(path)) {
+    const tagOpenIndex = options.originalText.lastIndexOf(
+      "<",
```
evilebottnawi: Let's add test: `<?php echo '<'; ?><?php echo '<'; ?>`. I don't sure what using only `<` is good solution

mgrip:
awesome 👍 ill try to get the ast change and `"<"` issues fixed tomorrow!

mgrip:
After more investigation i think something like the solution i outlined above might be necessary to work - i was only looking at examples with 1 node, but for any inline code with multiple statements, we really need a parent grouping for the `align()` to work...

What do you guys think about something like this?

> We could maybe do something hacky in our parser file to look through the ast we get from php-parser and create a faux top-level node for anything surrounded by inline nodes? Thats the only think I can think of atm

mgrip:
Opened this on the parser side just to start a discussion https://github.com/glayzzle/php-parser/issues/159

mgrip:
Unfortunately this turned out to be more difficult than I thought - I'm still unsure of the best way to solve this...

mgrip:
eclipsed by https://github.com/prettier/plugin-php/pull/431

PR #410, opened 2018-05-03
Fix inline comments before else if

### LINKED ISSUE #330: Inline comment before elseif changes elseif into "else if"
PHP: 7.0
Node: 9
prettier-php: 26c6262
prettier: <code>[d676188](https://github.com/prettier/prettier/commit/d676188dbb0c99de57bdffda96ab0cccb1abfa1d)</code>

### Input
```php
class Page_Title {
	public function get_title() {
		if ( is_main_site() && is_front_page() ) {
			return '';
		}

		// Blog
		if ( is_home() ) {
			$title = __( 'News', 'tribe' );
		}

		// Search
		elseif ( is_search() ) {
			$title = __( 'Search Results', 'tribe' );
		}

		// 404
		elseif ( is_404() ) {
			$title = __( 'Oops! Page Not Found', 'tribe' );
		}

		// Singular
		elseif ( is_singular() ) {
			$title = get_the_title();
		}

		// Archives
		else {
			$title = get_the_archive_title();
		}
		
		return $title;
	}
}
```

### Output
```php
class Page_Title
{
	public function get_title()
	{
		if (is_main_site() && is_front_page()) {
			return '';
		}

		// Blog
		if (is_home()) {
			$title = __('News', 'tribe');
		} else// Search
		if (is_search()) {
			$title = __('Search Results', 'tribe');
		} else// 404
		if (is_404()) {
			$title = __('Oops! Page Not Found', 'tribe');
		} else// Singular
		if (is_singular()) {
			$title = get_the_title();
		} else {
			// Archives
			$title = get_the_archive_title();
		}

		return $title;
	}
}
```

### Expected Output
```php
class Page_Title
{
	public function get_title()
	{
		if (is_main_site() && is_front_page()) {
			return '';
		}

		// Blog
		if (is_home()) {
			$title = __('News', 'tribe');
		} elseif (is_search()) {
			// Search
			$title = __('Search Results', 'tribe');
		} elseif (is_404()) {
			// 404
			$title = __('Oops! Page Not Found', 'tribe');
		} elseif (is_singular()) {
			// Singular
			$title = get_the_title();
		} else {
			// Archives
			$title = get_the_archive_title();
		}

		return $title;
	}
}
```

---

I'd also be OK with having the comments moved to the end of the line, after the `{`.

Regardless, I would not expect my `elseif` to change into `else if`. Or, alternatively, I would expect all `elseif`s to change into `else if`s, which is not happening.

czosel:
Fixes #330

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -1030,11 +1030,15 @@ if (1) { // Comment
     // Comment
     echo "Foo"; // Comment
     // Comment
-} elseif (2) { // Comment
+}
+// Comment
+elseif (2) { // Comment
     // Comment
     echo "Bar";
     // Comment
-} else { // Comment
+}
+// Comment
+else { // Comment
     // Comment
     echo "FooBar";
     // Comment
@@ -1045,9 +1049,10 @@ if (/* Comments */ true /* Comments */) {
 } else {}
 
 
-  if (2 /* MB_OVERLOAD_STRING */ & (int) ini_get('mbstring.func_overload')) {
-      mb_internal_encoding('ASCII');
-  }~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+if (2 /* MB_OVERLOAD_STRING */ & (int) ini_get('mbstring.func_overload')) {
+  mb_internal_encoding('ASCII');
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
 // Comment
 if (1) {
@@ -1055,12 +1060,13 @@ if (1) {
     // Comment
     echo "Foo"; // Comment
     // Comment
-} elseif (2) {
+} elseif (2) {// Comment
     // Comment
     // Comment
     echo "Bar";
     // Comment
-} else {
+} else {// Comment
+    // Comment
     // Comment
     // Comment
     echo "FooBar";
--- a/src/comments.js
+++ b/src/comments.js
@@ -30,7 +30,8 @@ const handleOwnLineComment = (comment, text, options) => {
     handleFunctionParameter(comment, text, options) ||
     handleFunction(comment, text, options) ||
     handleForLoop(comment) ||
-    handleTryCatch(comment)
+    handleTryCatch(comment) ||
+    handleAlternate(comment)
   );
 };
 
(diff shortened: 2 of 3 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1045,22 +1049,26 @@ if (/* Comments */ true /* Comments */) {
...
-} elseif (2) {
+}
+// Comment
+elseif (2) {
     // Comment
     // Comment
     echo "Bar";
     // Comment
 } else {
+    // Comment
```
czosel: This one is still at the wrong place - not exactly sure how to fix that...

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1045,22 +1049,26 @@ if (/* Comments */ true /* Comments */) {
...
 // Comment
 if (1) {
     // Comment
     // Comment
     echo "Foo"; // Comment
     // Comment
-} elseif (2) {
+}
+// Comment
+elseif (2) {
```
nicoder: this does not look like valid PSR-2: https://www.php-fig.org/psr/psr-2/#51-if-elseif-else

but to make it valid we would have to move (or remove) the comment

czosel: Oh right, I didn’t even think of that! Shall we just move the comment to the top of the body of the elseif/else?

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1045,22 +1049,24 @@ if (/* Comments */ true /* Comments */) {
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
 // Comment
 if (1) {
     // Comment
     // Comment
     echo "Foo"; // Comment
     // Comment
-} elseif (2) {
+} elseif (2) {// Comment
```
czosel: Alright, I think this is more PSR-compatible @nicoder. Not sure if there should be a space before the comment?

tests/comments/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: I agree looks very ugly, but PSR-12 don't have section about comments and where they should be placed. Let's do it pretty and move this comments in some other place.

tests/comments/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: Let's left as is, if somebody have problem with this we can have discussion about this and fix it late

tests/comments/__snapshots__/jsfmt.spec.js.snap
czosel: @evilebottnawi I agree - what do you think about how it's done now? (see below)

tests/comments/__snapshots__/jsfmt.spec.js.snap
virgofx: Space before comment

PR #418, opened 2018-05-12
Add tests for PHP code blocks in markdown

czosel:
While this is working already, there are a few issues:

* code is indented with 2 spaces instead of 4
* the leading `<?php` is required


czosel:
Shall we add commits to this PR that address to open issues?

evilebottnawi:
Fixed :+1:

Issue #423, opened 2018-06-14
PHP Fenced Code blocks breaking in Markdown files


joemaller:
PHP-flavored fenced code blocks in markdown files are breaking when the Prettier PHP Plugin is active. The code blocks are recognized as PHP, but the last newline in the code seems to be missing. That causes the closing code fence to wrap up a line, breaking Markdown's syntax. 

Markdown input:

    ```php
    $foo = 'bar';

    ```

Prettier output (with PHP plugin active)

    ```php
    $foo = 'bar';```

Disabling the PHP plugin, removing the language label or changing the label to anything but `php` yields correct formatting: 

Expected Prettier output:

    ```php
    $foo = 'bar';
    ```

I tried bisecting for an earlier working version of the plugin, but didn't find one. This bug appears to have been around for a while.

One important note: This is not a one-off problem. The issue compounds on subsequent Prettier calls, creating a mess of backticks as the Markdown parser attempts to close the fenced code block. The result of several passes ends up something like this:

This is the result of calling Prettier 5 times in a row:

    ````````````````````````````````php
    $foo = 'bar';```````````````````````````````````````````````````````````````

Again, this doesn't happen unless the PHP plugin is installed. 

Hope this is helpful, thanks for the fantastic tool.

joemaller:
Possibly related to these trailing newline issues? https://github.com/prettier/plugin-php/issues/304  https://github.com/prettier/plugin-php/issues/305

mgrip:
@joemaller Thanks for reporting!

mgrip:
So after a little investigating this is actually a different issue - unlike js, php needs the open tags `<?php` to be recognized, and they aren't there when prettier calls the php parser. Currently the php code is just being treated as an inline node, which is why we aren't adding the newline (and also the php code isn't actually getting formatted at all). We'll probably need to do something a little hacky to add the open tag or something if we know its valid php.

mgrip:
Although to your point - even if its not recognized as php code it still should include that line break. I'll get a PR posted soon

mgrip:
@joemaller should be all set now - let us know if you see any other problems!

evilebottnawi:
@mgrip maybe patch release?

mgrip:
oh right good call haha - im gonna see if i can deal with #305 as well, if I can't figure it out by eod I'll release an update 👍

PR #429, opened 2018-06-20
add newline for inline if it exists in the input

mgrip:
fixes https://github.com/prettier/plugin-php/issues/423


tests/markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,13 @@
...
+exports[`markdown.md 1`] = `
+\`\`\`php
+$foo = 'bar';
+
+\`\`\`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+\`\`\`php
+$foo = 'bar';
+\`\`\`
+
```
evilebottnawi: I think we should avoid newline in end in markdown

mgrip: I think we need to though, or else it will break the markdown? Unless you know of a another way to fix #423?

evilebottnawi: @mgrip I looked wrong place, all right here

tests/markdown/markdown.md
```diff
@@ -0,0 +1,4 @@
+```php
+$foo = 'bar';
+
+```
```
evilebottnawi: Let's add test
```php
$foo = 'bar';
```

mgrip: you mean just w/o the newline?

mgrip: if yes, just added

mgrip:
should be good for another look @evilebottnawi 👍

evilebottnawi:
Feel free to merge

Issue #305, opened 2018-04-02
Inline hanging comments at end of file with extra newlines dropped or moved outside of php


cmancone:
Obviously related to #304, but since there is a bit extra here (and it is potentially a breaking issue) so I'm creating another issue: feel free to close if appropriate.  On a file with extra newlines at the end **and** a comment at the end of the file, that comment can either be lost or moved outside of the PHP all together.  Two examples:

**Input:**
```php
<?php
class test{
} // some comment
?>


```
**Output:**
```php
<?php
class test
{

}
?>
```

Escaping PHP example
**Input:**
```php
<?php
class test{
}
// some comment
?>


```
**Output:**
```php
<?php
class test
{

}
?>// some comment

```

mgrip:
Similar to #304, output is now

```php
<?php
class test
{

} // some comment


```

Do we think that's correct, or do we care about the closing tag?

evilebottnawi:
@mgrip yep, as i said before, we should output all php tags as is, we prettier not codemod

cmancone:
Obviously do whatever you guys want, but removing a closing PHP tag at the end of a file is not a breaking change (I've actually started removing them from our codebase).  That being said there is plenty of room for trouble doing so if the file has hanging newlines after the closing PHP tag.  If you remove a PHP tag in that case you would modify the meaning of the file (i.e. were it a template file, removing an ending PHP close tag when there are empty new lines after it would effectively remove the empty new lines from the end of the generated HTML).

evilebottnawi:
@cmancone because you have `HTML` and `PHP`, for other formats (not PHP) it can be danger and can break output

mgrip:
Maybe we could check the source and see if there's anything existing outside the program node? (ie whitespace), and if so, keep the closing tag?

evilebottnawi:
@mgrip yep source code is good place to detect when we should output `<php` or `?>`

Issue #304, opened 2018-04-02
Newlines after closing PHP tag


cmancone:
I noticed some issues related to new lines after a closing PHP tag.  The primary effect is the stripping of some of these newlines.  For a file that contains just a class definition this would actually be a good thing (although it causes debug-check to crash), but for something more like a template file it might be considered a change (although HTML obviously ignores whitespace).

**Input:**
```php
<?php
$code = true;
?>


```
**Output:**
```php
<?php
$code = true;
?>
```
running with `--debug-check` on results in:
```
[error] test.php: prettier(input) !== prettier(prettier(input))
[error] Index: 
[error] ===================================================================
[error] --- 
[error] +++ 
[error] @@ -1,3 +1,2 @@
[error]  <?php
[error]  $code = true;
[error] -?>
[error] 
error An unexpected error occurred: "Command failed.
Exit code: 2
Command: sh
Arguments: -c prettier --plugin=. --parser=php test.php --debug-check
Directory: /var/www/prettier/plugin-php
Output:
".
```

evilebottnawi:
@cmancone known issue, we will fix it in near future

mgrip:
So the output is now

```php
<?php
$code = true;

```

Do we think thats correct? Do we care about the closing tag?

evilebottnawi:
@mgrip yep, we should output all php tags as is

czosel:
The closing tag should follow naturally once we have proper inline support, right? Until then, I wouldn't implement any special case for this.

mgrip:
@czosel what do you mean by "proper inline support" exactly?

mgrip:
fixed in https://github.com/prettier/plugin-php/pull/430

PR #430, opened 2018-06-20
print closing php tags if they exist in the source

mgrip:
fixes https://github.com/prettier/plugin-php/issues/305, https://github.com/prettier/plugin-php/issues/304


src/util.js
```diff
@@ -358,7 +358,12 @@ function lineShouldEndWithSemicolon(path) {
 }
 
 function lineShouldHaveEndPHPTag(path) {
-  return isNextNodeInline(path);
+  const node = path.getValue();
+  const isProgramNode = node.kind === "program";
+  return (
+    isNextNodeInline(path) ||
+    (isProgramNode && node.loc.source.trim().endsWith("?>"))
```
evilebottnawi: Looks good, I'm not sure that this check is good, but we can add more tests in future and fix this problems in future

PR #440, opened 2018-06-22
tests: comment before inline

### LINKED ISSUE #435: Comments being moved above opening tag for inline php
This might be a duplicate of the other comment bugs, but listing here just in case.

Version: 0.2.2

Input/Expected:

```php
some text
<?php
/**
 * A comment
 */
$thing = 1;
```

Output:

```php
some text
/**
 * A comment
 */
<?php
$thing = 1;
```

evilebottnawi:
fixes #435

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -1150,6 +1150,22 @@ include // Comment
 
 `;
 
+exports[`inline.php 1`] = `
+some text
+/**
+* A comment
+*/
+<?php
+$thing = 1;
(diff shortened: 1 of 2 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1150,6 +1150,22 @@ include // Comment
...
+* A comment
+*/
+<?php
+$thing = 1;
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+some text
+/**
+* A comment
+*/
+<?php $thing = 1;
```
evilebottnawi: Will be fixed in other PR

Issue #419, opened 2018-05-23
Formatting of `::` 


mathiasrw:
I'm super excited for this plugin. Letting hybrid web-apps use one formatter for everything is awesome. 

> Please note that this plugin is currently in alpha stage and still under active development. We encourage everyone to try it and give feedback

On the note of "We encourage everyone to try it and give feedback" I have the following feedback:

Recently the plugin formatted like this

```
DB::table('identity')->insert([
	'ref' => $ref,
	'handle' => $handle
]);
```

however, something changed the last week so it now formats like this

```
DB
	::table('identity')
		->insert([
			'ref' => $ref,
			'handle' => $handle
		]);
```

Is anyone aware if this reflects a change I prettier itself, or the plugin-php?

Asked a few programmers and they all really prefer the previous formatting. I have a feeling the cognitive connection of having "table", "Identity" and "Insert" is helpful. 

Is there a thread discussing the effects of, for example, placing the `::` part on a new line?

evilebottnawi:
@mathiasrw thanks for PR, it is bug

evilebottnawi:
Should be easy fix :+1:

mgrip:
Interesting though that it changed recently... I don't think there have been any changes deployed to this repo in at least a couple weeks

mathiasrw:
Sounds good. Ill close it...

evilebottnawi:
Need add tests

mgrip:
For reference: https://prettier.io/playground/#N4Igxg9gdgLgprEAuEARAQgdwJYwBYA20A5lAIYC2cSSMZARgXABQDk2AJgjLgJ6sBKAHTYoAZzgAnGACU4ZAgV5yFSlYt4AZEgDlKLYWIhUA8vinq18jdqjFmAgNwgANCAgAHHtDHJQZSUkITAAFAIRfFDIANwhOVxB6STIwAGs4GABlDxTRYmQYSQBXODdRCWkQ5OIKMmQAMwUJNwArMQAPdGS0jMz9TVE4BqbSkBzJCuREhl4iKASPSVEYAHVOfGQADgAGN0WICRXkjynFuAroobdJOABHIuwbqrIauqRGgmaQCQpsAuLRmI8kwAIpFCDwYafUZ0ehrDgbJAAJjchTI2AIeQAwsZalMoNAriAihIACoMSIfCQAX2pQA

mathiasrw:
🎉

Issue #420, opened 2018-06-07
Shouldn't break immediately for method chains


mgrip:
We always break method chains after the 1st member, but looks like prettier JS does it after the second

output:
```php
Logger
    ::use_logger('albus')
    ->info('Albus: failed to find active version for albus object', [
         'uuid' => $uuid,
         'requested_date' => $date
     ]);
```

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAZCBzDcBOA6AVwGc4B9AG02xwAoByAQ3ICNi6BKAHSgAI-+BeAJZQAZhHoBBFsSQ9RDIeTgATHjAjyRahmBhCAbnB5GcRIdHkQcPJqyI8IzAFZw9dADQ9g3AX-50BARCKnRyACRBIR6+-n50OHAAjgRwRPAqpCoM8GE84dnw3AC+7ADc3NzoWLgA6kIwABboUBgAcgwAtnCEJADKEN0A8k24ACqNIhj1TS3tXXA07HiNcOSUS3jQcACy1ovL23uJAKIAHl0ADsozzdDzDI9P3UtlIB4gEJf60ETIoAwcDgIAB3AAKgIQfxQDAMEBC7xAzBwugA1nAYH1LropsgYDhUh8RCQcDAwSiMJ0GMgFOQSB9nEQzgAhFFgdGYhaoERwGlMekgbFmXDIJEMZgAT0oUERlxwIhg9RUTWQAA4AAwfOUQEi1FGXUVytK4IyIxIpISJckMSnUpC0gUkTpCPEEuAfcytZQARQIEHgfLp7pAMHFSpVSAATB98YpyFMAMKDKmiqDbRHEOBjcXQh1wYrFIA

evilebottnawi:
[...]

PR #447, opened 2018-06-26
fix: removing constref from member groups when chained

mgrip:
fixes https://github.com/prettier/plugin-php/issues/419

### DIFF
--- a/tests/call/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/call/__snapshots__/jsfmt.spec.js.snap
@@ -128,7 +128,7 @@ DB::table('identity')->insertReallyReallyReallyLongName([
 	'handle' => $handle
 ]);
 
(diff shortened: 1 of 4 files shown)


tests/call/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -253,4 +268,20 @@ implode(
     }, array_merge($arr1, $arr2, $arr3))
 );
 
+DB::table('identity')->insertReallyReallyReallyLongName([
+    'ref' => $ref,
+    'handle' => $handle
+]);
+
+DBwithlongname::table('identity') .
```
mgrip: whoops didn't notice the spacing here - will udpate

mgrip:
Sorry closing this - didn't realize its actually tied to https://github.com/prettier/plugin-php/issues/420 as well

PR #448, opened 2018-06-28
fix: method chain grouping

mgrip:
Fixes https://github.com/prettier/plugin-php/issues/419 https://github.com/prettier/plugin-php/issues/420

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -325,76 +325,95 @@ function printMemberChain(path, options, print) {
   //   a()
   //     ->b->c()
   //     ->d();
-  const groups = [];
-  let currentGroup = [printedNodes[0]];
-  let i = 1;
-  for (; i < printedNodes.length; ++i) {
-    if (
-      printedNodes[i].node.kind === "call" ||
-      (printedNodes[i].node.kind === "offsetlookup" &&
-        printedNodes[i].node.offset &&
-        printedNodes[i].node.offset.kind === "number")
-    ) {
-      currentGroup.push(printedNodes[i]);
-    } else {
-      break;
-    }
-  }
-  if (printedNodes[0].node.kind !== "call") {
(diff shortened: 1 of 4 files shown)


tests/member_chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -539,8 +538,7 @@ $window->Data['key']("foo")
         return $b;
     });
 
-$window->Data['key']
-    ['foo']("foo")
+$window->Data['key']['foo']("foo")
```
evilebottnawi: Regression

src/printer.js
```diff
@@ -325,76 +325,95 @@ function printMemberChain(path, options, print) {
...
+      currentGroupIndex: 0,
+      shouldStartNewGroup: true,
+      delayedNewGroup: false,
+      groups: []
     }
-  }
-  if (currentGroup.length > 0) {
-    groups.push(currentGroup);
-  }
+  );
```
evilebottnawi: I don't think it is good solution:
1. It breaks compatibility with the original prettier algo
2. A lot of bug and regressions, try to run this on laravel-framework
3. It can be solve in more easy way

mgrip: Ok I thought this simplified things and made it more understandable (i couldn't figure out how the existing one worked), but fair points. Do you have examples for 2? We can add more tests. Also what for 3. what are you thinking?

evilebottnawi: @mgrip i will do PR with fix in tomorrow with tests and example how we solve this in easy way :+1:

evilebottnawi: sorry guys for delay, a lot of work :disappointed: , send the PR tomorrow :star:

evilebottnawi:
Close PR due already solved :+1:

PR #455, opened 2018-07-04
Update prettier devDependency

czosel:


package.json
```diff
@@ -20,7 +20,7 @@
     "eslint-plugin-prettier": "2.4.0",
     "jest": "21.1.0",
     "jest-runner-eslint": "0.3.0",
-    "prettier": "prettier/prettier#4c464133ac37a966c792cacc660853d74c776841"
+    "prettier": "prettier/prettier"
```
czosel: I'm just linking to the current master here, because it might help us discover issues early in the tests - do you agree?

evilebottnawi: :+1:

Issue #456, opened 2018-07-05
SyntaxError for multiline string


shreyas-a:
Following is a valid `PHP` code snippet, still prettier throws a syntax error:

`SyntaxError: Parse Error : syntax error, unexpected '{'`

```
<?php

$chart = '
{
  "chart": {
    "caption": "Worlds Top 10 largest companies in 2017"
  }
}';

?>
```

evilebottnawi:
@czosel feel free to send PR with test example, maybe old plugin version

evilebottnawi:
Fixed https://github.com/prettier/plugin-php/pull/457 and https://github.com/prettier/plugin-php/pull/462

PR #457, opened 2018-07-05
Tests: Multiline strings

czosel:
refs #456

### DIFF
--- a/tests/string/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/string/__snapshots__/jsfmt.spec.js.snap
@@ -1,5 +1,30 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`multiline.php 1`] = `
+<?php
+$string = 'foo
+bar';
+$chart = '
+{
+  "chart": {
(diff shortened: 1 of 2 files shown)


tests/string/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,5 +1,29 @@
...
+  "chart": {
+    "caption": "Worlds Top 10 largest companies in 2017"
+  }
+}';
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+$string = 'foo
+bar';
+$chart =
+    '
```
czosel: By the way: Do we expect this linebreak?

evilebottnawi: @czosel no, looks bug, need create issue

czosel: Done! #461

Issue #454, opened 2018-07-03
support --require-pragma and --insert-pragma


roippi:
https://prettier.io/docs/en/options.html#require-pragma
https://prettier.io/docs/en/options.html#insert-pragma

These are fantastic options to have when attempting to incrementally adopt prettier into a large codebase.   Would be great to have them supported in PHP mode.

PR #458, opened 2018-07-05
implements --require-pragma and --insert-pragma

roippi:
Closes https://github.com/prettier/plugin-php/issues/454.

### DIFF
--- a/src/pragma.js
+++ b/src/pragma.js
@@ -1,0 +1,58 @@
+"use strict";
+
+const docblock = require("jest-docblock");
+const parse = require("./parser");
+
+const extractDocBlocks = text => {
+  const parsed = parse(text);
+  return parsed.comments.filter(el => el.kind === "commentblock");
+};
+
+const hasPragma = text => {
+  const [firstDocBlock] = extractDocBlocks(text);
+  if (firstDocBlock) {
+    const { value } = firstDocBlock;
+    return value.indexOf("@prettier") !== -1 || value.indexOf("@format") !== -1;
+  }
+  return false;
+};
+
+const insertPragma = text => {
+  const [firstDocBlock] = extractDocBlocks(text);
+  let before;
+  let after;
+  let parsedDocblock;
+  if (firstDocBlock) {
+    const {
+      start: { offset: startOffset },
+      end: { offset: endOffset }
+    } = firstDocBlock.loc;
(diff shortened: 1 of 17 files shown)


tests/insert-pragma/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,122 @@
...
+
+echo "Hello World";
+
+// non-docblock comment
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<div>
+  inline stuff up top
+</div>
+
+<?php /** @format */
```
evilebottnawi: Not sure it is good idea do this, maybe better insert new php tag at the top of file with pragma, but we can do this after solve problem with inline printing

roippi: Agree, I wasn't thrilled with doing it this way

roippi: I think I will just drop `--insert-pragma` support for files that don't start with a php tag.  Seems simplest.

src/pragma.js
```diff
@@ -0,0 +1,66 @@
+"use strict";
+
+const parse = require("./parser");
+const { EOL } = require("os");
+
+const extractDocBlocks = text => {
+  const parsed = parse(text);
```
mgrip: any idea how often this gets called or if there's any performance concern? (I'm guessing once per file which probably isn't a huge deal)

roippi: Yup, just checked:

- once per file, if you use no flags or `--require-pragma`
- twice per file, if you use `--insert-pragma` (or both flags)

Some room for optimization.  I didn't know that this hook is still called even when you use no flags.

roippi: I'll do a couple quick optimizations here, give me a bit.

roippi: Test suite slowed down by ~1% with my original patch, now closer to .1%.  Pretty much noise.

LMK if you don't want the new npm dependency, memoize impls are pretty interchangeable.

evilebottnawi:
@mgrip need minor release

PR #460, opened 2018-07-06
feat: better print empty declarations

evilebottnawi:


src/printer.js
```diff
@@ -1420,15 +1420,30 @@ function printStatement(path, options, print) {
             ])
           )
         : "";
+
+      const hasEmptyBodyContents =
```
czosel: Nitpick: I'd call it `hasEmptyBody` (same as below)

evilebottnawi: @czosel :+1:

PR #462, opened 2018-07-06
fix: don't break assign with multiline

### LINKED ISSUE #461: Linebreak in multiline strings
Some multiline strings seem to have a linebreak before the opening quote. It seems to have something to do both with line length and number of lines :confused: 

Input:
```php
$multiline = '
  looooooooooooooooooooooooooooooooooooooooooooooooooooooong
  more
';
```
Output:
```php
$multiline =
     '
  looooooooooooooooooooooooooooooooooooooooooooooooooooooong
  more
';
```

Excpected output: Same as input.

evilebottnawi:
fixes #461

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1745,6 +1745,8 @@ function printStatement(path, options, print) {
         (node.right.kind === "retif" && node.right.test.kind === "bin") ||
         isPropertyLookupChain(node.right);
       const isNumberNode = node.right.kind === "number";
+      const isMultiLineString =
+        node.right.kind === "string" && node.right.raw.includes("\n");
       return group(
         concat([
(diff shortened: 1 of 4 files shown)


src/printer.js
```diff
@@ -1731,13 +1731,20 @@ function printStatement(path, options, print) {
         //     : false;
         (node.right.kind === "retif" && node.right.test.kind === "bin") ||
         isPropertyLookupChain(node.right);
+      const isMultiLineString =
+        node.right.kind === "string" && node.right.raw.indexOf("\n") !== -1;
```
czosel: You can use `includes()` instead of `indexOf() !== -1`

PR #470, opened 2018-07-14
Respect short echo syntax

### LINKED ISSUE #467: Short echo syntax
Seeking some clarification on this behavior:

- Input: `There are <?= $variable; ?> examples.`
- Want: `There are <?= $variable; ?> examples.`
- Actual output: `There are <?php echo $variable; ?> examples.`

Is the [“short syntax”](https://php.net/echo) for `echo` being translated intentionally or is this a bug?

(Also, I'm not vouching for my exact use of optional spaces and semicolon — the PHP docs actually use the very compact `<?=$variable?>` — just wanted to clarify if short syntax is being rewritten on purpose.)

nikolasmatt:
Resolves #467

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -112,7 +112,7 @@ function genericPrint(path, options, print) {
       return dedentToRoot(
         group(
           concat([
-            "<?php",
+            node.kind === "echo" && node.shortForm ? "<?=" : "<?php",
             align(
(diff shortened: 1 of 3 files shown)


evilebottnawi:
@czosel need patch release i think

czosel:
Agreed - just released v0.3.1

PR #475, opened 2018-07-18
fix: print return type declaration for closure

evilebottnawi:
Critical bug :smile: 

Also move all `closure` from `function` directory to `closure` directory

### DIFF
--- a/tests/closure/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/closure/__snapshots__/jsfmt.spec.js.snap
@@ -66,6 +66,67 @@ $foo->bar(
 $emptyFunc = function(){};
 
 $emptyFuncWithComment = function(){ /* Comment */ };
+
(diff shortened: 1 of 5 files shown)


tests/closure/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -135,4 +196,60 @@ $emptyFuncWithComment = function () {
...
+
+$shortArgs_longVars = function ($arg) use (
+    $longVar1,
+    $longerVar2,
+    $muchLongerVar3
+) {
+    // body
+};
+
+$rTD = function (array $array): int {
```
evilebottnawi: We don't print `int` (type) :disappointed:

mgrip: @evilebottnawi what's the problem exactly?

PR #478, opened 2018-07-20
fix: handle empty lines between comments in ast with children

### LINKED ISSUE #368: Blank lines between comments removed
```php
<?php

// The first line of a comment
// The second line of a comment

// A third line which has a blank line before it
```
is transformed to
```php
<?php
// The first line of a comment
// The second line of a comment
// A third line which has a blank line before it
```

I think it would be nicer if we preserved blank lines between comments.

evilebottnawi:
fixes #368 

Need fix problem with inline (full refactor logic), before we can't solve problems with inline nodes we can't release stable version, I ask for help from everyone who is interested in stable release

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -172,6 +172,27 @@ $bool = /*Comment */ true /*Comment */ || /*Comment */ false; /*Comment */
 
 `;
 
+exports[`blank_lines.php 1`] = `
+<?php
+
+// The first line of a comment
+// The second line of a comment
+
+// A third line which has a blank line before it
+
+
+// Other comment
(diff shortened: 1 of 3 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -172,6 +172,27 @@ $bool = /*Comment */ true /*Comment */ || /*Comment */ false; /*Comment */
...
+
+
+// Other comment
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+// The first line of a comment
+// The second line of a comment
+
+// A third line which has a blank line before it
+
+<?php
```
mgrip: this is broken right?

evilebottnawi: And fix lines between comments

PR #483, opened 2018-07-23
test: string with unicode codepoint escape syntax

evilebottnawi:


evilebottnawi:
Merge as is, because tests

PR #491, opened 2018-07-24
fix: extra curly around variable

### LINKED ISSUE #270: Extra curly
Input:
```php
$id = $element->$id_field;
```
Output:
```php
$id = $element->{$id_field};
```

evilebottnawi:
fixes #270 

Note: 
- Fix bug when we break `php5` code (http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.variable-handling.indirect). All values should be printed as is, don't changed.
- Find bug in parser https://github.com/glayzzle/php-parser/issues/167
- Refactor `variable` inside `encapsed` (fast exit, increase perf)

### DIFF
--- a/tests/variable/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/variable/__snapshots__/jsfmt.spec.js.snap
@@ -18,6 +18,38 @@ $_4site = 'not yet';
 $täyte = 'mansikka';
 $bar = &$foo;
 $bar = "My name is $bar";
+$id = $element->{$id_field};
+$id = $element[$id_field];
+$var = <<<EOT
+My name is "$name". I am printing some $foo->foo.
+Now, I am printing some {$foo->bar[1]}.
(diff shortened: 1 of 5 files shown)


tests/variable/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -26,5 +58,38 @@ $_4site = 'not yet';
 $täyte = 'mansikka';
 $bar = &$foo;
 $bar = "My name is $bar";
+$id = $element->$id_field;
+$id = $element[$id_field];
+$var = <<<EOT
+My name is "$name". I am printing some {$foo->foo}.
+Now, I am printing some {$foo
```
mgrip: this looks weird but im guessing separate issue?

evilebottnawi: @mgrip yep, we use same logic for this as prettier use for template literals

PR #492, opened 2018-07-24
feat: better print inline

### LINKED ISSUE #441: Inline groups break prematurely
input/output
```php
<?php declare(strict_types=1); ?>
<html>
    <body>
        <?php
          $a = 1;
        ?>
    </body>
</html>
```
Expected
```php
<?php declare(strict_types=1); ?>
<html>
    <body>
        <?php $a = 1; ?>
    </body>
</html>
```

cc @evilebottnawi

evilebottnawi:
Feel free to feedback, it was very hard, but looks i found how we can solve this problems

Shortly:
1. Remove old logic and implement new (also remove utils around inline).
2. A lot of new tests (you can see what all prints better).
3. A lot of fixed with inline nodes (with old logic you can move all test and looks what it is break code in many places and very very ugly)

After long investigate, I came to the conclusion what we should print space/newline after `<?php` and before `?>` as is, some people love write like:
```php
<?php if ($var) { ?>
  <div>test</div>
<?php } ?>
```
Other:
```php
<?php 
if ($var) { 
?>
  <div>test</div>
<?php 
} 
?>
```

Also it is very difficult to break group when you have inline nodes inside control structure. As is say above i think better print as is - it is doesn't break code and doesn't make it ugly.

fixes #441


src/parser.js
```diff
@@ -2,6 +2,10 @@
 const engine = require("php-parser");
 
 function parse(text) {
+  // By default php parser don't create `inline` node between `?>\n<?`, we recreate their
+  // Need add option to parser to avoid this behaviour
+  text = text.replace(/\?>\n<\?/g, "?>\n___PSEUDO_INLINE_PLACEHOLDER___<?");
```
evilebottnawi: Hack, we need create require on new option to create inline nodes between `?><?php`

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -212,33 +207,26 @@ echo 'bar';
...
+<?php echo 'String.'; ?>
 <p>Test.</p>
 <?php if ($expression == true): ?>
   <p>Test.</p>
 <?php else: ?>
   <p>Test.</p>
-<?php endif;
-echo '<input type="hidden" value="' . htmlspecialchars($data) . '" />'; ?>
+<?php endif; ?>
+<?php echo '<input type="hidden" value="' . htmlspecialchars($data) . '" />'; ?>
```
evilebottnawi: Multiple `<?php ...?><?php ... ?>` now work as expected

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -113,12 +111,9 @@ function inlineNested1()
 
 function inlineMultipleStatements()
 {
-    $a = 1;
-     ?>
+    $a = 1; ?>
```
evilebottnawi: No ` ` before `?>`

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -277,17 +265,13 @@ echo 'bar';
...
   <p>Test.</p>
-<?php
-else:
-    ?>
+<?php else: ?>
   <p>Test.</p>
-<?php
-endif;
-$var = 1; ?>
+<?php endif; ?>
```
evilebottnawi: Great print control structures

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -432,41 +415,36 @@ exports[`mixed-6.php 1`] = `
...
-            ]);
-            ?>
+    ]); ?>
+  </header> <?= $someOtherVar()->element('collection/item-title', [
+      'class' => $itemTitleClass,
+      'small_title' => $item->small_title,
+      'title' => $item->title,
+      'url' => $url,
+      'color' => $item->title_color ?: 'primary'
+  ]); ?>
```
evilebottnawi: Fix aligement

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -482,6 +460,559 @@ There are <?php echo $variable; ?> examples
...
+<div>
+    <?php echo "1";
+    echo "2";
+    echo "3"; ?>
+</div>
+<div>
+    <?php echo "1";
+    echo "2";
+    echo "3";
+    echo "4";
```
evilebottnawi: Support align with multiple statements

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -482,6 +460,559 @@ There are <?php echo $variable; ?> examples
...
+    FooBar
+    <?php
+};
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+$greet = function ($name) {
+    ?>
+    FooBar
+    <?php
+};
```
evilebottnawi: Tests with all nodes which can have inline statement

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -482,6 +460,559 @@ There are <?php echo $variable; ?> examples
...
+    </div>
+<?php } ?>
+<div>Bar</div>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<div>Foo</div>
+<?php for (
+    $really_really_long_variable_name = 0;
+    $really_really_long_variable_name <= 5;
+    $really_really_long_variable_name++
+) { ?>
```
evilebottnawi: Groups break great

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1310,14 +1309,14 @@ exports[`no_code.php 1`] = `
 ?>
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
-/**
+
+?>/**
```
czosel: Regression?

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -482,6 +460,559 @@ There are <?php echo $variable; ?> examples
...
+          echo "1";
+          echo "2";
+          ?>
+      </div>
+    </div>
+  </div>
+</div>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<div>
+    <?php echo "1";
```
czosel: Would it be possible to preserve the linebreak after `<?php`?
```php
<?php
echo "1";
echo "2";
?>
```

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -482,6 +460,559 @@ There are <?php echo $variable; ?> examples
...
+        <div>Newspage</div>
+        <?php break; ?>
+    <?php case 2: ?>
+    <div>Forum</div>
+    <?php break; ?>
+    <?php } ?>
+</div>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<div>
+    <?php switch ($variable) { case 1: ?>
```
czosel: The original version with the linebreak looks better, doesn't it?

evilebottnawi: @czosel `switch` is very difficult print with inline nodes, current output same as was, i think we can do this in other PR, because he is already big

tests/inline/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: @czosel it is hard, works on this right now, maybe impossible

tests/comments/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: @czosel no, it is bug not related to inline (look on preview print, he is also buggy), we should fix it in `comments.js`

evilebottnawi:
@czosel I'll try to document the behavior in the code as much as possible :+1:

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -113,16 +142,40 @@ function inlineNested1()
...
 function inlineMultipleStatements()
 {
-    $a = 1;
-     ?>
+    $a = 1; ?>
     <span>Hello World!</span>
-    <?php
-    $b = 1;
-    ?>
+    <?php $b = 1; ?>
```
mgrip: 🎉

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -113,16 +142,40 @@ function inlineNested1()
...
     <span>Hello World!</span>
     <?php $c = 1;
 }
 
+function test()
+{
+    ?>
+  <span>Hello World!</span>
+  <?php $a = 1;
```
mgrip: should this be on a new line?

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -212,33 +265,26 @@ echo 'bar';
...
-?>
+<?php echo 'String.'; ?>
 <p>Test.</p>
 <?php if ($expression == true): ?>
   <p>Test.</p>
 <?php else: ?>
   <p>Test.</p>
-<?php endif;
-echo '<input type="hidden" value="' . htmlspecialchars($data) . '" />'; ?>
+<?php endif; ?>
```
mgrip: 🎉

tests/inline/__snapshots__/jsfmt.spec.js.snap
mgrip: it's not just switch though right? there are a bunch of examples above (`echo`, `function`) that should break as well

tests/inline/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: @mgrip should be, but difficult, work on this right now

tests/inline/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: I agree switch looks some ugly, let's improve output control structures in other PR, they require logic inside `node`, not only in `printLines`. Main purpose this PR is solve problem with `<?php ... ?><?php ... ?>` and indenting multiple statements inside `<?php ... ?>`.

Also i will try to solve in this PR break multiple statement.

evilebottnawi:
Also it is PR added many tests to avoid breaking something in future

evilebottnawi:
/cc @czosel @mgrip ready for review

In other next PR we solve:
1. Break `<?php function { ... } ?>` (same for methods/try/catch/etc).
2. Break `<?php if (very-long-condition): ?>Content HERE<? endif;?>`.
3. Better print `switch`.

tests/inline/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: I do it :+1:

czosel:
@evilebottnawi done!

PR #496, opened 2018-07-26
fix: alignment statement in block node with inline node

evilebottnawi:
:star:

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1227,22 +1227,22 @@ function wrapPartsIntoGroups(parts, indexes) {
   return indexes.reduce((accumulator, index) => {
     const { start, end, alignment, before, after } = index;
 
+    const printedPartsForGrouping = concat([
+      before || "",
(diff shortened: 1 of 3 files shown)


src/printer.js
```diff
@@ -1227,22 +1227,22 @@ function wrapPartsIntoGroups(parts, indexes) {
   return indexes.reduce((accumulator, index) => {
     const { start, end, alignment, before, after } = index;
 
+    const printedPartsForGrouping = concat([
+      before ? before : "",
```
czosel: maybe easier to just write

src/printer.js
```diff
@@ -1227,22 +1227,22 @@ function wrapPartsIntoGroups(parts, indexes) {
...
+      concat(parts.slice(start, end)),
+      after ? after : ""
```
czosel: same here

src/printer.js

PR #499, opened 2018-07-26
test: unnecessary parens

evilebottnawi:
Just tests. But you can look what we have unnecessary parens around number. Solution:
1. Create issue about implement `parenthesis` through `extra` option (look https://astexplorer.net/#/gist/3b8d354d8e9c4d18758916a3c3fc3f3b/5f02fe391ef5b6727c0632afa857b442d232d936), but it is require a lot of work on parser side.
2. Create function to remove unnecessary parens (we can implement this right now and when parser will support parens through extra options just remove).


mgrip:
I think we should add this as an issue, but I'd be fine with not including it for 1.0, what do you guys think?

evilebottnawi:
@mgrip agree, please create issue in our repo and parser repo, not at the computer right now (inconvenient to print issue text)

mgrip:
https://github.com/prettier/plugin-php/issues/500

PR #501, opened 2018-07-27
Keep linebreaks in arrays

### LINKED ISSUE #476: Multiline arrays collapsed into a single-line array
Prettier should only inline an array definition if the source is already on a single line. This is how it works in JS.

Input/Expected:

```php
$paginate = ['limit' => 32, 'order' => ['Competitions.publish_date' => 'DESC']];

$paginate = [
    'limit' => 32,
    'order' => ['Competitions.publish_date' => 'DESC']
];

$paginate = [
    'limit' => 32,
    'order' => [
        'Competitions.publish_date' => 'DESC'
    ]
];
```
Output:

```php
$paginate = ['limit' => 32, 'order' => ['Competitions.publish_date' => 'DESC']];

$paginate = ['limit' => 32, 'order' => ['Competitions.publish_date' => 'DESC']];

$paginate = ['limit' => 32, 'order' => ['Competitions.publish_date' => 'DESC']];
```

JS prettier reference:

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEASADgQwOYEsqbwAEAvEcEQDa4C2uMSRAzAEwA0REATgCZxeMKAHRABhCDXRwY9XNADOAOnQBXAEbV5ACwD6PQnBGMRAEQCiAZVEiiAXzsBuIVGcYc+A6XLOhManQZmdh8Ybj4BciIRcUlpWQVldU1dfXgjKJBzKxtbZ1snFyg3PAJiMmAQ-3pGVjYQsP5BZyIW1uiJKRkZBNUNXG09A3TTS2sQZtbcqHyfKBA2EAh0bqh5ZFBMLi4IAHcABU2ENZRMADcIXB55kDUuTDAAa2kLLDB8bGQYLhU4Bfx5fgwPZ3bA0TDIABmmEoAIWACt5AAPABCd0ez0wNDgABl8HBIdDYSAsFwAVxkDdMGoAJ6UaDXdDbAEAdTu6ApjLgZNO+IWXDgAEcVLh+cCcGCCTDfiAAXRPt9pfJ3pQ4ABFFQQeCSokwKnMy4wLTIdggL6YXDUKDYGISlBQaC8kAqAEAFSpxy+PwWjPwMH1PENyAAjCwAAy2WxAA

czosel:
Fixes #476

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1131,6 +1131,8 @@ function printExpression(path, options, print) {
           ])
         );
       }
+      const isAssociative = !!node.items[0].key;
+      const shouldBreak = isAssociative && node.loc.source.includes("\n");
       return group(
         concat([
           open,
@@ -1141,7 +1143,8 @@ function printExpression(path, options, print) {
           comments.printDanglingComments(path, options, true),
           softline,
           close
-        ])
+        ]),
+        { shouldBreak }
       );
     }
     case "yield":
--- a/tests/array/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/array/__snapshots__/jsfmt.spec.js.snap
@@ -394,6 +394,45 @@ $arr = [
             return "other-value";
         }
 ];
+
+$singleLine = [ 'do' => [ 'collapse' => 'this' ]];
+$singleLine = array('do' => array('collapse' => 'this'));
+
+$multiline = [
+  'do' => [
+    'not' => 'collapse'
+  ]
+];
+$multiline = ['don\\'t' => 'collapse'
+];
+$multiline = array('don\\'t' => 'collapse'
+);
+
+$mixed = [
+  'only' => [ 'partly' => 'collapsed' ]
+];
+$mixed = [ 'only' => [
+  'partly' => ['coll', 'apsed']
+]];
+
+$arr = array($var
+);
+$arr = array(
+$var);
+$arr = [$var
+];
+$arr = [
+$var];
+$arr = [$foo => $bar];
+$arr = [$foo =>
+$bar];
+
+$collapse = [
+  "not",
+  "associative"
+];
+$collapse = ["not", "associative"
+];
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
 
@@ -419,18 +458,44 @@ $other_test = [1, 2, 123.4324, $hi];
 $keyed_test = array('one' => 123, 'two' => $other);
 $keyed_test = array('one' => 123, 'two' => $other);
 $keyed_test = array('one' => 123, 'two' => $other);
-$keyed_test = array('one' => 123, 'two' => $other);
-$keyed_test = array('one' => 123, 'two' => $other);
-$keyed_test = array('one' => 123, 'two' => $other);
-$keyed_test = array('one' => 123, 'two' => $other);
+$keyed_test = array(
+    'one' => 123,
+    'two' => $other
+);
+$keyed_test = array(
+    'one' => 123,
+    'two' => $other
+);
+$keyed_test = array(
+    'one' => 123,
+    'two' => $other
+);
+$keyed_test = array(
+    'one' => 123,
+
+    'two' => $other
+);
 
 $keyed_test = ['one' => 123, 'two' => $other];
 $keyed_test = ['one' => 123, 'two' => $other];
 $keyed_test = ['one' => 123, 'two' => $other];
-$keyed_test = ['one' => 123, 'two' => $other];
-$keyed_test = ['one' => 123, 'two' => $other];
-$keyed_test = ['one' => 123, 'two' => $other];
-$keyed_test = ['one' => 123, 'two' => $other];
+$keyed_test = [
+    'one' => 123,
+    'two' => $other
+];
+$keyed_test = [
+    'one' => 123,
+    'two' => $other
+];
+$keyed_test = [
+    'one' => 123,
+    'two' => $other
+];
+$keyed_test = [
+    'one' => 123,
+
+    'two' => $other
+];
 
 $short_keyed_test = array(
     'reallyreallyreallyreallylongkey' => 1248,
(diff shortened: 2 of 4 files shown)


tests/array/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -539,4 +639,9 @@ $arr = [
     }
 ];
 
+$paginate = [
+    'limit' => 32,
+    'order' => ['Competitions.publish_date' => 'DESC']
```
czosel: It doesn't seem to work for nested arrays yet.

czosel:
I probably won't have time to continue with this today or tomorrow, so if someone wants to fix the support for nested arrays, feel free!

evilebottnawi:
@czosel i think better look on original prettier logic, they already have good algorithm

czosel:
@evilebottnawi Original prettier logic is using `shouldBreak` (see [here](https://github.com/prettier/prettier/blob/a938076ae5a43e2eea1fb7a0a113e7e038076f55/src/language-js/printer-estree.js#L1096)), and it seems that's working out for us as well :wink: 
Ready for review!

src/printer.js
```diff
@@ -1131,6 +1131,8 @@ function printExpression(path, options, print) {
           ])
         );
       }
+      const isAssociative = !!node.items[0].key;
+      const shouldBreak = isAssociative && node.loc.source.includes("\n");
```
mgrip: Is this what JS does? Seems a little strange to me, but I'm fine w/ it if so

czosel: Do you mean the check for `\n`? JS is calling `hasNewlineInRange`, which might be safer ([see here](https://github.com/prettier/prettier/blob/a938076ae5a43e2eea1fb7a0a113e7e038076f55/src/language-js/printer-estree.js#L1111))

evilebottnawi: @czosel let's use `hasNewlineInRange` :+1: ,maybe we should create issue about move this to shared utils in prettier

czosel: Ok, I'll refactor this PR based on `hasNewlineInRange` and open a PR in prettier core that adds it. Only downside would be that we'd have to point to a specific (unreleased) prettier core version in our `package.json` again :stuck_out_tongue:

evilebottnawi: @czosel we can add own `hasNewlineInRange` util right now and after upgrade remove this code from utils (just left TODO about this)

src/util.js
```diff
@@ -429,6 +429,15 @@ function isProgramLikeNode(node) {
   return ["program", "declare", "namespace"].includes(node.kind);
 }
 
+function hasNewlineInRange(text, start, end) {
```
evilebottnawi: Don't forgot about `TODO` :smile:

src/printer.js
```diff
@@ -1131,6 +1132,14 @@ function printExpression(path, options, print) {
           ])
         );
       }
+      const isAssociative = !!node.items[0].key;
+      const shouldBreak =
+        isAssociative &&
+        hasNewlineInRange(
+          options.originalText,
```
czosel: I think I'd actually prefer the implementation I had before, because we can access the relevant part of the original source code via `node.loc.source`. I think `hasNewlineInRange` only exists because in JS the original source is only accessed "in full" via `options.originalText`.

tests/parens/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -43,7 +43,9 @@ exports[`new.php 1`] = `
         new MessageFormatter(),
         'en',
         $someOtherVar,
-        array('foo' => 'bar')
+        array(
+            'foo' => 'bar'
+        )
```
czosel: Also, the change to `hasNewlineInRange` seems to introduce this regression (I didn't see yet why)

czosel:
@evilebottnawi Do you agree to revert to the original implementation?

tests/parens/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: @czosel Very interesting, we should investigate why it is happens, because it can be parser bug and we should report about this

evilebottnawi:
@czosel let's investigate why it is happens (maybe create issue in parser) and return original implementation)

tests/parens/__snapshots__/jsfmt.spec.js.snap
czosel: Ah, I found the issue: 
```php
text = text.replace(/\?>\n<\?/g, "?>\n___PSEUDO_INLINE_PLACEHOLDER___<?");
```
is breaking the calculation of `offset`, because the placeholders are not contained in `options.originalText`.

I'll revert the implementation, so this issue won't affect this PR anymore. We should probably fix the offset/placeholder issue though?

czosel:
@evilebottnawi Reverted, should be ready to merge now. 

Please also see [my comment above](https://github.com/prettier/plugin-php/pull/501#discussion_r206428928) for the reason why `hasNewlineInRange` didn't work.

tests/parens/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: @czosel i think we can solve this only when parser will be create pseudo inline nodes, somebody can create issue in parser?

tests/array/arrays.php
```diff
@@ -391,3 +391,20 @@ function () {
...
+];
+
+$collapse = [
+  "not",
+  "associative"
+];
+
+$mixed = [
+  'only' => [ 'partly' => 'collapsed' ]
+];
```
evilebottnawi: Can we add tests like:
```php
array($var
);
```
```php
array(
$var);
```

czosel: @evilebottnawi Added some more tests!

tests/parens/__snapshots__/jsfmt.spec.js.snap
czosel: Yeah, this should definitely be solved in the parser. I'll create an issue tomorrow!

tests/array/arrays.php
evilebottnawi: @czosel don't see :smile:

czosel: Oops, I pushed to the wrong remote :sweat_smile:

PR #502, opened 2018-07-27
Improve developer onboarding

Epskampie:
I had some trouble getting started working on this. These changes to CONTRIBUTING.md might help a bit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,4 +18,4 @@ Here's what you need to know about the tests:
 
 Run `yarn lint --fix` to automatically format files.
 
-If you can, take look at [commands.md](commands.md) and check out [Wadler's paper](http://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf) to understand how Prettier works.
+If you can, take look at [commands.md](https://github.com/prettier/prettier/blob/master/commands.md) and check out [Wadler's paper](http://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf) to understand how Prettier works.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -7,6 +7,8 @@ yarn
 yarn test
 ```
 
+Run `yarn prettier path/to/file.php` to format a single file.
```
evilebottnawi: If we add `yarn`, we also should add `npm` instruction

Epskampie: Well, the repo only has a `yarn.lock` file so it's pretty much set up to work with yarn. But I'm happy to add `npm` instructions as well, it might just cause problems for people then they don't get locked depency versions.

evilebottnawi: @Epskampie npm support `package-lock.json` since `npm@5`, but how it is related to issue?

Epskampie: Yeah, I know it does, but there is no `package-lock.json` present in the repo, look: https://github.com/prettier/plugin-php

Anyway, I added npm instructions!

Epskampie:
@evilebottnawi npm instructions added!

Epskampie:
@evilebottnawi npm instructions added!

CONTRIBUTING.md
czosel: @evilebottnawi I think for contributing to the plugin, we should enforce the usage of yarn over npm. Prettier core does the same ([see here](https://github.com/prettier/prettier/blob/master/CONTRIBUTING.md))

Epskampie: No worries, it's reverted!

PR #507, opened 2018-08-01
Preserve linebreaks in only-comment blocks

### LINKED ISSUE #506: Preserve linebreaks in only-comment code blocks
Input:
```php
<?php
//A

// B
?>
<h1>Test</h1>
```
Output:
```php
<?php
//A
// B
?>
<h1>Test</h1>
```

czosel:
Fixes #506

### DIFF
--- a/src/comments.js
+++ b/src/comments.js
@@ -278,14 +278,7 @@ const handleAlternate = comment => {
 
 const handleInlineComments = comment => {
   const { precedingNode, enclosingNode, followingNode } = comment;
-  if (
-    !enclosingNode &&
-    followingNode &&
-    !precedingNode &&
-    followingNode.kind === "inline"
-  ) {
-    // addLeadingComment(followingNode, comment);
-    comment.printed = true;
+  if (!enclosingNode && followingNode && followingNode.kind === "inline") {
     return true;
   } else if (
     !enclosingNode &&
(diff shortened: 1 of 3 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1304,6 +1306,7 @@ exports[`inline6.php 1`] = `
 /**
  * test
  */
+
```
czosel: This should probably go, right?

src/comments.js
```diff
@@ -278,14 +278,7 @@ const handleAlternate = comment => {
...
   const { precedingNode, enclosingNode, followingNode } = comment;
-  if (
-    !enclosingNode &&
-    followingNode &&
-    !precedingNode &&
-    followingNode.kind === "inline"
-  ) {
-    // addLeadingComment(followingNode, comment);
-    comment.printed = true;
+  if (!enclosingNode && followingNode && followingNode.kind === "inline") {
```
czosel: This is just a clean-up that i meant to already in the previous PR.

tests/comments/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: I think no, but it is not related to this PR, we insert hardline before last closing tag (if newline exists in source code), and here we have two newlines, better create new issue

czosel: @evilebottnawi fixed it!

src/printer.js
```diff
@@ -1335,6 +1335,21 @@ function printLines(path, options, print, childrenAttribute = "children") {
     }
 
     if (isInlineNode) {
+      const printComments = comments => {
```
evilebottnawi: Let's move this function to `comments.js`

czosel: Good idea :+1: Also refactored a little.

Issue #514, opened 2018-08-08
AST_COMPARE Troubleshooting


ionrocks:
A few days back, I inherited an older code base that I really wanted to make prettier before diving in! I found this plugin and thought I'd give it a try (despite the alpha status), but unfortunately there's a few parts of the code base that are tripping it up. I dove in to the code to see if I could figure out what was going on, but I hit a snag early on with the tests not seeming to run correctly!

Steps To Reproduce:

```bash
git clone https://github.com/prettier/plugin-php.git
cd plugin-php
yarn
yarn test
AST_COMPARE=1 yarn test
```

With and without the `AST_COMPARE=1` setting, the result is the same: all tests pass. I don't think this is correct, even for some of the existing test cases!

Long story short, some logging showed that the comparison wasn't even running, so I made some changes, got it working for me, and pushed that code here:

https://github.com/ionrocks/plugin-php/tree/testing-updates

I didn't make a pull request because I don't know this code base very well, and am hoping someone will tell me if I'm missing something obvious. 😃

czosel:
Hmm, if I'm not mistaken we _do_ run the AST comparison, but we accept some differences between the original and the formatted AST. In prettier core this "cleanup" stage is called `massageAstNode`, and we implented here: https://github.com/prettier/plugin-php/blob/master/src/clean.js

evilebottnawi:
@czosel what do you think about this? How we can solve this?

PR #515, opened 2018-08-08
Be more specific about when to not wrap constref nodes in property lookups

ionrocks:
Running the tests with AST comparison brought this to my attention. Removing the curly braces in this context means `foo` is no longer being evaluated as a constant.

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -606,7 +606,16 @@ function printArgumentsList(path, options, print, argumentsKey = "arguments") {
 }
 
 function wrapPropertyLookup(node, doc) {
-  const addCurly = !["variable", "constref"].includes(node.offset.kind);
+  let addCurly = true;
+
+  if (node.offset.kind === "variable") {
+    addCurly = false;
+  } else if (
+    node.offset.kind === "constref" &&
+    typeof node.offset.name === "string"
+  ) {
+    addCurly = false;
+  }
 
   return addCurly ? concat(["{", doc, "}"]) : doc;
 }
--- a/tests/propertylookup/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/propertylookup/__snapshots__/jsfmt.spec.js.snap
@@ -26,7 +26,7 @@ $a = $obj->value;
 $a = &$obj->getValue();
 $obj->value = 2;
 
-$b = $this->foo;
+$b = $this->{foo};
 $b = $this->{foo['bar']};
 $b = $this->{'foo' . $bar};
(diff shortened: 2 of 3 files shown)


evilebottnawi:
Please don't send PR request in future while this was not discussed, first create issue and provide examples.

czosel:
Oh, it seems that `$foo->bar` and `$foo->{bar}` is not exactly the same. Consider
```
php > const foo = "bar";
php > class Test { public $bar = "!!!"; }
php > $test = new Test();
php > echo $test->{foo};
!!!
php > echo $test->foo;
PHP Notice:  Undefined property: Test::$foo in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
```

On a side note: I think we welcome PRs as much as we do isses. Changes can be discussed in PRs just as in issues. Thanks a lot for contributing @ionrocks!

evilebottnawi:
@czosel hm, you are right, more tests will be great

czosel:
Hmm, theoretically there should be a place in `clean.js` which is normalizing the AST difference, but I don't see it...

evilebottnawi:
@czosel need investigate :+1:

czosel:
Turns out AST comparison indeed isn’t running, see #514

ionrocks:
This is my first time contributing to a public project like this, so I wasn't sure if an issue was required or not. I should at least have provided an example, not just a one-sentence explanation, but I was posting it right before bed. I'll keep it in mind for next time! 😄

evilebottnawi:
@ionrocks can you add more tests and we can merge this?

ionrocks:
@evilebottnawi Soryr, I'm not quite sure what other kinds of tests to add. This exact fix is covered by the `$a = $obj->value;` and `$b = $this->{foo};` lines in the existing tests, and I don't know any other contexts where this could pop up.

evilebottnawi:
@ionrocks just add `$b = $this->foo;` before you previous test line and update snapshot :+1:

ionrocks:
@evilebottnawi I added the test, and it worked fine locally, but one of the Travis runs failed. Is there a way to ask Travis to retry? It doesn't look like it even started to run the suite.

evilebottnawi:
@ionrocks Thanks!

czosel:
@ionrocks I restarted the build earlier (I think only collaborators can do this in travis), and it turned green :wink:

ionrocks:
Phew, glad the second build worked! Thanks both!

PR #516, opened 2018-08-08
fix: AST comparison during testing

ionrocks:
This is in reference to issue #514.

This code works for me and for what I was trying to test, but I don't know for sure that, for example, `parsers[0] === "php"` is quite the right check, as I'm not very familiar with the internals yet.

How in-depth are pull requests supposed to be? I can add more explanations for the changes if needed!

### DIFF
--- a/tests_config/run_spec.js
+++ b/tests_config/run_spec.js
@@ -3,9 +3,6 @@
 const fs = require("fs");
 const { extname } = require("path");
 const prettier = require("prettier");
-const plugin = require("../src");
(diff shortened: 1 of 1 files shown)


evilebottnawi:
@ionrocks thanks for work! 

@czosel @mgrip  Looks we have problem with ast, we should need fix their first and then merge this PR.

Feel free to fix this problems in other PRs.

evilebottnawi:
Close in favor https://github.com/prettier/plugin-php/pull/550, thanks for work again

PR #518, opened 2018-08-09
fix: empty return with comments

evilebottnawi:
:star: Logic from prettier


evilebottnawi:
Merge as is, because it high priority bug and no big changes

PR #520, opened 2018-08-09
fix: better print key/value in foreach

evilebottnawi:
key/value in own group


tests/foreach/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -135,4 +149,13 @@ foreach (
     var_dump($x, $y);
 }
 
+foreach (
+    [
+        'one' => ['string', 'other-string'],
+        'two' => ['string', 'other-string']
+    ]
+    as $key => $aliases
```
evilebottnawi: Without group `$aliases` moved on next line (ugly), but there are a lot of space

PR #523, opened 2018-08-09
feat: better print `retif`

evilebottnawi:


tests/member_chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -171,17 +171,20 @@ $object[$valid
...
+    : $helper->responseBody($this->defaultUser)
+)
     ->map()
     ->filter();
 
 ($valid
     ? $helper->responseBody($this->currentUser)
-    : $helper->responseBody(defaultUser))->map();
+    : $helper->responseBody(defaultUser)
+)->map();
```
evilebottnawi: A `prettier` do  :star:

tests/retif/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -93,4 +120,59 @@ return (
...
+$var =
+    'string' .
+    $someOtherReallyReallyLongVariable .
+    $someOtherReallyReallyLongVariable
+        ? 1
+        : 2;
+$var = 1
+    ? 'string' .
+        $someOtherReallyReallyLongVariable .
+        $someOtherReallyReallyLongVariable
```
evilebottnawi: Indent :star:

src/printer.js
```diff
@@ -2158,14 +2158,65 @@ function printStatement(path, options, print) {
       ]);
     }
     case "retif": {
-      const printedExpr = concat([
+      const parts = [];
+      const parent = path.getParentNode();
+      const parentParentNode = path.getParentNode(1);
```
czosel: If this is the parent's parent, should we call this `grandParentNode`?

src/printer.js
```diff
@@ -2158,14 +2158,65 @@ function printStatement(path, options, print) {
...
+      const part = concat([
         line,
         "?",
-        node.trueExpr ? concat([" ", path.call(print, "trueExpr"), line]) : "",
+        node.trueExpr
+          ? concat([
+              " ",
+              node.trueExpr.kind === "bin"
+                ? indent(printedTrueExpr)
+                : path.call(print, "trueExpr"),
```
czosel: Replace with `printedTrueExpr`?

src/printer.js
```diff
@@ -2158,14 +2158,65 @@ function printStatement(path, options, print) {
...
       ]);
-      return group(concat([path.call(print, "test"), indent(printedExpr)]));
+
+      parts.push(part);
+
+      // We want a whole chain of retif to all break if any of them break.
+      const maybeGroup = doc =>
+        parent === firstNonRetifParent ? group(doc) : doc;
+
+      // Break the closing paren to keep the chain right after it:
```
czosel: `paren` -> `parens`?

src/printer.js
```diff
@@ -2158,14 +2158,65 @@ function printStatement(path, options, print) {
...
+      // ($a
+      //   ? $b
+      //   : $c
+      // )->call()
+      const breakLookupNodes = ["propertylookup", "staticlookup"];
+      const breakClosingParen =
```
czosel: `breakClosingParens`?

src/printer.js
evilebottnawi: @czosel looks as good name, but in prettier and other places we use `parentParentNode`, i think better left as is

src/printer.js

src/printer.js

src/printer.js

evilebottnawi:
@czosel ready for review

PR #529, opened 2018-08-10
feat: better print retif in array key/value

evilebottnawi:


tests/array/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -649,4 +662,23 @@ $arr = [
...
+        ? 1
+        : 2,
+    'key-two' => true
+        ? 'VeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongString'
+        : 2,
+    'key-two' => true
+        ? 1
+        : 'VeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongString',
+    'key-two' => 'VeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongString' .
+    'VeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongString'
```
evilebottnawi: Same indent with key is not related to this PR, we should improve printing `test` expression of `retif`, let's do it other PR

PR #543, opened 2018-08-16
feat: better print number in entry

evilebottnawi:
Also more tests

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -2432,7 +2432,9 @@ function printNode(path, options, print) {
           concat([
             path.call(print, "key"),
             " =>",
-            ["array", "call", "closure", "retif"].includes(node.value.kind)
+            ["array", "call", "closure", "retif", "number"].includes(
+              node.value.kind
+            )
               ? concat([" ", printed])
               : indent(concat([line, printed]))
           ])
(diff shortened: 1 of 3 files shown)


czosel:
Are you sure this is making sense? Numbers typically aren't that long, and prettier for JS also doesn't seem to do this: https://prettier.io/playground/#N4Igxg9gdgLgprEAuEwA6UBmEJIAQCMRxJpZ5FlV1NGAviADQgQAOMAltAM7KgCGAJ0EQA7gAUhCXin4A3CBwAmTEACNB-MAGs4MAMqstHKAHNkMQQFc4zE9ziCY4zaYC2-ZJn4AbB8wArbgAPACFNHT19fjc4ABkTOC9ffxAjQQdBZHV+NQBPH2hVVkETGAB1ZRgAC2QADgAGZhKIB3LNVmySuEy5JOZBOABHKw5Bl353TyRvP1sQBzcOC2t57hNTHzgARSsIeGS55hhcyqUa5AAmY80OHw2AYQg3D2yoaH6QKwcAFVyZWYOOh0IA

evilebottnawi:
@czosel we don't break number (move on new line) as prettier https://github.com/prettier/plugin-php/pull/543/files#diff-76f26752c3761490942405c0704a8e1dR2559, also you can use number with long indent (we avoid break in this situation).
```php
function test() {
function test() {
function test() {
function test() {
function test() {
function test() {
$array = [
    'veryLongKey' -> 1
];
}
}
}
}
}
}
```

czosel:
Oh right - i was misreading the diff. This one probably needs to be rebased now that #542 was merged.

evilebottnawi:
@czosel already rebased :+1:

PR #544, opened 2018-08-17
fix: alignment for inline nodes

evilebottnawi:
:star:

### DIFF
--- a/tests/inline/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/inline/__snapshots__/jsfmt.spec.js.snap
@@ -1478,6 +1478,152 @@ function testTwo()
 
 `;
 
+exports[`mixed-36.php 1`] = `
+<div>
+    <div>
+        <div>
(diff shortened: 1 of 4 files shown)


tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1476,6 +1476,152 @@ function testTwo()
...
+            <div>
+                <input id="payment_method_<?php echo esc_attr($gateway->id); ?>"
+                    type="radio"
+                    class="input-radio"
+                    name="payment_method"
+                    value="<?php echo esc_attr($gateway->id); ?>"
+                    <?php checked($gateway->chosen, true); ?>
+                    data-text="<?php echo esc_attr(
+                        $gateway->order_button_text
+                    ); ?>"
```
evilebottnawi: let's break group in this case in other PR, here only alignment

PR #545, opened 2018-08-17
Fix better print binary

evilebottnawi:
:star: :star: :star: 

All new examples will be tested on prettier playground and all output same as prettier

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -445,7 +445,9 @@ function isProgramLikeNode(node) {
   return ["program", "declare", "namespace"].includes(node.kind);
 }
 
-function getNodeKindIncludeLogical(node) {
+// Return `logical` value for `bin` node containing `||` or `&&` type otherwise return kind of node.
+// Require for grouping logical and binary nodes in right way.
+function getNodeKindIncludingLogical(node) {
   if (node.kind === "bin" && ["||", "&&"].includes(node.type)) {
     return "logical";
   }
@@ -478,5 +480,5 @@ module.exports = {
   getFirstNestedChildNode,
   getLastNestedChildNode,
   isProgramLikeNode,
-  getNodeKindIncludeLogical
+  getNodeKindIncludingLogical
 };
--- a/src/util.js
+++ b/src/util.js
@@ -66,11 +66,17 @@ function getPrecedence(op) {
 }
 
 const equalityOperators = ["==", "!=", "===", "!==", "<>", "<=>"];
+const additiveOperators = ["+", "-"];
 const multiplicativeOperators = ["*", "/", "%"];
 const bitshiftOperators = [">>", "<<"];
 
 function shouldFlatten(parentOp, nodeOp) {
   if (getPrecedence(nodeOp) !== getPrecedence(parentOp)) {
+    // x + y % z --> (x + y) % z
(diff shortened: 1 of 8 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -153,17 +153,11 @@ $bool = /*Comment */ true /*Comment */ || /*Comment */ false /*Comment */;
...
-    ) ||
+    ((is_array($this->modelData) || $this->modelData instanceof \\Countable) &&
+        0 === count($this->modelData)) ||
     // traversables that are not countable
-    (
-        $this->modelData instanceof \\Traversable &&
-            0 === iterator_count($this->modelData)
-    );
+    ($this->modelData instanceof \\Traversable &&
+        0 === iterator_count($this->modelData));
```
evilebottnawi: Looks great :+1:

tests/retif/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -63,10 +63,8 @@ return ($testReallyReallyLong >= 1)
     ? $someOtherReallyReallyLongVariable
     : $thisOtherVariable;
 return true ?: 2;
-return (
-    $testReallyReallyLong === $someOtherReallyReallyLongVariable &&
-        $otherTest != $someOtherTest
-)
+return ($testReallyReallyLong === $someOtherReallyReallyLongVariable &&
+    $otherTest != $someOtherTest)
```
evilebottnawi: as prettier

tests/bin/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -166,6 +244,196 @@ $value =
...
+}
+if (
+    $someReallyReallyReallyLongBooleanVariable +
+    $someReallyReallyReallyLongBooleanVariable % 100
+) {
+}
+if (
+    $someReallyReallyReallyLongBooleanVariable %
+        $someReallyReallyReallyLongBooleanVariable +
+    100
```
evilebottnawi: Before we don't add indent and some cases looks very ugly

src/util.js
```diff
@@ -429,6 +445,14 @@ function isProgramLikeNode(node) {
   return ["program", "declare", "namespace"].includes(node.kind);
 }
 
+function getNodeKindIncludeLogical(node) {
```
czosel: I'm not so happy with the name of this function, but after reading what it does it seems hard to come up with a better name. Is the naming from prettier core?

tests/comments/__snapshots__/jsfmt.spec.js.snap
czosel: :100:

czosel:
:+1: Awesome work!

src/util.js
evilebottnawi: @czosel no, `babylon` separate operators on `bin` and `logical`, in php all operators are `bin`

evilebottnawi: Anyway if somebody can find better name - PR welcome. We can change name in any time.

czosel: Ok... a minor improvement might be `getNodeKindIncludingLogical` and a little comment saying pretty much what you just commented?

PR #549, opened 2018-08-29
feat: better print `list` node

evilebottnawi:
Note:
- Use `array` logic for print `list`, because `list` is same as array, also in next version `php-parser` `[,$a,$b,$c] = $arr` is `list` node with `shortForm`

> Like array(), this is not really a function, but a language construct. list() is used to assign a list of variables in one operation.

From http://php.net/manual/en/function.list.php

- Fixed problem with leading `,` as in prettier
- Rename `trailing_commas` to `array`
- `list` support `trailing_commas` with 5 version as `array`
- Respect newlines in `list`
- Respect original input for `list` (move on newline or not based on original input) as in prettier


evilebottnawi:
Merge as is to continue work, feel free to feedback :+1:

PR #551, opened 2018-08-30
feat: update parser

### LINKED ISSUE #541: update `php-parser` to `3.0.0-alpha.3`
Need do this for easy migrate between alpha releases

evilebottnawi:
fixes #541 Wait review to continue work on update `3.0.0-prerelease.1`

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -235,7 +235,7 @@ function isFirstChildrenInlineNode(path) {
  * appear as function arguments or array elements
  */
 function docShouldHaveTrailingNewline(path) {
-  return ["entry"].includes(path.getParentNode().kind);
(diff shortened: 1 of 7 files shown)


src/util.js
```diff
@@ -235,7 +235,7 @@ function isFirstChildrenInlineNode(path) {
  * appear as function arguments or array elements
  */
 function docShouldHaveTrailingNewline(path) {
-  return ["entry"].includes(path.getParentNode().kind);
+  return ["array"].includes(path.getParentNode().kind);
```
czosel: `return path.getParentNode().kind === "array"`

evilebottnawi: @czosel in next parser update we will have not only `array` (i know :smile: ), let's left this as is here :+1:

PR #553, opened 2018-08-31
fix: closure by ref

evilebottnawi:
bug :smile:


evilebottnawi:
@czosel I know about a couple of bugs :smile: Will fix it in near future

PR #555, opened 2018-08-31
fix: don't add extra newline in last inline node

evilebottnawi:
this bug also exists in https://github.com/prettier/plugin-php/pull/550

### DIFF
--- a/tests/markdown/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/markdown/__snapshots__/jsfmt.spec.js.snap
@@ -158,7 +158,6 @@ if (true) {
 }
 echo 'bar';
 ?>
-test
-\`\`\`
+test\`\`\`
 
 `;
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Looks all code in markdown now interpreted as inline :confused: It is big bug

tests/markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -158,7 +158,6 @@ if (true) {
 }
 echo 'bar';
 ?>
-test
-\`\`\`
+test\`\`\`
```
evilebottnawi: Bug in prettier https://github.com/prettier/prettier/issues/5037

evilebottnawi:
Merge as is, we can't fix prettier bug here, feel free to feedback

PR #556, opened 2018-08-31
fix: empty markdown code block

evilebottnawi:
bug


evilebottnawi:
Today a lot of work, feel free to feedback, merge as is for continue work

PR #558, opened 2018-08-31
fix: extra space in namespace without name

evilebottnawi:
ugly


evilebottnawi:
Feel free to feedback

PR #559, opened 2018-08-31
feat: markdown support

evilebottnawi:
Awesome :star: 

We support:
```php
$a=1;
```

```php
<?php
$a=1;
```

Some people use `<?php`, some not, let's support both


evilebottnawi:
Feel free to feedback :+1: Need merge to continue work

PR #563, opened 2018-08-31
feat: better print `traituse` node

evilebottnawi:
:star:

### DIFF
--- a/tests/traits/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/traits/__snapshots__/jsfmt.spec.js.snap
@@ -86,7 +86,9 @@ class ImplementingClass
 
 class ShortImplementingClass
 {
-    use testTrait { A::testFunction insteadof C; }
+    use testTrait {
+        A::testFunction insteadof C;
+    }
 }
 
 class Partial
--- a/src/printer.js
+++ b/src/printer.js
@@ -2479,17 +2479,17 @@ function printNode(path, options, print) {
       return group(
         concat([
           "use ",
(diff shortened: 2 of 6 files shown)


tests/traits/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -86,7 +86,9 @@ class ImplementingClass
 
 class ShortImplementingClass
 {
-    use testTrait { A::testFunction insteadof C; }
+    use testTrait {
+        A::testFunction insteadof C;
+    }
```
evilebottnawi: We always add `newline` after `{` (in function, class, closure and etc), let's do this here for consistently output

evilebottnawi: I am wrong here, better use 
```
use testTrait {A::testFunction insteadof C;}
```

evilebottnawi: From PSR-12:
> When using the insteadof and as operators they must be used as follows taking note of indentation, spacing and new lines.

https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#42-using-traits

czosel: :+1: Which is eaxtly what you implemented now, right?

evilebottnawi: @czosel Yep :+1:

PR #565, opened 2018-08-31
refactor: lookup nodes

evilebottnawi:
And more tests


src/printer.js
```diff
@@ -2091,9 +2091,8 @@ function printNode(path, options, print) {
               firstNonMemberParent.left.kind !== "variable"))) ||
         node.kind === "offsetlookup" ||
         (node.what.kind === "variable" &&
-          (node.offset &&
-            (node.offset.kind === "constref" ||
-              node.offset.kind === "variable")) &&
+          (node.offset.kind === "constref" ||
```
evilebottnawi: offset always exists in lookup nodes

PR #569, opened 2018-09-03
feat: better print for/foreach comments

evilebottnawi:
Logic from prettier


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -993,18 +1019,41 @@ foreach ($iterable as $value) {
...
+    as /* Comments */ $value /* Comments */
+) {
+}
+
+foreach (
+    /* Comments */ $iterable /* Comments */
+    as /* Comments */ $key /* Comments */ =>
+        /* Comments */ $value /* Comments */
+) {
+}
```
evilebottnawi: Now we respect comments inside `foreach` and between key/values

PR #571, opened 2018-09-03
feat: better print `assign` comments

evilebottnawi:


src/comments.js
```diff
@@ -8,7 +8,8 @@ const {
   isNextLineEmpty
 } = require("prettier").util;
 const { concat, join, indent, hardline } = require("prettier").doc.builders;
-
+// TODO: use shared utils after resolve https://github.com/prettier/prettier/pull/5049
+const { hasNewline } = require("prettier/src/common/util");
```
czosel: Hmm, does this work when the plugin is installed in packaged form via npm?
```bash
$ ls node_modules/prettier/
bin-prettier.js    parser-flow.js      parser-parse5.js      parser-yaml.js
index.js           parser-glimmer.js   parser-postcss.js     README.md
package.json       parser-graphql.js   parser-typescript.js  standalone.js
parser-babylon.js  parser-markdown.js  parser-vue.js         third-party.js
```

evilebottnawi: @czosel yep you are right, we should move this logic into plugin right now and when delete this utils

PR #575, opened 2018-09-06
test: parens in default function argument

evilebottnawi:
Just tests


evilebottnawi:
Feel free to feedback

PR #576, opened 2018-09-06
fix: add newline after heredoc and nowdoc parameter

evilebottnawi:
big bug, also better print their before `=`


evilebottnawi:
Feel free to feedback

PR #577, opened 2018-09-06
feat: better print parameter node

evilebottnawi:
:star:


evilebottnawi:
bug in prettier (very rare use case), we can merge this, feel free to feedback

PR #579, opened 2018-09-06
feat: avoid unnecessary parens for top level children

evilebottnawi:


evilebottnawi:
Feel free to feedback

PR #581, opened 2018-09-06
feat: avoid unnecessary parens around new node

evilebottnawi:


evilebottnawi:
Feel free to feedback

src/needs-parens.js
```diff
@@ -38,8 +40,16 @@ function needsParens(path) {
...
+        default:
+          return false;
+      }
+    case "new": {
+      if (isLookupNode(parent)) {
+        return true;
       }
-    // else fallthrough
+
+      return false;
```
czosel: `return isLookupNode(parent)`

evilebottnawi: @czosel i use `false` for future, we only start print parens better, code change very often, i refactor all code after complete main improvement :+1:

PR #583, opened 2018-09-06
refactor: use print argument logic for `isset`, `unset` and `empty` nodes

evilebottnawi:


evilebottnawi:
Feel free to feedback

PR #585, opened 2018-09-07
fix: parens around call and lookup nodes

evilebottnawi:


evilebottnawi:
Feel free to feedback

PR #589, opened 2018-09-10
chore: update `php-parser` to `3.0.0-prerelease.4`

evilebottnawi:
:star:


tests/markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -158,6 +161,7 @@ if (true) {
 }
 echo 'bar';
 ?>
-test\`\`\`
+test
+\`\`\`
```
evilebottnawi: Changes above expected because prettier fix this problem

evilebottnawi:
Feel free to feedback, merge as is to continue work

PR #592, opened 2018-09-10
feat: avoid unnecessary parens for `call` node

evilebottnawi:


tests/parens/call.php
```diff
@@ -0,0 +1,43 @@
...
+$var = (($this)->$name());
+
+$var = Foo::call();
+$var = (Foo::call());
+
+$var = (array("Foo", "bar"))();
+$var = (array(new Foo, "baz"))();
+$var = ((string) 1234)();
+$var = "Foo::bar"();
+$var = ("Foo::bar")();
```
evilebottnawi: I did not know that in php it is possible :smile:

evilebottnawi:
Feel free to feedback :+1:

PR #597, opened 2018-09-11
feat: avoid unnecessary parens around cast node

evilebottnawi:
We always print parens with `cast` into `bin` nodes to avoid this:
```php
(int) '1' + '2'; 
```
Why? Very misleading what developer mean:
- `((int) '1') + '2'`;
- `(int) ('1' + '2');`


src/util.js
```diff
@@ -302,7 +302,8 @@ function lineShouldEndWithSemicolon(path) {
     "number",
     "nowdoc",
     "encapsed",
-    "variable"
+    "variable",
+    "cast"
```
evilebottnawi: Also fix bug for single `cast`

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1317,7 +1317,7 @@ if (/* Comments */ true /* Comments */) {
 } else {
 }
 
-if (2 /* MB_OVERLOAD_STRING */ & (int) ini_get('mbstring.func_overload')) {
+if (2 /* MB_OVERLOAD_STRING */ & ((int) ini_get('mbstring.func_overload'))) {
```
evilebottnawi: Expected. Look on this `if ((int) $a > $b) {}`
Very meas leading too.
- `if (((int) $a) > $b) {}`
- `if ((int) ($a > $b)) {}`

evilebottnawi: Also `(` and `)` add readability in conditions.

evilebottnawi:
Feel free to feedback. Continue work

PR #599, opened 2018-09-11
Simplify conditions

sroze:
Just because... it's simpler that way :)


evilebottnawi:
@sroze thanks for PR, right now we prettify parens output and refactoring is worth doing later

evilebottnawi:
Already done in master, thanks!

PR #600, opened 2018-09-11
feat: avoid unnecessary parens around `clone` nodes

evilebottnawi:


src/util.js
```diff
@@ -303,7 +303,8 @@ function lineShouldEndWithSemicolon(path) {
     "nowdoc",
     "encapsed",
     "variable",
-    "cast"
+    "cast",
+    "clone"
```
evilebottnawi: Fix bug with single `clone`

PR #601, opened 2018-09-11
feat: avoid unnecessary parens around `isset` and `empty` nodes

evilebottnawi:


evilebottnawi:
Feel free to feedback

PR #603, opened 2018-09-11
feat: avoid unnecessary parens around `closure` node

evilebottnawi:


src/printer.js
```diff
@@ -1778,6 +1778,7 @@ function printNode(path, options, print) {
         !node.body.comments;
 
       return concat([
+        node.isStatic ? "static " : "",
```
evilebottnawi: Also fix big bug

tests/parens/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -800,6 +800,78 @@ $var = (clone $var->foo)->foo;
...
+};
+$func = static function () {
+};
+
+function foo()
+{
+    $var = function () {
+    };
+    $var = (function () {
+    })();
```
evilebottnawi: Look ugly, but not related to PR, let's solve this in other PR

evilebottnawi:
Feel free to feedback

Issue #611, opened 2018-09-11
parens around include


evilebottnawi:
```php
$var = include './foo.php';
// VS
$var = (include './foo.php');
```

:+1: for first
:heart: for second

QWp6t:
Only when it's used as assignment like that. Otherwise no parentheses.

austinpray:
Why tho

evilebottnawi:
@austinpray ?

evilebottnawi:
Right now we use `first` example, let's close issue. Feel free to feedback.

PR #621, opened 2018-09-12
refactor: need parens code

evilebottnawi:
Also add tests for include and fix behavior for `bin` node, here issue about how better should looks `print` node https://github.com/prettier/plugin-php/issues/611

### DIFF
--- a/src/needs-parens.js
+++ b/src/needs-parens.js
@@ -39,8 +39,6 @@ function needsParens(path) {
   }
 
   switch (node.kind) {
-    case "call":
-      return false;
     case "pre":
     case "post":
       if (parent.kind === "unary") {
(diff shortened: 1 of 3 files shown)


src/needs-parens.js
```diff
@@ -161,43 +164,13 @@ function needsParens(path) {
...
-      if (parent.kind === "cast") {
-        return true;
-      }
-    }
+    case "include":
+      return parent.kind === "bin";
   }
 
-  return node.parenthesizedExpression;
+  return false;
```
evilebottnawi: By default we don't print parens, it is allow to catch all problems with parens, `prettier` use same logic https://github.com/prettier/prettier/blob/master/src/language-js/needs-parens.js#L604

evilebottnawi:
Feel free to feedback

PR #622, opened 2018-09-12
feat: parens around bin node

evilebottnawi:
I could have missed something because it's really hard, just create a new issue if you find that I made a mistake

### DIFF
--- a/tests/echo/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/echo/__snapshots__/jsfmt.spec.js.snap
@@ -276,7 +276,7 @@ echo 1;
 echo 1;
 echo 1, 1;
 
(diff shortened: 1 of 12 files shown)


evilebottnawi:
Wait feedback

tests/echo/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -276,7 +276,7 @@ echo 1;
 echo 1;
 echo 1, 1;
 
-echo "The sum is " . 1 | 2; // output: "2". Parentheses needed.
+echo ("The sum is " . 1) | 2; // output: "2". Parentheses needed.
```
czosel: :+1:

PR #624, opened 2018-09-12
fix: `nowdoc/heredoc` newline

### LINKED ISSUE #618: broken echo output with heredoc/nowdoc
Input:
```php
echo <<<END
This uses the "here document" syntax to output
multiple lines with $variable interpolation. Note
that the here document terminator must appear on a
line with just a semicolon. no extra whitespace!
END
,
    'test';
```

Output:
```php
echo <<<END
This uses the "here document" syntax to output
multiple lines with $variable interpolation. Note
that the here document terminator must appear on a
line with just a semicolon. no extra whitespace!
END,
    'test';
```

`,` should be on next line

evilebottnawi:
fixes #618

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -241,7 +241,36 @@ function isFirstChildrenInlineNode(path) {
  * appear as function arguments or array elements
  */
 function docShouldHaveTrailingNewline(path) {
-  return ["array", "parameter"].includes(path.getParentNode().kind);
+  const node = path.getValue();
+  const parent = path.getParentNode();
+
+  if (!parent) {
+    return false;
+  }
+
+  if (parent.kind === "echo") {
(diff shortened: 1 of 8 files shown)


src/util.js
```diff
@@ -241,7 +241,36 @@ function isFirstChildrenInlineNode(path) {
...
+
+  if (parent.kind === "parameter") {
+    const parentParent = path.getParentNode(1);
+    const lastIndex = parentParent.arguments.length - 1;
+    const index = parentParent.arguments.indexOf(parent);
+
+    return index !== lastIndex;
+  }
+
+  return false;
```
evilebottnawi: Avoid extra newline if `nowdoc/heredoc` is last

PR #625, opened 2018-09-12
fix: bin string concatenation

evilebottnawi:
my bug :disappointed:


evilebottnawi:
Feel free to feedback

PR #626, opened 2018-09-12
fix: parens for equal bin operator

evilebottnawi:
My bug too :disappointed:


evilebottnawi:
Feel free to feedback

tests/parens/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -599,11 +603,11 @@ $var = $var += 10;
 
 $var = 10 + 20 + 30;
 $var = 10 + 20 + 30;
-$var = 10 + 20 + 30;
+$var = 10 + (20 + 30);
```
czosel: From a mathematical perspective, the braces could be removed here because `+` is associative. Is prettier for JS implementing that?

aboyton: Mathematically it is associative, often in programming languages with floating point numbers it is not (see for example this example I posted about JavaScript https://github.com/prettier/prettier/issues/187#issuecomment-385807181). I still find it really weird to have them there and I did a maths degree :)

czosel: Nevermind, turns out floating point addition is not always associative (see [here](https://github.com/prettier/prettier/issues/187#issuecomment-385807181)).

PR #627, opened 2018-09-12
fix: broken `retif` without parens

evilebottnawi:
Let's output `parens` for all `nested` `retif`, ternary in `js` is difference than in `php` and it is really difficult to read their without `parens`.

But we can avoid parens for `test` and `trueExpr` properties, but i find this very ugly.

Without `trueExpr`:
```php
true ? 1 ? 2 : 3 : (4 ? 5 : 6); // Inconsistently and ugly
// vs
true ? (1 ? 2 : 3) : (4 ? 5 : 6);
```

Without `test`:
```php
1 ? 2 : 3 ? (1 ? 2 : 3) : (4 ? 5 : 6); // Looks ugly
vs
(1 ? 2 : 3) ? (1 ? 2 : 3) : (4 ? 5 : 6);
```

### DIFF
--- a/tests/retif/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/retif/__snapshots__/jsfmt.spec.js.snap
@@ -90,26 +90,26 @@ $category_color = get_field(Category_Meta::COLOR, 'category_' . $term_id)
     ?: 'gold';
 
 $var = ($someOtherReallyReallyLongVariable
-        ? $someOtherReallyReallyLongVariable
+        ? ($someOtherReallyReallyLongVariable
             ? $someOtherReallyReallyLongVariable
-            : $someOtherReallyReallyLongVariable
+            : $someOtherReallyReallyLongVariable)
         : $someOtherReallyReallyLongVariable)
     ? $someOtherReallyReallyLongVariable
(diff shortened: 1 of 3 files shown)


tests/retif/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -171,28 +171,28 @@ $var = 1
...
+    ?: (true
         ? $someOtherReallyReallyReallyReallyReallyReallyReallyReallyReallyReallyLongVariable
-        : 2;
+        : 2);
 $var = $test
-    ?: true
+    ?: (true
         ? 1
-        : $someOtherReallyReallyReallyReallyReallyReallyReallyReallyReallyReallyLongVariable;
+        : $someOtherReallyReallyReallyReallyReallyReallyReallyReallyReallyReallyLongVariable);
```
evilebottnawi: `falseExpr` always should have parens to avoid
```
echo true ?: (true ? 'foo' : 2); // Works as expected
echo true ?: true ? 'foo' : 2; // Works unexpected
```

evilebottnawi:
Feel free to feedback

PR #628, opened 2018-09-13
fix: broken cast in `retif` and `silence` node

evilebottnawi:
My bug, also refactor code and little fix for array


evilebottnawi:
Feel free to feedback

PR #631, opened 2018-09-13
refactor: stuff

evilebottnawi:
reorder print logic for nodes with similar functionality, if somebody know how it can be better PR welcome. Also remove `extractTokens`, it should increase perf. Refactor `continue/break` nodes


evilebottnawi:
Feel free to feedback

PR #632, opened 2018-09-13
fix: identifier printing

evilebottnawi:
rare use case, but we have bug with this


evilebottnawi:
Feel free to feedback

PR #636, opened 2018-09-14
Convert double quotes to single quotes

### LINKED ISSUE #633: Convert double quotes to single
**Input:**

```php
$var = "just string without expressions";
```

**Output:**

```php
$var = "just string without expressions";
```

**Expected behavior:**

```
$var = 'just string without expressions';
```

Only for string without `variables`, `call` and etc.

Feel free to feedback. Add :+1: if you want this.

czosel:
...if the string is "simple", i.e. doesn't use any features that require
double quotes:
* embedded variables
* linebreaks
* special characters like \n, \t, ", ', ...
See http://php.net/manual/en/language.types.string.php#language.types.string.syntax.double

Fixes #633


czosel:
@evilebottnawi The issue and the PR got 5 upvotes combined. Shall we move forward? I'm not sure who to ask anymore :smile:

evilebottnawi:
@czosel I think we should wait a little longer, very small votes

czosel:
@evilebottnawi ok.

evilebottnawi:
@czosel anyway good work, and I hope we will integrate it without negative feedback

czosel:
@evilebottnawi let me know when you think we can merge this, then I'll resolve the conflicts.

vielhuber:
Does this mean prettier-php in the future respects the option --single-quote ? That would be awesome.

czosel:
@vielhuber Well, this feature is still somewhat controversial (see [here](https://github.com/prettier/plugin-php/issues/633#issuecomment-427869095)). Hiding it behind the `--single-quote` (and thereby making it opt-in) might be a good way to introduce it though.
@evilebottnawi What do you think about this?

karptonite:
If it were hidden behind an opt-in option, it is OK with me, although I still think that it is a problematic feature. This is functionally more link converting js backticks ` to single quotes, which the js version of prettier doesn't do.

vielhuber:
Yes, let the people choose if they want to use it.

czosel:
@karptonite While the analogy to JS backticks definitely makes sense, I could imagine that in many PHP projects double quotes are used just like single quotes, even if the extra functionality the add is not needed - in JS that might be different, because

* the backticks were introduced much later (ES6)
* they're not as common as single or double quotes

The main reason to use the option would be to achieve consistent string formatting in projects where single and double quotes were used interchangeably.

src/util.js
```diff
@@ -624,6 +624,27 @@ function hasNewline(text, index, opts) {
...
+function useSingleQuote(node, options) {
+  return (
+    !node.isDoubleQuote ||
+    (options.singleQuote &&
+      !node.raw.match(/\\n|\\t|\\r|\\t|\\v|\\e|\\f/) &&
+      !node.value.match(
+        /["'$\n]|\\[0-7]{1,3}|\\x[0-9A-Fa-f]{1,2}|\\u{[0-9A-Fa-f]+}/
+      ))
+  );
+}
```
evilebottnawi: I think better avoid using `RegExp`, better traverse `nodes` on only string nodes

evilebottnawi:
Also maybe related change `HereDoc` to `NowDoc` (it should be safe using same algorithm), but let's do it another PR and create issue about this.

src/util.js
czosel: `useSingleQuote` is only called on String nodes. I moved it into utils because I thought I'd also need to call it from `clean.js` but it seems that `massageASTNode` doesn't have access to the `options` object...

evilebottnawi: @czosel hm, let's fix it in prettier, should be easy

czosel: Just pushed an alternative implementation that just removes `isDoubleQuote` in `clean.js`. Shall we suggest the change in prettier and refactor in another PR?

evilebottnawi: @czosel yep, let's add TODO

czosel: @evilebottnawi anything else left to be done?

czosel:
@evilebottnawi We also need to start documenting the options we support and what they do exactly, right?

evilebottnawi:
@czosel yep, will be great

vielhuber:
Thanks for the update and keep up the good work.

czosel:
@evilebottnawi Can you finish your review? :wink:

vielhuber:
Hello evilebottnawi,

is there a way to get this feature without having to wait for a release?

Like a nightly build or something.

czosel:
Hi @vielhuber,

I think you should be able to install the current master or a specific commit by adding `prettier/plugin-php#master` or `prettier/plugin-php#06302b8e7c51e9fe378178802f1241c0b16c7fb3` to your dependencies.

vielhuber:
I've installed prettier php with:
```
npm install --global prettier prettier/plugin-php#master
npm update --global prettier prettier/plugin-php#master
```

And also per project like this:
```
npm install prettier/plugin-php#master
```

I go into a php file with VS Code:
```
<?php
$data["foo"] = 'bar' ;
```

Output (php prettier works, but single quotes are not working):

```
<?php
$data["foo"] = 'bar' ;
```

czosel:
Hi @vielhuber,

I think you're missing the `--single-quote` option. See [this playground example](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8MCIKgt9beLJUBhdAFtrbAAoAJe-hCYsulhwhsAxlABlGABPNVcwbAgAWgAmEBZoAEdBeGh7H2U7fBl5IxBiEQgYdLZMtmy5EQBfFigKeBh0FDR4LElQNmwYCtya-MwwQW68EAAeAH4wajAAHUw5gBIKNnYAbRmQKXR0DYBdAAIAXn2Acg42CBP9gG4QKqA).

Support for `--single-quote` has been released meanwhile, so you can use the latest release instead of `master`.

vielhuber:
Hello!

Thanks for your answer. How can I add this option? I don't use RunOnSave or any other plugin in VS Code. How can I use this command line option there?

czosel:
I don't use prettier-vscode myself, but it there are [serveral ways](https://github.com/prettier/prettier-vscode#configuration) to set options.

vielhuber:
I have "prettier.singleQuote": true, in my "settings.json" of vscode and in JS files it works.
PHP prettier ignores this setting.

czosel:
Strange, I just tried it and it works for me, both using `.prettierrc` as well as with the "Prettier: Single Quote" setting of prettier-vscode.

Some infos about my setup:
prettier (local): 1.19.1
prettier-vscode (local): 0.11.2
vscode: 1.40.2
prettier-vscode: 3.11.0

Did you already try using a `.prettierrc` file?

vielhuber:
Sorry for the confusion.

It seems that PHP IntelliSense has formatted my documents all the time without knowing it.

```
"editor.defaultFormatter": "esbenp.prettier-vscode",
```

fixed this.

PR #638, opened 2018-09-17
refactor: comments

evilebottnawi:
As in prettier https://github.com/prettier/prettier/blob/master/src/language-js/comments.js

It looks like we are already close to a stable release, so my roadmap:
- Improve comments output on this week.
- Improve `inline` output on next week.

Also refactoring/fixes edge cases/maybe update parser to master

And we can do stable release

### DIFF
--- a/src/comments.js
+++ b/src/comments.js
@@ -27,54 +27,59 @@ args:
   isLastComment
 */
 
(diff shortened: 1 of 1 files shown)


evilebottnawi:
Merge as is feel free to feedback

czosel:
Awesome, can't wait! :tada: 
Refactoring looks good :+1:

PR #639, opened 2018-09-17
feat: better print variable comments

evilebottnawi:
From here https://github.com/prettier/prettier/blob/master/src/language-js/comments.js#L690


evilebottnawi:
Feel free to feedback

PR #640, opened 2018-09-17
feat: better handle entry comments

evilebottnawi:
From prettier, same logic for `{prop: value}` handlePropertyComments


evilebottnawi:
Feel free to feedback

PR #646, opened 2018-09-17
feat: better print comments in `if` node

evilebottnawi:


evilebottnawi:
Some comments are ugly, let's improve their in future

PR #652, opened 2018-09-19
refactor: anonymous class

evilebottnawi:
Without global variable, because it can be dangerous

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -761,16 +761,6 @@ function printBinaryExpression(
   return parts;
 }
 
-// so this is a bit hacky, but for anonymous classes, there's a chance that an
-// assumption core to prettier will break - that child nodes will not overlap. In
-// this case, if we have something like this:
-//   $test = new class($arg1, $arg2) extends TestClass {};
-// we end up with a parent "new" node, which has children "arguments", and "what"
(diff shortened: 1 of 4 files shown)


src/printer.js
```diff
@@ -1932,13 +1922,14 @@ function printNode(path, options, print) {
       ]);
     }
     case "new": {
-      // if the child node is an anonymous class, we need to store the arguments
-      // so the child class node can access them later
+      // TODO: maybe need rewrite after resolve https://github.com/glayzzle/php-parser/issues/187
```
evilebottnawi: Also create issue

evilebottnawi:
Also more test, feel free to feedback.

tests/new/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -202,4 +224,38 @@ EOF
...
+    $arg,
+    'string',
+    2100,
+    $var ? true : false,
+    $other_arg,
+    function () {
+        return 1;
+    }
+) extends SomeClass implements SomeInterface
+{
```
evilebottnawi: Looks very ugly, let's fix it in other PR

PR #669, opened 2018-09-24
feat: consistently print assign node with lookup nodes

evilebottnawi:
Why? Because `::` in js is not same in `php`

### DIFF
--- a/tests/member_chain/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/member_chain/__snapshots__/jsfmt.spec.js.snap
@@ -834,8 +834,9 @@ $page = TableRegistry::insertReallyReallyReallyLongName('Pages')
     ::insertReallyReallyReallyLongName();
 
 $page = TableRegistry::insertReallyReallyReallyLongName[0];
-$page = TableRegistry
-    ::insertReallyReallyReallyReallyReallyReallyReallyReallyReallyLongName[0];
+$page =
+    TableRegistry
+        ::insertReallyReallyReallyReallyReallyReallyReallyReallyReallyLongName[0];
 $page = TableRegistry::insertReallyReallyReallyReallyReallyReallyReallyReallyReallyLongName()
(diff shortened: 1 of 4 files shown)


tests/member_chain/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -834,8 +834,9 @@ $page = TableRegistry::insertReallyReallyReallyLongName('Pages')
     ::insertReallyReallyReallyLongName();
 
 $page = TableRegistry::insertReallyReallyReallyLongName[0];
-$page = TableRegistry
-    ::insertReallyReallyReallyReallyReallyReallyReallyReallyReallyLongName[0];
+$page =
+    TableRegistry
+        ::insertReallyReallyReallyReallyReallyReallyReallyReallyReallyLongName[0];
```
evilebottnawi: Look weird, but prettier do [same](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEASSBbADtBMAEAvPqgG4CGATgDpT734B0AllAM5yUwBKc5ANvwCevAcNGCRfSRPHS5YqYtlKZ81QrXL1AGWgBzAHLkMcANoAGALoBuEABoQELDGbQ2yUFUoQA7gAUqBA8UclIIZgATBxAAI0pyMABrOBgAZSxE1n1kGEoAVzhHVg4ufwT9DHJkADMBDkcAKzYADwAhBOTUtJM4HVY4WvqikEzKUuQ48lihfmgYrEpWGAB1KJgAC2QADgtHRYgOFYSsScW4UtJBx0o4AEd85lvy8krqpDr+BpAODGZcgojNjZfhwACK+Qg8CGXxGMGma0im2QABZHHlyMx+NkAMIQDBVSZQXAxfIcAAq0xCnw4AF9aUA), also as i said above `::` in js is not same in php, `::`(staticlookup) closer to `->` (propertylookup) in php

PR #670, opened 2018-09-24
refactor: print declaration block

evilebottnawi:
We can get `node` from `path`, no need pass this as argument


evilebottnawi:
Feel free to feedback

Issue #653, opened 2018-09-19
ugly anonymous classes 


evilebottnawi:
**Prettier 1.14.2**

**PHP Plugin 0.6.0**

**Input:**

```php
$var = new class(
    $arg,
    'string',
    2100,
    $var ? true : false,
    $other_arg,
    function () {
        return 1;
    }
) extends SomeClass implements SomeInterface {
    public function log($msg)
    {
        echo $msg;
    }
};
```

**Output:**

```php
$var = new class(
    $arg,
    'string',
    2100,
    $var ? true : false,
    $other_arg,
    function () {
        return 1;
    }
) extends SomeClass implements SomeInterface 
{
    public function log($msg)
    {
        echo $msg;
    }
};
```

**Expected behavior:**

Expected as is, because `{` on same line is ok for anonymous classes

evilebottnawi:
I think i wrong here.

From PSR-12:
> The opening brace MAY be on the same line as the class keyword so long as the list of implements interfaces does not wrap. If the list of interfaces wraps, the brace MUST be placed on the line immediately following the last interface.

I think this rule applies to arguments too. 

I.e. if between `class` keyword and class body something on next line `{` should be start with new line.

czosel:
I agree. I'd interpret the rule like this: If the class definition spans multiple lines, the brace MUST be placed on a new line. That's also fitting to the two examples mentioned in [PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#8-anonymous-classes).

evilebottnawi:
@czosel yep, let'c close issue, if someone has a different opinion feel free to create new issue

PR #671, opened 2018-09-24
refactor: print `class`, `interface` and `trait` nodes

evilebottnawi:
We should separate functions for print `class`/class like nodes (`interface` and `trait`) and function/function like nodes (`closure` and `method`). Will be done in next PR. It is first part.

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1026,7 +1026,164 @@ function printLines(path, options, print, childrenAttribute = "children") {
   return concat(wrappedParts);
 }
 
-function printDeclarationBlock({
+function printClass(path, options, print) {
+  const node = path.getValue();
+  const parentNode = path.getParentNode();
+
+  const declaration = [];
+
+  if (node.isFinal) {
+    declaration.push("final ");
+  }
+
+  if (node.isAbstract) {
+    declaration.push("abstract ");
+  }
+
+  declaration.push(node.kind);
+
+  if (node.name) {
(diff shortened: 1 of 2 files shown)


src/printer.js
```diff
@@ -1026,7 +1026,165 @@ function printLines(path, options, print, childrenAttribute = "children") {
...
+                        comments.printDanglingComments(
+                          implementPath,
+                          options,
+                          true
+                        ),
+                        hardline,
+                        print(implementPath)
+                      ])
+                    : concat([line, print(implementPath)]);
+                }, "implements")
```
evilebottnawi: Now we respect comments in implements and extends for `class/interface/trait`, tests will add in other PR

tests/class/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -176,8 +176,10 @@ $instance = new class extends \\Foo implements \\HandleableInterface {
 };
 
 // Parenthesis on the next line
-$instance = new class extends \\Foo
-    implements \\ArrayAccess, \\Countable, \\Serializable
+$instance = new class extends \\Foo implements
+        \\ArrayAccess,
+        \\Countable,
+        \\Serializable
```
evilebottnawi: Partial fix #653, but not full

evilebottnawi:
Feel free to feedback, merge to continue work, want done refactoring today :+1:

PR #673, opened 2018-09-24
refactor: standardize function arguments

evilebottnawi:
Part 3. Why?

Main print function (`genericPrint(path, options, print)`) uses this arguments order, let's use this order everywhere.


evilebottnawi:
Feel free to feedback

Issue #657, opened 2018-09-21
bug with `identifier`


evilebottnawi:
**Prettier 1.14.3**

**PHP Plugin 0.7.0**

**Input:**

```php
int();
iNt();
```

**Output:**

```php
int();
int();
```

**Expected behavior:**

```php
int();
iNt();
```

Discussion https://github.com/glayzzle/php-parser/issues/188

PR #674, opened 2018-09-24
fix: do not break code with `self`, `static` and `parent` identifier

evilebottnawi:
Partial #657


evilebottnawi:
Feel free to feedback

PR #676, opened 2018-09-27
test: ref in `foreach` node

evilebottnawi:
Just test.

Compare https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/PrettyPrinter/Standard.php with our output (also create issue for new node/modify in our parser for better printing).


evilebottnawi:
Feel free to feedback

PR #677, opened 2018-09-27
refactor: `traitalias` node

evilebottnawi:


src/printer.js
```diff
@@ -1514,8 +1514,7 @@ function printNode(path, options, print) {
       ]);
     case "traitalias":
       return concat([
-        path.call(print, "trait"),
-        node.trait ? "::" : "",
+        node.trait ? concat([path.call(print, "trait"), "::"]) : "",
```
evilebottnawi: not necessary execute `path.call(print, "trait")` when `trait` is `null`

evilebottnawi:
Feel free to feedback

Issue #167, opened 2018-03-06
Better output `extends` and `implements`


evilebottnawi:
From PSR-12: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#41-extends-and-implements

```php
class ClassName extends ParentClass implements
    \ArrayAccess,
    \Countable,
    \Serializable
{
    // constants, properties, methods
}
```
vs
```php
class ClassName 
    extends ParentClass 
    implements \ArrayAccess, \Countable, \Serializable
{
    // constants, properties, methods
}
```

In php __ONLY__ one class maybe extends. Vote welcome.
- :+1: - First example
- :heart: - Second example

czosel:
> Lists of implements and extends MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one interface per line.

Funny that PSR-12 speaks of "lists of extends" when there can be only one. I'd opt for the first style as because both styles look fine but the first seems to be preferred by PSR-12.

mgrip:
Should we consider preferring option 1 but breaking to option 2 if we hit the line length limit?

czosel:
Sounds good! That would probably mean that the "worst case" looks like this, right?
```php
class ClassNameLoooooooooooooooooooooooooooooooooong
    extends ParentClassLoooooooooooooooooooooooooooooooooong
    implements 
        \ArrayAccessLooooooooooooooooooong, 
        \CountableLooooooooooooooooooong, 
        \SerializableLooooooooooooooooooong
{
    // constants, properties, methods
}
```

evilebottnawi:
@mgrip @czosel how about open issue for https://github.com/php-fig/fig-standards about this. I am confused about this. Because extends class may have very long value and PSR-12 output is ugly

evilebottnawi:
Not fixed fully

PR #684, opened 2018-09-28
feat: better print multiple extends in interface

evilebottnawi:
Partial #167


tests/interface/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -38,4 +44,21 @@ interface EmptyInterfaceWithComments
...
+{
+}
+
+interface Foo extends
+    MyClass,
+    MyOtherClass,
+    MyOtherOtherOtherClass,
+    MyVeryVeryVeryLongClassName
+{
+}
```
evilebottnawi: As say PSR-12, `extends` on same line here class

PR #686, opened 2018-09-30
Remove extra braces in control structures containing and, or, xor

### LINKED ISSUE #685: Extra braces in ifexpr with or
I just tried updating my codebase to `0.7.0` and noticed this:

**Input:**

```php
if (!$foo or $bar == -1) {}
```

**Output:**

```php
if ((!$foo or $bar == -1)) {
}
```

**Expected behavior:**
```php
if (!$foo or $bar == -1) {
}
```

czosel:
Fixes #685

### DIFF
--- a/src/needs-parens.js
+++ b/src/needs-parens.js
@@ -38,10 +38,7 @@ function needsParens(path) {
   }
 
   // Avoid parens in short control structures like `if (expr) statement`
-  if (
-    ["if", "while", "for", "foreach"].includes(parent.kind) &&
-    (parent.body === node || parent.alternate === node)
-  ) {
+  if (["if", "do", "while", "for", "foreach", "switch"].includes(parent.kind)) {
     return false;
   }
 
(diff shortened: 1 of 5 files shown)


src/needs-parens.js
```diff
@@ -38,10 +38,7 @@ function needsParens(path) {
   }
 
   // Avoid parens in short control structures like `if (expr) statement`
-  if (
-    ["if", "while", "for", "foreach"].includes(parent.kind) &&
-    (parent.body === node || parent.alternate === node)
```
czosel: Not sure which case this was protecting, maybe you know more @evilebottnawi?

evilebottnawi:
@czosel it is interesting, i think we should change logic here as said here https://github.com/prettier/prettier/issues/5070, what do you think?

src/needs-parens.js
evilebottnawi: @czosel to avoid parens around
```php
if ($var = 1) ($var = 1);
```

Just uncomment this check and update snapshots and you can see what happens :smile:

czosel: Strange, I don't see that change. In which test file is it?

czosel: Also notice in the changes below that no existing test is changed.

evilebottnawi: @czosel strange, give me time to investigate

evilebottnawi: @czosel rebase PR and try to update snapshot for `parens` directory, you can see how it is break some cases

evilebottnawi:
@czosel Anyway i will WIP on this :+1:

evilebottnawi:
Better solution https://github.com/prettier/plugin-php/pull/698 :+1:

PR #688, opened 2018-10-01
chore: update parser

evilebottnawi:


tests/case/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -291,7 +291,7 @@ final class MyClass extends Bar implements iTemplate
 
     private $private = 'Private';
 
-    public $foo = 'Bar';
+    var $foo = 'Bar';
```
evilebottnawi: Big bug :smile: Original code use `var`, we change `var` on `public`

czosel: :smile:

PR #689, opened 2018-10-01
feat: better print classlike node comments

evilebottnawi:
Handle comments in `interface` and `trait`

### DIFF
--- a/src/comments.js
+++ b/src/comments.js
@@ -195,13 +195,33 @@ function handleForComments(
 }
 
 function handleClassComments(enclosingNode, followingNode, comment) {
-  if (enclosingNode && enclosingNode.kind === "class") {
+  if (
+    enclosingNode &&
+    ["class", "interface", "trait"].includes(enclosingNode.kind)
+  ) {
     // for extends nodes that have leading comments, we can store them as
     // dangling comments so we can handle them in the printer
-    if (followingNode === enclosingNode.extends) {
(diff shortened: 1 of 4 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2246,6 +2301,53 @@ throw new ClassName /* Comment */(
...
+{
+    // Comment
+}
+
+// Comment
+trait tFoo
+    /* Comment */
+    extends ReflectionMethod
+{
+
```
evilebottnawi: Extra empty line is not related to this PR, let's fix it in other PR :+1:

evilebottnawi:
Merge to continue work, feel free to feedback

src/comments.js
```diff
@@ -195,13 +195,33 @@ function handleForComments(
...
-      return true;
+
+    if (followingNode && enclosingNode.extends) {
+      if (!Array.isArray(enclosingNode.extends)) {
+        if (followingNode === enclosingNode.extends) {
+          addDanglingComment(followingNode, comment);
+          return true;
+        }
+      } else {
+        if (
```
czosel: `else if`

evilebottnawi: It will be necessary here to rewrite something here so I will correct in the following PRs

PR #692, opened 2018-10-01
feat: better print `retif` without `true` expression

### LINKED ISSUE #659: ugly retif without true expression
**Prettier 1.14.3**

**PHP Plugin 0.7.0**

**Input:**

```php
$databaseUser = (getenv('DATABASE_USER', true) ?: getenv('MYSQL_USER', true)) ?: 'user';
```

**Output:**

```php
$databaseUser = (getenv('DATABASE_USER', true)
        ?: getenv('MYSQL_USER', true))
    ?: 'user';
```

**Expected behavior:**

```php
$databaseUser = 
    getenv('DATABASE_USER', true) ?: 
    getenv('MYSQL_USER', true) ?: 
    'user';
```

I think better output this as `??` operator.

evilebottnawi:
fixes #659

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1367,8 +1367,9 @@ function printAssignmentRight(leftNode, rightNode, printedRight, options) {
   const canBreak =
     (rightNode.kind === "bin" && !shouldInlineLogicalExpression(rightNode)) ||
     (rightNode.kind === "retif" &&
-      rightNode.test.kind === "bin" &&
-      !shouldInlineLogicalExpression(rightNode.test)) ||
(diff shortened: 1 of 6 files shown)


src/printer.js
```diff
@@ -2347,6 +2291,90 @@ function printNode(path, options, print) {
         ])
       );
     }
+    case "retif": {
```
evilebottnawi: Move `retif` after `bin` for because `?:` and `??` is operators

evilebottnawi:
Merge this as is to continue work, feel free to feedback or create issue if you found something wrong

PR #693, opened 2018-10-01
fix: print retif without true expr

evilebottnawi:
Small fixes


evilebottnawi:
Feel free to feedback

PR #697, opened 2018-10-01
fix: parens around assign in control structures

### LINKED ISSUE #687: parens around `=`
**Prettier 1.14.3**

**PHP Plugin 0.7.0**

**Input:**

```php
if ($var = 1) {
}
```

**Output:**

```php
if ($var = 1) {
}
```

**Expected behavior:**

https://github.com/prettier/prettier/issues/5070#issuecomment-419487668

evilebottnawi:
fixes #687

### DIFF
--- a/src/needs-parens.js
+++ b/src/needs-parens.js
@@ -162,10 +162,7 @@ function needsParens(path) {
         (parent.init.includes(node) || parent.increment.includes(node))
       ) {
         return false;
-      } else if (
-        ["if", "while", "do", "switch"].includes(parent.kind) &&
-        name === "test"
-      ) {
+      } else if (parent.kind === "while" && node.left.kind === "list") {
         return false;
       } else if (parent.kind === "assign") {
         return false;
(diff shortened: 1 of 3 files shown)


src/needs-parens.js
```diff
@@ -162,10 +162,7 @@ function needsParens(path) {
         (parent.init.includes(node) || parent.increment.includes(node))
       ) {
         return false;
-      } else if (
-        ["if", "while", "do", "switch"].includes(parent.kind) &&
-        name === "test"
-      ) {
+      } else if (parent.kind === "while" && node.left.kind === "list") {
```
evilebottnawi: We need allow print `while` without parens in this case:
```
while (list($id, $name, $salary) = $result->fetch(PDO::FETCH_NUM)) {
    echo " <tr>\n" .
          "  <td><a href=\"info.php?id=$id\">$name</a></td>\n" .
          "  <td>$salary</td>\n" .
          " </tr>\n";
}
```

From http://php.net/manual/en/function.list.php.
Many people use this construction.

evilebottnawi:
Feel free to feedback

PR #703, opened 2018-10-02
feat: better print multiple implements

evilebottnawi:
Partial fix https://github.com/prettier/plugin-php/issues/167. In some cases ugly, but we already have a lot of ugly examples, i will fix it in next PR, hard to fix all cases in one PR


evilebottnawi:
Merge to continue work, feel free to feedback

PR #704, opened 2018-10-02
feat: better print extends for class

evilebottnawi:
In some cases still ugly, will be fixed and refactored in next PRs.

Next:
- Better print extends and implements together
- Refactor
- More tests with comments


evilebottnawi:
Merge to continue work

czosel:
:+1:

PR #705, opened 2018-10-02
refactor: print class part

evilebottnawi:
For future union `extends` with one class and multiple


evilebottnawi:
Merge to continue work

Issue #665, opened 2018-09-23
If plain html code is provided, empty lines are added


vielhuber:
I have alot of php files that contain pure html code.

Prettier PHP takes those files and adds plain empty html lines between every line.

Before:
```php
<section class="[ c-social-foot ]  [ has-text-centered ]">
  <div class="[ container  is-fluid ]">
    <div class="[ columns  is-centered ]">
      <div class="[ column ]">
```

After
```php
<section class="[ c-social-foot ]  [ has-text-centered ]">

  <div class="[ container  is-fluid ]">

    <div class="[ columns  is-centered ]">

      <div class="[ column ]">
```

This is very odd, it should remain as it was.

evilebottnawi:
@vielhuber can you provide fill example of file

vielhuber:
1.php: Original file
2.php: File after first save
3.php: File after second save

I'm using Windows 10 Pro and Visual Studio Code 1.27.2.

[1-2-3.zip](https://github.com/prettier/plugin-php/files/2412759/1-2-3.zip)

evilebottnawi:
@vielhuber interesting, how you run `prettier`?

evilebottnawi:
OR maybe you can create minimum reproducible test repo?

vielhuber:
```
npm install --global prettier @prettier/plugin-php
```

vscode config
```
    "emeraldwalk.runonsave": {
        "commands": [
            {
                "match": "\\.php$",
                "cmd": "prettier ${file} --write --print-width 200 --tab-width 4 --single-quote"
            }
        ]
    }
```

vielhuber:
I have NO problems with normal php files.

Have you tested this on windows? I think it is a line break issue.

Myhlamaeus:
@evilebottnawi It seems that (outside of PHP code) every CR is replaced with two CRs. I’ve tried adding tests for this, but it seems that snapshots are normalized to LF-only.

evilebottnawi:
@Myhlamaeus can't reproduce on windows, maybe you have something wrong in editor? Maybe you can record video with all steps?

vielhuber:
here is a video of the behaviour:

![prettier-bug](https://user-images.githubusercontent.com/3183737/46312645-fa196e80-c5c5-11e8-9088-1fcaaabfc0c9.gif)

evilebottnawi:
@vielhuber can you send me original file to `sheo13666q@gmail.com`?

vielhuber:
[...]

PR #708, opened 2018-10-02
test: HTML line endings

Myhlamaeus:
* Make changes in CR characters visible (4c38a0e)
* Add tests for CRLF in HTML (related to #665) (660cbae)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 tests/newline/CRLF.php text eol=crlf
+tests/newline/html-CRLF.php text eol=crlf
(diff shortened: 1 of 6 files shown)


Myhlamaeus:
I would have to remove the part which prints the line endings, then, meaning that changes concerning them would not be visible (or I would have to manually edit all the tests, which seems obnoxious). I don’t know why the previous snapshots are actually correct on Travis, though, as they certainly aren’t for me locally.

evilebottnawi:
@Myhlamaeus because in prettier test we format tests as before, don't change all tests, only what show and fix problem.

evilebottnawi:
@Myhlamaeus Thanks for example and test, WIP on this

evilebottnawi:
Close in favor https://github.com/prettier/plugin-php/pull/712

PR #709, opened 2018-10-03
feat: better print empty anonymous class

evilebottnawi:


evilebottnawi:
Feel free to feedback

PR #712, opened 2018-10-03
fix: extra line in plain html with CRLF

### LINKED ISSUE #665: If plain html code is provided, empty lines are added

evilebottnawi:
fixes #665

### DIFF
--- a/tests_config/run_spec.js
+++ b/tests_config/run_spec.js
@@ -27,7 +27,11 @@ function run_spec(dirname, parsers, options) {
       filename[0] !== "." &&
       filename !== "jsfmt.spec.js"
     ) {
-      const source = read(path).replace(/\r\n/g, "\n");
+      let source = read(path);
+
+      if (!options.keepEOL) {
+        source = source.replace(/\r\n/g, "\n");
+      }
 
       const mergedOptions = Object.assign(mergeDefaultOptions(options || {}), {
         parser: parsers[0]
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 tests/newline/CRLF.php text eol=crlf
+tests/newline/html-CRLF.php text eol=crlf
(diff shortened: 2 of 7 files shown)


tests_config/run_spec.js
```diff
@@ -27,7 +27,11 @@ function run_spec(dirname, parsers, options) {
       filename[0] !== "." &&
       filename !== "jsfmt.spec.js"
     ) {
-      const source = read(path).replace(/\r\n/g, "\n");
+      let source = read(path);
+
+      if (!options.keepEOL) {
```
evilebottnawi: New option for test to allow keep original EOL

evilebottnawi:
Feel free to feedback

vielhuber:
- When will there be a release?

- How can I test this? When I run ````npm update --global prettier/plugin-php```, I still have 0.7.0.

evilebottnawi:
@vielhuber Plan on today:
1. More test with CRLF and fix problems
2. Add appveyor CI
3. Release 0.8.0

czosel:
Is a new option really the way to go here? How is Prettier for JS handling files with windows line endings?

evilebottnawi:
@czosel prettier for js doesn't support html + php, also it is not public option, it is option only for tests

evilebottnawi:
I.e. you can't set this option, just to allow testing CRLF and LF

czosel:
Ah ok, if the option is not public then it's fine :+1:

PR #713, opened 2018-10-03
fix: pseudo inline node with CRLF

evilebottnawi:
I think we have more problem with this in some cases, WIP on this


evilebottnawi:
Feel free to feedback

PR #716, opened 2018-10-03
fix: handle block comment with CRLF

evilebottnawi:


evilebottnawi:
Don't look on appveyor (WIP)

PR #718, opened 2018-10-03
chore: add codecov

evilebottnawi:
Also update same deps


evilebottnawi:
Feel free to feedback

PR #721, opened 2018-10-03
fix: do not mix LF and CRLF

evilebottnawi:
No tests, because prettier for pragma in test change newline on OS based newline

Also refactor


evilebottnawi:
Feel free to feedback

PR #728, opened 2018-10-03
feat: better print assign and retif without true expr

evilebottnawi:


evilebottnawi:
Feel free to feedback

PR #732, opened 2018-10-03
fix: cast and bin parens

evilebottnawi:
Really **BIG BUG** bug https://github.com/glayzzle/php-parser/issues/172. Need fix in parser. Let's output parens around `cast` as is to avoid break code.


evilebottnawi:
Will be great if somebody fix this problem in parser, i don't have enough time on prettier plugin and parser :disappointed: 

Feel free to feedback.

PR #735, opened 2018-10-04
fix: indent return

evilebottnawi:
Try to prettify `laravel/framework` and found some bugs, let's fix it


tests/retif/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -259,7 +259,7 @@ $var =
         : $this->basePath . DIRECTORY_SEPARATOR . 'database') .
     ($path ? DIRECTORY_SEPARATOR . $path : $path);
 return ($this->databasePath ?:
-        $this->basePath . DIRECTORY_SEPARATOR . 'database') .
+    $this->basePath . DIRECTORY_SEPARATOR . 'database') .
```
evilebottnawi: Expected - remove extra indent. I think we should use other logic for print this, i will do this in next PR

evilebottnawi:
Feel free to feedback

PR #737, opened 2018-10-04
fix: consistent print static member chain

evilebottnawi:
problem in parser, issue created, also add `TODO`


evilebottnawi:
Feel free to feedback

PR #738, opened 2018-10-04
fix: don't add newlines inside `encapsed` with member chain call

evilebottnawi:
Strange, `removeLines` doesn't remove lines in some cases, investigate


evilebottnawi:
Close in favor https://github.com/prettier/plugin-php/pull/839

PR #740, opened 2018-10-05
fix: bug with cast and parens

evilebottnawi:
**Big bug**


evilebottnawi:
Feel free to feedback

PR #744, opened 2018-10-08
feat: better print empty body for `try`, `catch` and `finally`

evilebottnawi:


src/printer.js
```diff
@@ -1268,19 +1269,14 @@ function printFunction(path, options, print) {
   }
 
   const printedDeclaration = concat(declaration);
-  const hasEmptyBody =
-    node.body &&
-    node.body.children &&
-    node.body.children.length === 0 &&
-    !node.body.comments;
```
evilebottnawi: Also refactoring

evilebottnawi:
Feel free to feedback

PR #746, opened 2018-10-08
feat: better print empty `class`, `trait` and `interface`

evilebottnawi:


evilebottnawi:
Feel free to feedback

PR #753, opened 2018-10-09
fix: trailing comment in call

evilebottnawi:
Before we get error about we don't print comment


evilebottnawi:
Feel free to feedback

PR #754, opened 2018-10-09
feat: better print halt comments

evilebottnawi:
Fix edge case


evilebottnawi:
Feel free to feedback

PR #755, opened 2018-10-09
fix: trailing comment in `function` and `method` node

evilebottnawi:
Before we got Error about what we don't print comment


evilebottnawi:
Feel free to feedback

PR #756, opened 2018-10-09
fix: indent for `print` node with end of line comment

evilebottnawi:
Looks we forget about this problem :smile:


evilebottnawi:
Feel free to feedback

PR #758, opened 2018-10-09
fix: indent for `yield` with end of line comment

evilebottnawi:
Looks better :smile:


evilebottnawi:
All test passed on travis, but labels above do not updating :confused: Feel free to feedback

PR #763, opened 2018-10-09
fix: indent for `echo` node with end of line comment

evilebottnawi:
Also refactor

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -1248,16 +1248,71 @@ echo // Comment
     $bar // Comment
 ; // Comment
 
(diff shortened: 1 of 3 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1248,17 +1248,72 @@ echo // Comment
...
+EOD
+    ,
+    $a;
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
 
 // Comment
 echo // Comment
-$foo, // Comment
+    $foo, // Comment
```
evilebottnawi: Looks we already have this bug

evilebottnawi:
Feel free to feedback

PR #767, opened 2018-10-09
feat: print comment for `return` on same line

### LINKED ISSUE #764: return with comments
**Prettier 1.14.3**

**PHP Plugin 0.8.0**

**Input:**

```php
return // Comment
    $a;
```

**Output:**

```php
return 
    // Comment
    $a;
```

**Expected behavior:**
/cc @vjeux 
Both looks good, but need advice. In PHP we don't print `(` and `)` for `return` because no one do it. Should we add `newline` if comment exists or not?

evilebottnawi:
fixes #764

Let's print `return` with comment(s) consistently, as we print other keywords with comment(s)

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -1189,8 +1189,7 @@ $constraint = new UniqueEntity(array(
     // no "em" option set
 ));
 
-return
-    // parens test
+return // parens test
     // some comment
     $test;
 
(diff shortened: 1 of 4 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2807,6 +2828,25 @@ function f()
     return /* a */;
 }
 
+return; // Comment
```
evilebottnawi: Also fix this bug (exists in previous implementation)

evilebottnawi:
Feel free to feedback

PR #769, opened 2018-10-09
[DNT] fix: semicolon for declare before `inline`

evilebottnawi:
Need discussion, no merge


evilebottnawi:
From PSR-12 (https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#3-declare-statements-namespace-and-import-statements), look on penultimate example

evilebottnawi:
Close in favor https://github.com/prettier/plugin-php/pull/877

PR #771, opened 2018-10-09
[DNT] fix: `namespace` keyword

evilebottnawi:
Very rare use case, some buggy in parser


evilebottnawi:
In theory it is not break code because `namespace` in new equals current top namespace, but it is change ast, feel free to feedback

evilebottnawi:
Close in favor https://github.com/prettier/plugin-php/pull/872

PR #775, opened 2018-10-10
fix: extra comment and overlaps for inline nodes

### LINKED ISSUE #509: One-line node with just a comment
Input:
```php
<?php // test ?>
<?php echo 1; ?>
```
Output:
```php
<?php // test
echo 1; ?>
```

Input:
```php
<?php // test ?>
<?php echo 1; ?>
<?php echo 2; ?>
```
Output
```
[error] stdin: Error: Comment location overlaps with node location
[error]     at decorateComment (...)
```

evilebottnawi:
fixes #509 

Some `inline` nodes doesn't look good, but not related to the PR, let's fix it in next PRs


evilebottnawi:
Feel free to feedback

PR #776, opened 2018-10-10
feat: better print if else comment

evilebottnawi:
As prettier in many cases

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -2102,6 +2102,23 @@ function f() {
         /* Allow Untracked */ $allowUntracked = true;
     }
 }
+
+if (true)
+    // Comment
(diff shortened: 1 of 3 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2255,28 +2267,28 @@ function f()
...
-        $untrackedChoice === 1 /* Add */ // Add
-    ) {
+    } elseif ($untrackedChoice === 1) {
+        /* Add */ // Add
         $shouldAmend = true;
-    } elseif (
-        $untrackedChoice === 2 /* Allow Untracked */ // Allow Untracked
-    ) {
+    } elseif ($untrackedChoice === 2) {
+        /* Allow Untracked */ // Allow Untracked
```
evilebottnawi: In original all comments above will be between `) {`

PR #777, opened 2018-10-10
fix: better print multiline `string`

evilebottnawi:
Now multileline string looks everywhere very good. Need same for encapsulated, do this in next PR


evilebottnawi:
Feel free to feedback

PR #778, opened 2018-10-10
feat: better print if comments

evilebottnawi:
Format comment as prettier for `if/else`, support comments between `if` and `elseif`, `elseif`  and `else`. Also i found this in `laravel` and `symfony`


evilebottnawi:
Feel free to feedback

PR #782, opened 2018-10-11
fix: parens for null coalescing operator

evilebottnawi:
Bug


evilebottnawi:
Appveyor broken on their side

PR #784, opened 2018-10-11
fix: extra newline after php tag with inline nodes

evilebottnawi:
I think we should refactor all around `inline` nodes, i will see what i can do


tests/halt/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -27,7 +27,6 @@ __HALT_COMPILER();
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 <?php
 
-
```
evilebottnawi: Fix this bug

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1623,6 +1623,278 @@ func(
...
+exports[`mixed-47.php - php-verify: mixed-47.php 1`] = `
+<?php
+namespace Foo;
+$var = 1;
+$var = 2;
+?>
+<div></div>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+namespace Foo;
```
evilebottnawi: Need also fix, we should respect user input. If `newline` exists print this, if not don't add, i will fix it in next PR

evilebottnawi:
Feel free to feedback

PR #788, opened 2018-10-11
fix: consistently output encapsed and multiline string

evilebottnawi:
Full support for `string` and `encapsed`:
```php
call("
string
string 
string");

call(
    "
string
string 
string"
);

call('
string
string 
string');

call(
    '
string
string 
string'
);
```


evilebottnawi:
Feel free to feedback

PR #791, opened 2018-10-11
test: variable variables

evilebottnawi:
Just test. Looks we catch bug in parser


evilebottnawi:
No, problem in snapshot :smile:

PR #792, opened 2018-10-12
fix: newline after namespace

evilebottnawi:
In some cases we lose newline after namespace, let's fix it.

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -3473,6 +3473,293 @@ namespace Monolog\\Handler;
 
 `;
 
+exports[`namespace-10.php - php-verify: namespace-10.php 1`] = `
+<?php
+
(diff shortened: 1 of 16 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -3473,6 +3473,293 @@ namespace Monolog\\Handler;
...
+ */
+class FlowdockFormatter
+{
+}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+<?php
+
+namespace Monolog\\Formatter;
+
+// Comment
```
evilebottnawi: Comments in not right position, we need fix it, let's fix it in other PRs

evilebottnawi:
Feel free to feedback

PR #794, opened 2018-10-12
fix: respect simple and complex string syntax

evilebottnawi:
Right now we change simple syntax to complex when we use `propertylookup` or `offsetlookup`, i think it is not good, we should respect original user input, also a lot of projects like `laravel`, `symfony`, `drupal`, `WordPress` use simple syntax for simple expressions and concatenation for really complex expressions. Also we prettier not codemod tool and i think we should avoid this behavior even it is safe.


evilebottnawi:
Feel free to feedback

PR #797, opened 2018-10-12
[DNM] feat: improved array associative break heuristic

evilebottnawi:
Port https://github.com/prettier/prettier/pull/5205

Need review and discussoin

### DIFF
--- a/tests/array/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/array/__snapshots__/jsfmt.spec.js.snap
@@ -690,12 +690,8 @@ $multiline = [
         'not' => 'collapse'
     ]
 ];
-$multiline = [
-    'don\\'t' => 'collapse'
-];
-$multiline = array(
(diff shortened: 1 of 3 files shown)


tests/array/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -711,9 +707,7 @@ $arr = array($var);
 $arr = [$var];
 $arr = [$var];
 $arr = [$foo => $bar];
-$arr = [
-    $foo => $bar
-];
+$arr = [$foo => $bar];
```
czosel: Looks much better to me :+1:

evilebottnawi: @czosel :+1: need rebase, i will do this in nexr future

evilebottnawi:
Close in favor https://github.com/prettier/plugin-php/pull/916

PR #805, opened 2018-10-15
fix: inline `staticlookup` node with `static` keyword

evilebottnawi:
Also fix bug due parser problem


evilebottnawi:
Feel free to feedback

PR #807, opened 2018-10-15
fix: bin with `nowdoc` or `heredoc`

evilebottnawi:
big bug with `nowdoc` and `heredoc` inside `bin`


evilebottnawi:
I think we should refactor logic around `heredoc` and `nowdoc` in future

PR #809, opened 2018-10-17
refactor: `nowdoc` and `heredoc` nodes

evilebottnawi:
Just refactor.


tests/nowdoc/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -356,6 +389,40 @@ EOF
     .
     $var;
 
+$str = <<<'EOD'
+
+EOD;
```
evilebottnawi: Bug in parser https://github.com/glayzzle/php-parser/issues/191, but not critical

evilebottnawi:
Feel free to feedback

PR #812, opened 2018-10-17
chore(deps): update

evilebottnawi:
Update parser

### DIFF
--- a/tests/parens/call.php
+++ b/tests/parens/call.php
@@ -1,7 +1,8 @@
 <?php
 
 call();
-(call());
+// TODO: uncomment after resolve https://github.com/glayzzle/php-parser/issues/194#issuecomment-430657284
+// (call());
 
 $var = call();
 $var = (call());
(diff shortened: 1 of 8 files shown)


tests/preserve_line/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -527,9 +527,10 @@ $foo->bar->baz
     ->doOneMoreThing($config)
     ->bar->run(function () {
         $console->log('Bar');
-    }); // Run this
+    });
 
 // Hello world
+// Run this
```
evilebottnawi: Expected, because we have invalid comment place before

tests/parens/call.php
```diff
@@ -1,7 +1,8 @@
 <?php
 
 call();
-(call());
+// TODO: incomment after resolve https://github.com/glayzzle/php-parser/issues/194#issuecomment-430657284
+// (call());
```
evilebottnawi: Bug, need fix in `parser` :disappointed:

evilebottnawi:
What the hell with `dev` prettier

PR #815, opened 2018-10-19
refactor: anonymous class

evilebottnawi:
Also add more tests. 

Why i change output for anonymous class? Because it is looks really ugly in real use case.

Example.
Now:
```php
public static function createLogger($arg) {
    $logger = new class(
        $arg['type'],
        $arg['severity'],
        $arg['filter'], 
        $arg['bail'] 
    ) extends BaseLogger {};

   $logger->init();
   
   return $logger;
}
```

Before:
```php
public static function createLogger($arg) {
    $logger = new class(
        $arg['type'],
        $arg['severity'],
        $arg['filter'], 
        $arg['bail'] 
    ) extends BaseLogger {
    };

   $logger->init();
   
   return $logger;
}
```

Before example have ugly `{` and `}`, there is no reason to add newline after `{`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -33,41 +33,7 @@ const languages = [
   }
 ];
 
-// for anonymous classes, we need to keep track of the parent "new" node.
-// this is because in a case like this:
-//   $test = new class($arg1, $arg2) extends TestClass {};
-// we actually want to pretend that the class node starts after the "new"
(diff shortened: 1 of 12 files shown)


evilebottnawi:
We can change this behavior in future - after feedback.
Feel free to feedback

PR #816, opened 2018-10-19
fix: nowdoc and heredoc with static node

evilebottnawi:
Just bug

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -50,7 +50,8 @@ const {
   hasEmptyBody,
   hasNewline,
   isNextLineEmptyAfterNamespace,
-  shouldPrintHardlineBeforeTrailingComma
+  shouldPrintHardlineBeforeTrailingComma,
+  isDocNode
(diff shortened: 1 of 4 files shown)


src/printer.js
```diff
@@ -768,12 +765,8 @@ function printBinaryExpression(
         getNodeKindIncludingLogical(node);
 
     parts.push(
-      node.left.kind === "nowdoc" ||
-        (node.left.kind === "encapsed" && node.left.type === "heredoc") ||
-        (node.left.kind === "bin" &&
-          (node.left.right.kind === "nowdoc" ||
-            (node.left.right.kind === "encapsed" &&
-              node.left.right.type === "heredoc")))
+      isDocNode(node.left) ||
```
evilebottnawi: Also create util `isDocNode` to avoid duplicate logic

evilebottnawi:
Feel free to feedback

PR #817, opened 2018-10-19
fix: nowdoc and heredoc output in `case` node

evilebottnawi:
Also a lot of tests


evilebottnawi:
Feel free to feedback

PR #820, opened 2018-10-19
fix: nowdoc and heredoc binary newline

evilebottnawi:
Fix a lot of bugs with `nowdoc` and `heredoc`


evilebottnawi:
Feel free to feedback

PR #839, opened 2018-11-06
fix: don't add newlines in `encapsed` node

evilebottnawi:


evilebottnawi:
Feel free to feedback

PR #851, opened 2018-12-01
Document supported options

### LINKED ISSUE #414: Include Options for Readme + Good Job Contributors
Hi there.

First off, I'm super excited for this project to reach a stable milestone as so many hybrid web-apps (PHP + JS) can now use one formatter for everything. This plugin is getting a lot of recent attention in the community. Having clean integration with no more blocking issues ---- will be huge for prettier adoption in the PHP community. So thumbs up to all those making this project possible.

I was wondering if it's possible to include the available options on the Readme? I think this would be useful. Obviously, I'm aware of the issue with tons of options; however, there are still the core ones (e.g. trailingCommas, etc.) ... it would be nice to have those all documented for easy reference. A small section in Readme would suffice.

czosel:
Fixes #414

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,20 @@ If you installed globally, run
 prettier path/to/file.php --write
 ```
 
+## Configuration
+
+Prettier for PHP supports the following options:
+
+| Name            | Default  | Description                                                                                                                                                          |
+| --------------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


czosel:
#636 Should be merged before this.

czosel:
The `--end-of-line` option didn't seem to work for me. Might that be because we're depending on an older version of prettier?

evilebottnawi:
@czosel no need documentation `--end-of-line` option, prettier support testing CRLF, LF and etc right now, we need just refactor

PR #868, opened 2019-01-17
chore(deps): use stable version of prettier

evilebottnawi:


evilebottnawi:
jest have problem in deps, let's wait when it was fixed

evilebottnawi:
Ref: https://github.com/salesforce/tough-cookie/issues/140

PR #872, opened 2019-01-18
fix: namespace rn resolution

evilebottnawi:
We can break code in some places without this, so it is bug fix


evilebottnawi:
Feel free to feedback

czosel:
:+1:

PR #886, opened 2019-01-18
fix: avoid semicolon in short echo

### LINKED ISSUE #609: semicolons in short echo tags
Has there been any discussion about semicolons in short echo tags? 

Using dev builds, this:

```php
<?= $content ?>
```
Changes to this (added semicolon):
```php
<?= $content; ?>
```

Nearly every PHP short echo tag example I found omits semicolons. Because short echo tags are usually found in template contexts, the less-verbose the better. 

There don't appear to be any PSR recommendations about this, though the CakePHP project (the only big-project guidance I could find) [recommends omitting](https://book.cakephp.org/3.0/en/contributing/cakephp-coding-conventions.html#short-echo) short echo tag semicolons:

```php
// wrong = semicolon, no spaces
<td><?=$name;?></td>

// good = spaces, no semicolon
<td><?= $name ?></td>
```

If short echo tag semicolons are the way Prettier-PHP is going, I'd like to start moving my code in that direction. If short echo tag semicolons are under discussion or being considered as an option, I vote for omitting them.

evilebottnawi:
fixes #609

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -355,6 +355,9 @@ function lineShouldEndWithSemicolon(path) {
   if (!nodeHasStatement(parentNode)) {
     return false;
   }
+  if (node.kind === "echo" && node.shortForm) {
+    return false;
+  }
   const semiColonWhitelist = [
     "expressionstatement",
     "array",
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Feel free to feedback

czosel:
:+1: nice!

PR #899, opened 2019-01-18
Refactor tests utils

evilebottnawi:
Refactor tests and test runner (based on prettier code).

https://github.com/prettier/prettier/blob/master/tests_config/run_spec.js

Now it is better reading and looks.


src/clean.js
```diff
@@ -34,6 +34,14 @@ function clean(node, newObj) {
     delete newObj.isDoubleQuote;
   }
 
+  if (node.kind === "inline") {
+    if (node.value.includes("___PSEUDO_INLINE_PLACEHOLDER___")) {
+      return null;
+    }
+
+    newObj.value = newObj.value.replace(/\r\n?|\n/g, "");
+  }
```
czosel: Why do we need this?

czosel:
I like the new output :+1:

src/clean.js
evilebottnawi: @czosel workaround in `newlines` for `inline` nodes, we can fix it after implement better ast for inline

czosel:
@evilebottnawi can you rebase?

src/clean.js
czosel: Alright! Was just a bit confused because this seems unrelated to the test util refactor.

evilebottnawi:
@czosel done :+1:

PR #912, opened 2019-01-23
chore: deps update parser

evilebottnawi:
Includes:
- Change `static` ast printing
- Fixed `constantstatement` and `classconstant` printing with `value` as prettier format `const`/`let`/`var`
- Tests

Before we use same algorithm for `*constant` and `static` nodes. Now they are same in some places, but it is really difference ast, also `*constant` nodes ast will be changed in near future

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,14 +11,15 @@ build: off
 deploy: off
 
 cache:
-  - node_modules
-  - "%LOCALAPPDATA%/Yarn"
+  - node_modules -> appveyor.yml, package.json, yarn.lock
+  - "%LOCALAPPDATA%/Yarn -> appveyor.yml, package.json, yarn.lock"
 
 install:
   - ps: Install-Product node $env:nodejs_version
   - cmd: npm -v
   - cmd: node -v
   - cmd: npm install -g yarn
+  - cmd: yarn --version
   - cmd: yarn
 
 test_script:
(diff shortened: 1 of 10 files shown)


src/printer.js
```diff
@@ -2374,8 +2405,7 @@ function printNode(path, options, print) {
       return printAssignment(
         node.left,
         path.call(print, "left"),
-        // Assignments nested inside a static declaration doesn't have the operator set, so printing manually
-        concat([" ", node.operator ? node.operator : "="]),
+        concat([" ", node.operator]),
```
evilebottnawi: Not `static` have valid ast, no need this hack

evilebottnawi:
I hate `AppVeyor` :disappointed: cache invalidation doesn't work in many many many cases

evilebottnawi:
@czosel any ideas what is wrong with `AppVeyor`?

.appveyor.yml
```diff
@@ -11,14 +11,15 @@ build: off
 deploy: off
 
 cache:
-  - node_modules
-  - "%LOCALAPPDATA%/Yarn"
+  - node_modules -> appveyor.yml, package.json, yarn.lock
+  - "%LOCALAPPDATA%/Yarn -> appveyor.yml, package.json, yarn.lock"
```
evilebottnawi: Setup cache invalidation :+1: to avoid problem in future

czosel: nice!

PR #916, opened 2019-01-24
feat: improved array and list associative break heuristic

evilebottnawi:
Ref: https://github.com/prettier/plugin-php/pull/797

### DIFF
--- a/tests/parens/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/parens/__snapshots__/jsfmt.spec.js.snap
@@ -2407,7 +2407,9 @@ new Translator(
     new MessageFormatter(),
     'en',
     $someOtherVar,
-    array('foo' => 'bar')
+    array(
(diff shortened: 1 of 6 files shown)


tests/parens/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2407,7 +2407,9 @@ new Translator(
     new MessageFormatter(),
     'en',
     $someOtherVar,
-    array('foo' => 'bar')
+    array(
+        'foo' => 'bar'
+    )
```
evilebottnawi: Should be collapsed, but it is problem with inline printing, we fix it in future, i think in next release we solve problem with printing `inline` nodes and release stable version

czosel: Alright! :+1:

PR #918, opened 2019-01-24
feat: add `braceStyle` option

### LINKED ISSUE #107: Option for opening brace

evilebottnawi:
fixes #107


evilebottnawi:
@czosel i fix some problem in comments tomorrow and do release :+1:

PR #920, opened 2019-01-25
refactor: simplify code for `traituse` node

evilebottnawi:
Just refactor + tests, increase print speed, `traituse` can't contain inline nodes


evilebottnawi:
Feel free to feedback

czosel:
👍

PR #921, opened 2019-01-25
refactor: simplify class body printing

evilebottnawi:
Refactor + tests, `class`/`trait`/`interface` can't contain `inline`


evilebottnawi:
@czosel yes, let's do it

PR #932, opened 2019-01-29
fix: better print multiline function and method declaration

### LINKED ISSUE #929: [bug] prettier-php formats some methods in my class with brace and closing paren  both on separate lines
**Prettier 1.15.3**

**PHP Plugin 0.9.0**

```sh
# Options (if any):
--single-quote
```

**Input:**

Original file is https://github.com/silverorange/swat/blob/master/Swat/SwatString.php

```php
public static function ellipsizeMiddle(
		$string,
		$max_length,
		// @codingStandardsIgnoreLine
		$ellipses = ' … ', // the spaces are non-breaking spaces
		&$flag = null
	) {
		$string = trim($string);

...
```

**Output:**

```php
public static function ellipsizeMiddle(
		$string,
		$max_length,
		// @codingStandardsIgnoreLine
		$ellipses = ' … ', // the spaces are non-breaking spaces
		&$flag = null
	)
	{
		$string = trim($string);

...
```

**Expected behavior:**

I expected the opening brace to go on the same line as the paren. Other multiline method signatures in the same file are formatted with the paren and brace on the same line.

The file is quite large, but if I delete everything except the problem method the problem remains. I suspect the problem is related to the inline comment in the method signature. I can adjust my code to remove the inline comments but thought I should file a bug regardless.

evilebottnawi:
fixes #929

Looks we already have invalid printing in our code. Also we need implement `printFunctionParams` for function and method, now we use `printArgumentsList` and it is may output ugly paramaters in some very rare cases, also it is increase printing time. Let's do it in other PR

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1371,6 +1371,12 @@ function printFunction(path, options, print) {
     ]);
   }
 
+  const willBreakDeclaration = declaration.some(willBreak);
+
+  if (willBreakDeclaration) {
+    return concat([printedDeclaration, " ", printedBody]);
+  }
+
   return conditionalGroup([
     concat([
       printedDeclaration,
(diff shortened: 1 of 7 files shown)


evilebottnawi:
Feel free to feedback

PR #934, opened 2019-01-29
refactor: newlines/spaces before and after php tags

evilebottnawi:
Avoid regexp, fix a lot of bugs with `inline` nodes


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -5963,7 +5964,6 @@ printWidth: 80
  * @package WooCommerce/Templates
  * @version 2.0.0
  */
-
 ?>
 
 ================================================================================
```
evilebottnawi: Expected, it was bugs.

src/printer.js
```diff
@@ -1048,49 +1046,50 @@ function printLines(path, options, print, childrenAttribute = "children") {
...
+    const parts = [];
 
-    let afterOpenTag = " ";
-    let beforeCloseTag = " ";
+    const [firstNode] = node.children;
+    const hasStartTag = !firstNode || firstNode.kind !== "inline";
 
     if (hasStartTag) {
-      const between = originalText.trim().match(/^<\?(php|=)(\s+)?\S/);
+      const between = options.originalText.trim().match(/^<\?(php|=)(\s+)?\S/);
```
evilebottnawi: Will be fixed in next version of parser, we still need regexp here

evilebottnawi:
Feel free to feedback

PR #935, opened 2019-01-29
fix: respect newline before comment and in empty program node

evilebottnawi:


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -426,7 +426,6 @@ printWidth: 80
 =====================================output=====================================
 <?php
 
-
 // The first line of a comment
```
evilebottnawi: Extra newline was removed, because they doesn't exist in source file

PR #950, opened 2019-02-06
[WIP] Add build step for standalone version, resolves #842

### LINKED ISSUE #842: Including plugin-php in with standalone in Browser
I can't find any documentation on including the prettier parser with the standalone.js file for us in a browser project. 

Would be great if some docs would be added showing a clear way to do this if at all possible.

Loilo:
This pull request resolves #842, adding a standalone build to the Prettier PHP plugin.

### Build
The `standalone.js` in the root of the package is generated through running:

```bash
yarn run build-standalone
```

### Usage
The `standalone.js` is "standalone" in the sense that it is bundled into one file with no Node.js-specific dependencies left. It does however *not* include Prettier itself. Prettier's standalone version has to be included on its own first (order matters):

```html
<script src="https://unpkg.com/prettier/standalone.js"></script>
<script src="https://unpkg.com/@prettier/plugin-php/standalone.js"></script> <!-- does not exist yet -->
```

The plugin registers as a global variable named `prettier_pluginPhp`. Using Prettier with it look like this:

```js
prettier.format(YOUR_CODE, {
  plugins: [prettier_pluginPhp],
  parser: ["php"]
});
```

[Here's a demo.](https://jsbin.com/losogej/edit?html,output)

### Transformations
Things I instructed Rollup to do: 
* Mark the `prettier` module as `external`, so essentially all `require("prettier")` are replaced with the `prettier` global variable.
* Add a [shim for Node's `assert` module](https://github.com/Loilo/plugin-php/blob/standalone/build/shims/assert.js). I'm not sure why the `strictEqual` method should do nothing, but that's how it's done [in Prettier](https://github.com/prettier/prettier/blob/master/scripts/build/shims/assert.js).
* Add a shim for `Buffer.isBuffer()` (used by [the `php-parser` dependency](https://github.com/glayzzle/php-parser/blob/3ddcf93/src/index.js#L95)).
* Set `process.arch` to `x32` (relevant for maximum float size).

### ToDo
* [x] Write tests
* [x] Add documentation to readme

### Considerations
* [x] **Name of the global?**
  `prettier_pluginPhp` is not that catchy, suggestions are welcome.
  
  **✅ Resolved:** Like Prettier's built-in plugins, the plugin is registered on the global `prettierPlugins` object as `prettierPlugins.php`.
* [x] **Minify?**
  Prettier doesn't do it, therefore this pull request currently doesn't either.

  **✅ Resolved:** Prettier *does* actually apply minification, not to their `standalone.js`, but to each of their plugins. This PR matches this behavior and minifies the standalone plugin.
* [x] **Add the generated `standalone.js` to the repo?**
  Prettier does so.

  **❌ Rejected:** I was wrong, Prettier actually [does not](https://github.com/prettier/plugin-php/pull/950#issuecomment-460931496).
* [ ] **Is it reasonable to hardcode `process.arch` to `x32`?**
  Doing a quick search, I found no way to detect a machine's architecture with browser-side JS with a reasonable amount of confidence. `x32` should be a safe bet.
* [x] **Add documentation about how to use this plugin in bundlers?**
  The `standalone.js` is not really suited for usage with bundlers. To avoid confusion, an additional section about using bundlers in the readme (i.e. "rewrite all `prettier` calls to `prettier/standalone`") could be reasonable.

  **✅ Resolved:** No weird tips for bundler usage needed as this PR now also includes a pretty fool-proof way to integrate with bundlers.


czosel:
Wow, this looks great! :100: 

Concerning the questions:

- Name of the global: It doesn't sound too catchy indeed. Are there any conventions for this? @suchipi the name of the global for the lua plugin is generated from packd.now.sh, right?

- Minify: I agree - we can add this later if it becomes a requirement.

- Add built output to the repo? If I'm not mistaken, `standalone.js` in Prettier is not the actual bundle, but a module used as an entry point to generate the bundle. The actual bundle is generated in the `dist` folder which is not checked in to version control. I'd say we should do the same.

- process.arch: Good question. @ichiriac can you help us out with that? It seems that the relevant piece of code is this one: https://github.com/glayzzle/php-parser/blob/813caea94c5bb50370363cbc0843d721f218a3a7/src/lexer/numbers.js#L11

- I think the most important thing to document is how to include the plugin for usage in the browser. Usage with bundlers would be a nice addition though. This makes me wonder though: Could we somehow make it easier to bundle the plugin for in-browser usage? It would be nice if the rewrite step from `prettier` to `prettier/standalone` wouldn't be needed. In any case, this sounds like something we don't have to solve immediately.

Loilo:
> Are there any conventions for this?

None that I know of. I'd be very thankful for a better suggestion though. 😁

> If I'm not mistaken, `standalone.js` in Prettier is not the actual bundle, but a module used as an entry point to generate the bundle.

Absolutely. I looked that up but had already forgotten about it when I wrote down the pull request. 🙄 

However putting the standalone file into `dist` proabably wouldn't work here.

Users need to be able to access the standalone build directly via `@prettier/plugin-php/standalone`, just as in Prettier, for consistency. This works for Prettier because their repo is so huge that they basically compile their whole npm package into the `dist` folder — I think we currently don't have good reasons for doing that.

So I'd stay with compiling the standalone bundle to the project root but add it to `.gitignore`.

> Could we somehow make it easier to bundle the plugin for in-browser usage?

We could definitely make this a lot easier.

I could just add another Rollup config that differs slightly from the one in this PR:

* Instead of using the `prettier` global, it would rewrite `require("prettier")` to `require("prettier/standalone")`. This would also discharge the user of overriding the `prettier` module specifier globally, which gave me the shivers anyway.

  The user would still be in charge of importing `prettier/standalone` in their project (instead of just `prettier`), but that is a thing we should be able to clarify with documentation.
* Instead of generating a UWP bundle, it would create an ES module to export the plugin.

The resulting bundle would be listed in the `package.json`'s `browser` field, which bundlers prefer over the `main` field when targeting browsers.

Writing it down like that, the effort should actually be pretty moderate. Would you like me to attach it to this PR directly?

suchipi:
@czosel yeah- I assume it replaces characters that can't appear in a JS identifier with underscores (so `@prettier/plugin-lua` became _prettier_pluginLua`, replacing the `@` and `/`). I recommend `prettierPluginPhp`.

Loilo:
That's what I went with on first approach. Didn't like it that much, but I have to agree that mixed case is way worse.

czosel:
> So I'd stay with compiling the standalone bundle to the project root but add it to `.gitignore`.

:+1: 

> Writing it down like that, the effort should actually be pretty moderate. Would you like me to attach it to this PR directly?

Sounds great! :slightly_smiling_face: 
> I recommend `prettierPluginPhp`

:+1:

Loilo:
> > I recommend `prettierPluginPhp`
>
> 👍

Done.

> Write tests

Done as well.

I've approached this by renaming the `test` project to `test-node` and duplicating it as `test-standalone` with some necessary adjustments:

1. Exclude the `markdown` test — Prettier's standalone build does not include parsers, therefore the `markdown` parser is not available.
2. Add [a global `STANDALONE` variable](https://github.com/prettier/plugin-php/blob/36ab2362eff70d553ca63598acd117706e2c58a2/jest.config.js#L19-L21) through the Jest config to achieve [slight variations in the `run_spec.js` behavior](https://github.com/prettier/plugin-php/blob/36ab2362eff70d553ca63598acd117706e2c58a2/tests_config/run_spec.js#L13-L16).

I don't find either of these adjustments to be solved particularly elegant, so if a more experienced Jest user knows how to solve this in a nicer way, let me know!*

\* Solving the first adjustment differently actually is a *requirement*, since using a [neagive lookbehind assertion](https://github.com/tc39/proposal-regexp-lookbehind#negative-assertions) to exclude the `markdown` test [breaks tests in Node.js v6](https://travis-ci.org/prettier/plugin-php/jobs/489441639).

> Sounds great! 🙂

Alright, going to do that when the tests are fixed. 👍

evilebottnawi:
Awesome job! Something wrong in CI test  :confused:

evilebottnawi:
Any ideas how we can test this? Maybe run `standalone.js` on all tests and compare with snapshots too?

Loilo:
@evilebottnawi See my comment above — I've added tests already, also described how it works.

Failing CI is due to Node.js v6 not supporting RegEx backreferences (which I've used to exclude the `markdown` test from the standalone build). I haven't found an alternative way to exclude a test so far, would be nice if someone found a solution. 👍

Loilo:
Side note: I've actually discovered a better way for naming.

Prettier's very own parsers to be used with the standalone build (e.g. the [`parser-markdown.js`](https://unpkg.com/prettier@1.16.4/parser-markdown.js)) attach themselves to a global `prettierPlugins` object (or create that object if it doesn't exist).

This would be a pretty decent way for us as well, IMO. We'd go by `prettierPlugins.php`.

Loilo:
Okay, so I've fixed the tests by no longer excluding the markdown tests at all. I've rather included the markdown parser in the standalone test so this is now covered as well.

I had to update the snapshots of the markdown tests though since there was inconsistent behaviour between the builds: If no `tabWidth` is configured...
* ...the standalone build indents PHP code with the default width of the plugin (4 spaces)
* ...the Node.js build indents PHP code with the default width of Prettier (2 spaces)

This shoud definitely be streamlined, not sure which the desired behavior would be. This is a design decision. // cc @czosel @evilebottnawi

For now, I've changed the markdown test's config to explicitely use 4 spaces in every environment, this should however be reverted as soon as the behavior of both builds is aligned.

evilebottnawi:
@Loilo strange maybe something wrong on prettier side?

Loilo:
It definitely seems to be *something* on Prettier's side (or in the official Markdown plugin). The Lua plugin shows the same behavior.

jest.config.js
```diff
@@ -11,11 +11,24 @@ module.exports = {
...
+    },
+    {
+      displayName: "test-standalone",
+      setupFiles: ["<rootDir>/tests_config/run_spec.js"],
+      testRegex: "jsfmt\\.spec\\.js$|__tests__/.*\\.js$",
+      snapshotSerializers: ["jest-snapshot-serializer-raw"],
+      testEnvironment: "jsdom",
+      globals: {
+        STANDALONE: true
+      }
```
evilebottnawi: Maybe we can add SKIP_STANDALONE_TEST env for better test experience?

Loilo: Sorry, I'm not sure how your question relates to the code snippet?

evilebottnawi: @Loilo i mean `SKIP_STANDALONE_TEST=true npm run test/yarn test` (and no test for bundled)

evilebottnawi:
For me looks great, we need clarify indent problem and we can writing docs about usage in browser (we can start writing this now :+1: )

Loilo:
I don't even think the indentation problem should block this PR as it's not related, the PR tests just revealed the issue.

What's open yet?

1. Should `project.arch` always be `x32`?
2. Add extra build for bundlers
3. Write documentation

(1) is not in my skill range but I'd be fine with taking care of (2) and (3).

Loilo:
The problem with the indentation is that we cannot 100% prove that it's Prettier's responsibility — sure, PHP and Lua both have the issue, but that may still be coincidence.

Prettier itself has no internal plugin that defaults to anything other than 2 spaces, therefore we can't test the issue with the official standalone plugins.

evilebottnawi:
@Loilo 

> 1. Should `project.arch` always be `x32`?

Hm, where we use `project.arch` ?

> 2. Add extra build for bundlers
> 3. Write documentation

Let's do it :+1:

Loilo:
`project.arch` is used by the `php-parser` to check for 32 or 64 bit systems and therefore determines how many bits are available for numbers. @czosel already pinged the maintainer — if we don't get feedback on this, we should be safe with `x32` as far as I can tell.

evilebottnawi:
@Loilo i think we remove this from `parser`, if not we can resolve this in future :+1:

Loilo:
Sounds good to me. I'm on the build chain for bundlers right now and will provide docs after. Stay tuned. 😁

Loilo:
Okay, there we go – I have added support for bundlers. 🙂 

Now the `yarn build` chain consists of three Rollup builds — one that is generating the `standalone.js` for browsers and two which are creating bundler targets: `standalone.esm.js` (for webpack and Rollup) and `standalone.cjs.js` (for browserify).

---

I've also documented browser and bundler usage in the readme.

The section about browser usage contains a link to [a tiny demo](https://github.com/prettier/plugin-php/blob/cb6bef844b8d1dd564e3139b2ff621cd3a8d2a57/README.md#in-the-browser). This demo naturally does not work at the moment, as our `standalone.js` is not available on unpkg.com yet.

If you're interested in trying it out anyway, here's the same demo with a self-hosted version of this plugin: https://jsbin.com/mimokujowi/edit?html,output

---

If we're leaving the `process.arch` issue as it currently is (which should be a safe bet) then I guess this PR is ready to get merged. 👍

Any feedback on functionality or documentation is very welcome. 🙂

czosel:
Both docs and code look great to me! :100: 

I just have one more remark, because I misunderstood an earlier comment:

> Side note: I've actually discovered a better way for naming.
> Prettier's very own parsers to be used with the standalone build (e.g. the [`parser-markdown.js`](https://unpkg.com/prettier@1.16.4/parser-markdown.js)) attach themselves to a global `prettierPlugins` object (or create that object if it doesn't exist).
> This would be a pretty decent way for us as well, IMO. We'd go by `prettierPlugins.php`.

Couldn't we do _exactly_ the same as Prettier and attach ourselves to `prettierPlugins`? That would mean that usage of the PHP plugin in the Browser would look exactly the same as of Prettier core languages.

Otherwise, I'd prefer calling the global `prettierPluginPhp` (as originally suggested by @suchipi), because the dot in `prettierPlugins.php` looks a bit strange to me.

Loilo:
That's what I've done in 0bf6fa5. 👍

czosel:
@Loilo Ahh, that makes a lot of sense. Thanks for the enlightenment :wink:

README.md
```diff
@@ -117,6 +118,40 @@ If you installed globally, run
...
+```html
+<script src="https://unpkg.com/prettier/standalone.js"></script>
+<script src="https://unpkg.com/@prettier/plugin-php/standalone.js"></script>
+```
+
+Then use Prettier with PHP, just like this:
+
+```js
+prettier.format(YOUR_CODE, {
+  plugins: [prettierPlugins.php],
```
czosel: Wouldn't this be better then though?
```suggestion
  plugins: prettierPlugins,
```

Loilo: I'm not sure that this would work. `[prettierPlugins.php]` is an array while `prettierPlugins` is an object. Have you tried if it performs as expected?

package.json
```diff
@@ -40,8 +46,12 @@
   "scripts": {
     "pretest": "npm run build-standalone",
     "test": "jest",
+    "prepublishOnly": "npm run build",
     "prettier": "prettier --plugin=. --parser=php",
-    "build-standalone": "rollup -c build/rollup.config.js"
+    "build": "npm run build-standalone && npm run build-standalone-esm && npm run build-standalone-cjs",
+    "build-standalone": "rollup -c build/rollup-umd.config.js",
+    "build-standalone-esm": "rollup -c build/rollup-esm.config.js",
+    "build-standalone-cjs": "rollup -c build/rollup-cjs.config.js"
```
evilebottnawi: Hm, we really need this types of `standalone`? prettier have only one

/cc @suchipi

README.md
Loilo: I've just tested and it works like a charm. 👍

package.json
Loilo: That's because Prettier does not have a dependency on Prettier. 😁

The default `standalone.js` build relies on the presence of a global `prettier` which is not available in a bundler context. The other two builds `require`/`import` Prettier from `prettier/standalone`.

However, there may be a way to combine both approaches (use `require()` when in bundler context, use a global variable in browsers) but I'm not sure if Rollup supports generating anything like that.

If we'd find a way to generate such a unified UMD bundle, we could get rid of the `esm` and `cjs` builds.

Loilo: I've omitted the minification step and took a closer look at the generated UMD bundle. It seems that it actually *does* exactly that — it currently just `require()`s `prettier` instead of `prettier/standalone`, I'll take a look if we can fix that.

evilebottnawi: :+1:

Loilo: Wow, I've actually got it to work with a single build. Didn't expect that.

Thanks for staying sceptical!

I'll attach another commit and remove the unnecessary builds.

Loilo:
Thanks to the thought-provoking impulse from @evilebottnawi, we're now back to just a single UMD build that works universally in browsers and bundlers. ❤️

I've removed tracks of the other builds and updated the docs accordingly.

package.json
Loilo: Alright, we're back to one build. 👍

evilebottnawi:
@Loilo Great work! Anything left to finish?

Loilo:
I think we're good to go.

czosel:
I just published `v0.10.1` which contains the new and shiny bundle: https://unpkg.com/@prettier/plugin-php@0.10.1/standalone.js

Maybe someone is interested to try this with Prettier's playground? :wink:

Loilo:
I was just about to say, what would you think of adding the standalone build to the Prettier playground as `php (experimental)`?

suchipi:
@evilenottnawi could you create the issue please? I don't have all the context

evilebottnawi:
@suchipi I am also don't have, just review :smile: 

/cc @Loilo maybe you can create issue?

Loilo:
Sure.

Loilo:
I find this really hard to nail down. Gotta sleep now and won't have time tomorrow, but if anyone's curious, here's what I found out so far:

* I can *not* reproduce the issue with the standalone build.
* It still occurs in the plugin's `markdown` test when removing the explicit `tabWidth`.
* Now the most magical observation: It does *not* occur in the plugin's standalone tests if the according test config is unwrapped from the `projects` (in the `jest.config.js`) and exported directly:

  ```js
  module.exports = {
    displayName: "test-standalone",
    setupFiles: ["<rootDir>/tests_config/run_spec.js"],
    testRegex: "jsfmt\\.spec\\.js$|__tests__/.*\\.js$",
    snapshotSerializers: ["jest-snapshot-serializer-raw"],
    testEnvironment: "jsdom",
    globals: {
      STANDALONE: true
    }
  };
  ```

evilebottnawi:
@Loilo Maybe you can create issue in our repo so as not to forget about it when someone has time

Loilo:
Yep, going to do that. Wasn't able to narrow it down myself.

evilebottnawi:
Anyway, thanks for good job and helping :+1:

Loilo:
My pleasure. 🙂

PR #953, opened 2019-02-12
feat: add .phtml file extension

dhoulb:
This is a common file extension for PHP files that have more HTML content than PHP (Magento uses it extensively). But they're still just PHP scripts, there's nothing special about the syntax for these files.


czosel:
In principle I'm also good with this change, I just wonder if we should still wait with merging it a bit because `phtml` files are typically templates with lots of inline HTML - and our `inline` support is not as good as it should be yet.

On the other hand, we could also just describe the limitations around the current `inline` support in the README - people can still use a glob like `**/*.php` if they want to exclude their `.phtml` files.

What's your opinion on that @evilebottnawi ?

evilebottnawi:
@czosel i think it is normal merge, people find more problem with inline and it is increase fixing edge cases

evilebottnawi:
Let's merge this after resolve https://github.com/prettier/plugin-php/pull/954

dhoulb:
/cc @evilebottnawi Thanks should be merged correctly (I'd appreciate if you could check the changes quickly as it was a major conflict!)

@czosel If you want more convincing consider that VSCode doesn't specify a different language identifier for *.phtml files — it treats them as `php` — so it's reasonable for `@prettier/plugin-php` to do the same :)

czosel:
@dhoulb yeah, that's making a lot of sense :+1:

dhoulb:
@evilebottnawi

I quickly looked into adding some tests for .phtml files and I’m a bit unclear on the test strategy. Key thing we’d want to test is that a .phtml file is checked using the PHP parser when it’s run through Prettier right? i.e. checking that the .phtml extension is registered correctly?

But (and I might be wrong) I can’t see anything equivalent for .php or .php3 files etc. We’re currently just taking the file and _force_ running the PHP parser against it. I could add some .phtml files but it wouldn’t actually test what we want to test.

The right approach would be more of an integration test where we actually run Prettier (with plugin-php loaded) against a directory and check that it correctly parses files with different extensions.

Am I missing something obvious?

evilebottnawi:
@dhoulb just create `extensions` directory in tests and put couple example with `.phtml` extension

dhoulb:
@evilebottnawi No problem, tests added.

mdtrooper:
Related bug #1010 (or duplicated).

dhoulb:
@mdtrooper That bug isn't related to the file extension, it's just (I imagine) inline PHP tags being a WIP for prettier-php

PR #954, opened 2019-02-12
feat: use `linguist-languages`

evilebottnawi:


package.json
```diff
@@ -11,6 +11,7 @@
     "standalone.js"
   ],
   "dependencies": {
+    "linguist-languages": "^6.3.0",
```
evilebottnawi: /cc @ikatyang should prettier export `linguist-languages` or each plugins should have own version? Using own version increase bundle size

ikatyang: `linguist-languages` is a dev dependency (though we placed it in `dependencies` to get `npm install prettier/prettier` work) that is supposed to be bundled (only the data for target languages will be bundled) so it _will_ increase the bundle size on the Prettier side if we're going to export it.

evilebottnawi: @ikatyang so, should we export this on prettier side or not? :smile:

ikatyang: Probably not, bundling all the language data (400+) with only a few (< 10) of them used does not sound like a good idea.

evilebottnawi: @ikatyang so we can merge this as is? Anyway we can use tree shaking to avoid unnecessary data in bundle

ikatyang: Sure.

(Sorry I missed the pending review, I thought you just cc me. 😅)

PR #996, opened 2019-03-17
README: Fix editor integration ordering

czosel:


czosel:
Missed that in my review :sweat_smile:

PR #999, opened 2019-03-18
Update issue template to the latest plugin version

Loilo:
It should make sense to have the current version as the default. :)


Loilo:
@evilebottnawi I cannot view the details of the failed AppVeyor build (the page doesn't load), any idea why?

evilebottnawi:
Network problem

PR #1064, opened 2019-06-05
Fixed heredoc inside function calls inside other function calls

jodysimpson:
I found a formatting bug where heredoc inside of a function call inside of a second function call would format to a syntax error due to a comma being added after the heredoc delimiter.

Input:
```
$str = sprintf(sprintf(<<<EOD
foo
EOD
, true));
```

prettier output:
```
$str = sprintf(
    sprintf(
        <<<EOD
foo
EOD, 
        true
    )
);
```

correct output:
```
$str = sprintf(
    sprintf(
        <<<EOD
foo
EOD
        , 
        true
    )
);
```

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -258,7 +258,9 @@ function docShouldHaveTrailingNewline(path) {
   }
 
   if (
-    (parentParent && ["call", "new", "echo"].includes(parentParent.kind)) ||
+    (parentParent &&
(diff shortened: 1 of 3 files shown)


evilebottnawi:
Thanks!
/cc @czosel i think we should do patch release after merge

czosel:
Thanks @jodysimpson! :+1: 
Published as `v0.11.2`.

PR #1119, opened 2019-08-24
fix: update `php-parser`

evilebottnawi:
/cc @czosel

### DIFF
--- a/tests/parens/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/parens/__snapshots__/jsfmt.spec.js.snap
@@ -656,6 +656,27 @@ $var = ($a['apply_time'] > $b['apply_time']) ? -1 : +1;
 $var = $page > 0 || $page == 0 && $this->forcePageParam;
 $var = $page > 0 || ($page == 0 && $this->forcePageParam);
 
+@foo() || @foo();
+(@foo()) || (@foo());
+$var = @foo() || @foo();
+$var = (@foo() || @foo());
+
+@$i / 0;
+@($i) / 0;
+
+$a = (false && foo());
+$b = (true  || foo());
+$c = (false and foo());
+$d = (true  or  foo());
+
+$f = false or true;
+$h = true and false;
+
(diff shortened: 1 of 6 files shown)


tests/silent/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -24,7 +24,7 @@ $value = @     $cache[$key];
 <?php
 $variable = @foo();
 /* Intentional file error */
-$my_file = @file('non_existent_file') or
+($my_file = @file('non_existent_file')) or
```
evilebottnawi: I think it is not bug, it is feature, why:
```php
$c = false or foo();
```

Because `and` and `or` have less precedence, we assign `false` and call `foo()`, without parens, it is very misleading and some developers can confused, with parens all everything is clear

tests/parens/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2927,10 +2974,12 @@ try {
 }
 
 @list($width, $height) = getimagesize($file);
-@(list($width, $height) = getimagesize($file));
+// Todo https://github.com/glayzzle/php-parser/issues/356
+// @(list($width, $height) = getimagesize($file));
```
evilebottnawi: Bug in parser

tests/parens/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2927,10 +2974,12 @@ try {
...
 
 @$var += 10;
-@($var += 10);
+// Todo https://github.com/glayzzle/php-parser/issues/356
+// @($var += 10);
```
evilebottnawi: Bug in parser

evilebottnawi: Also parser has invalid ast before and now, so we can merge this commit and continue development, our plugin doesn't break output, only `AST_COMPARE` fails

czosel:
Looks good! 👍Just CI on node 8 seems to be failing...

PR #1126, opened 2019-08-26
fix: update `php-parser`

evilebottnawi:
Update parser:
- comment for constants in right places how

### DIFF
--- a/tests/comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/comments/__snapshots__/jsfmt.spec.js.snap
@@ -987,8 +987,7 @@ $var = new class {
 };
 
 $var = new class {
-    const CONSTANT =
-        // Comment 1
+    const CONSTANT = // Comment 1
         // Comment 2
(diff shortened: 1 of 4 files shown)


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -987,8 +987,7 @@ $var = new class {
 };
 
 $var = new class {
-    const CONSTANT =
-        // Comment 1
+    const CONSTANT = // Comment 1
```
evilebottnawi: As original input

evilebottnawi:
Travis has problem with invalid node_modules and using branch hashes :confused:

czosel:
Hmm, did you already try clearing the travis cache?

evilebottnawi:
@czosel yep, it is green only when caches were cleaned

evilebottnawi:
I will write in bug tracker travis in near future

PR #1127, opened 2019-08-26
Update rollup and related dependencies

czosel:


loilo:
New `standalone.js` for some reason does not work with the playground, currently investigating, please hold the line. 😁

loilo:
Nevermind, it's all fine. Nuxt just was confused by me trying to use a local copy of `@prettier/plugin-php`. Overriding the `standalone.js` in `node_modules` directly worked fine, so no problems on my side.

PR #1128, opened 2019-08-26
fix: update `php-parser`

evilebottnawi:


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -6234,10 +6234,10 @@ $var = new class {
...
 
 $var = new class {
-    // Comment 1
-    public $name =
-        // Comment 2
-        'string';
+    public // Comment 1
+        $name =
+            // Comment 2
+            'string';
```
evilebottnawi: Bug, i will open new issue, let's fix it in other PR, not critical

evilebottnawi: https://github.com/prettier/plugin-php/issues/1129

PR #1130, opened 2019-08-26
fix: update php-parser

evilebottnawi:


tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -7016,8 +7016,7 @@ use // Comment
 use some\\namespaces\\{
     /* Comment */ /* Comment */ /* Comment */ ClassA /* Comment */,
     /* Comment */ ClassB /* Comment */,
-    /* Comment */ ClassC /* Comment */
-    /* Comment */ as C
+    /* Comment */ ClassC as /* Comment */ /* Comment */ C
```
evilebottnawi: https://github.com/prettier/plugin-php/issues/1131

evilebottnawi: Anyway some comment right now in right position and no newline after `ClassC`

PR #1189, opened 2019-10-03
feat: update parser

### LINKED ISSUE #795: broken simple string syntax with offsetlookup
**Prettier 1.14.3**

**PHP Plugin 0.8.0**

**Input:**

```php
$var = "string ${juices['FOO']} string";
```

**Output:**

```php
$var = "string \${$juices['FOO']} string";
```

**Expected behavior:**

```php
$var = "string ${juices['FOO']} string";
```

Issue https://github.com/glayzzle/php-parser/issues/251

evilebottnawi:
fixes #795

/cc @czosel some ast feature for better encapsed strings

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -2344,13 +2344,28 @@ function printNode(path, options, print) {
       );
     case "variable": {
       const parent = path.getParentNode();
+      const parentParent = path.getParentNode(1);
+      const ampersand = parent.kind === "assign" ? "" : node.byref ? "&" : "";
+      const dollar =
+        (parent.kind === "encapsedpart" &&
+          parent.syntax === "simple" &&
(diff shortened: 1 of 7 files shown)


src/printer.js
```diff
@@ -2344,13 +2344,28 @@ function printNode(path, options, print) {
       );
     case "variable": {
       const parent = path.getParentNode();
+      const parentParent = path.getParentNode(1);
```
czosel: Nitpick: Would “grandparent” be a better name?

evilebottnawi: We already used `parentParent` and `parentParentParent` and etc in our code so i think better keep this name as is

czosel: Alright!

PR #1265, opened 2020-01-01
fix: namespace followed by inline node

### LINKED ISSUE #1052: Invalid code when namespace is followed by HTML comment
**@prettier/plugin-php v0.10.2**
[Playground link](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBYwIbeLJUBhdAFtrbfCExZdLDiYDGUAMowAnmocwbAgAWgAmEBZoAEdBeGgABRNlOz05BRBiEQgYJLYU+2l0qABfFigKeBh0FDR4LElQNmwYNIMyzMwwQVa8EAAeAH4wajAAAkw2ayhsMDZPMal0dABuMcGAPgAdTB3+gEIQkLGACVEAWQAZMfcbaZIxo+3MEBKgA)

**Input:**
```php
<?php namespace foo; ?>

<!-- HTML comment -->

```

**Output:**
```php
<?php namespace foo;



<!-- HTML comment -->

```

**Expected behavior:**

```php
<?php  namespace foo; ?>

<!-- HTML comment -->
```

Current output is invalid. 

**Notes:**
This seems to be specific to `namespace` declarations. I wasn't able to reproduce the problem with other declarations, assignments or function calls.

czosel:
Fixes #1052


czosel:
Thanks for the review @loilo! I added a fix.

czosel:
@loilo good to go?

loilo:
Seems to generate valid code now as far as I can tell. 👍 

I've found another somewhat related thing, but it was not introduced by this PR, so it should not block merging.

However, for completeness: The following code...

```php
<?php
namespace Foo;

bar();
?>

<!-- HTML comment -->
```

...is actually made uglier by Prettier. 😅 

```php
<?php
namespace Foo;

bar(); ?>

<!-- HTML comment -->
```

evilebottnawi:
:+1:

Issue #1053, opened 2019-05-15
Migrate on own parser


evilebottnawi:
/cc @czosel 

Current parser does not improve and does not fix bugs, I propose to go to your own parser, we can use https://github.com/nikic/PHP-Parser/tree/master/grammar as example. I know it is require time, but unfortunately it seems we have no other way.

loilo:
Any chance we could viably compile the original PHP parser to WebAssembly?

loilo:
PHP as a whole has already been [compiled to WASM](https://github.com/oraoto/pib), but I have no insight into that field and can't tell whether compiling parts (i.e. just the parser) would be feasible.

evilebottnawi:
@loilo i am afraid what original php parsers is not enough for prettier, a lot of information about nodes is missing because php no need this

karptonite:
Any possibility the plugin could use the https://github.com/nikic/PHP-Parser directly, or would this have the same issue with missing nodes? Obviously that means running php processes as part of prettier PHP (no necessarily simple), but it would have the benefit that it would be easier to stay up to date with future language changes.

evilebottnawi:
@karptonite same problem not enough node information :disappointed: I think I can implement parser using bison/yacc (https://github.com/nikic/PHP-Parser using modified version yacc, but it is not a problem) for two weeks, but i don't have time right now, if somebody want to help me will be great.

karptonite:
Perhaps the PHP version is extensible enough to allow it to be extended to get the additional node information? I'm looking at this issue as an example of getting information that doesn't normally get stored by the parser: https://github.com/nikic/PHP-Parser/issues/384
It also looks like there are some lexer options that might give you more information: https://github.com/nikic/PHP-Parser/blob/1.x/doc/component/Lexer.markdown#lexer-options
I don't know much about how these things work, though, so I'm sort of guessing here.

evilebottnawi:
To be honestly for us interesting only https://github.com/nikic/PHP-Parser/blob/master/grammar/php5.y and https://github.com/nikic/PHP-Parser/blob/master/grammar/php7.y and https://github.com/nikic/PHP-Parser/blob/master/grammar/tokens.y

What we should do:
- Install bison/yacc
- Create js version https://github.com/nikic/PHP-Parser/blob/master/grammar/parser.template
- Rewrite this on js https://github.com/nikic/PHP-Parser/blob/master/grammar/rebuildParsers.php
[...]

PR #1267, opened 2020-01-01
feat(docs): add hints on debugging and running tests

czosel:
cc @joemaller

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,17 +5,51 @@ To get up and running, install the dependencies and run the tests:
 ```bash
 yarn
 yarn test
-```
 
-Run `yarn prettier path/to/file.php` to format a single file.
+# format a single file
+yarn prettier path/to/file.php`
+
+# run a single test
+yarn test tests/{folder of your choice}
+```
 
 Here's what you need to know about the tests:
 
 * The tests use [Jest](https://facebook.github.io/jest/) snapshots.
 * You can make changes and run `jest -u` (or `yarn test -u`) to update the snapshots. Then run `git diff` to take a look at what changed. Always update the snapshots when opening a PR.
 * You can run `AST_COMPARE=1 yarn test` for a more robust test run. That formats each file, re-parses it, and compares the new AST with the original one and makes sure they are semantically equivalent.
-* If you would like to debug prettier locally, you can either debug it in node or the browser. The easiest way to debug it in the browser is to run the interactive `docs` REPL locally. The easiest way to debug it in node, is to create a local test file and run it in an editor like VS Code.
 
 Run `yarn lint --fix` to automatically format files.
 
 If you can, take look at [commands.md](https://github.com/prettier/prettier/blob/master/commands.md) and check out [Wadler's paper](http://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf) to understand how Prettier works.
+
+## Debugging
+
+### VScode
+
+Add the following configuration to `.vscode/launch.json`:
+
+```json
+{
+    "type": "node",
+    "request": "launch",
+    "name": "Prettify test.php",
+    "skipFiles": [
+        "<node_internals>/**"
+    ],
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -5,17 +5,51 @@ To get up and running, install the dependencies and run the tests:
 ```bash
 yarn
 yarn test
-```
 
-Run `yarn prettier path/to/file.php` to format a single file.
+# format a single file
+yarn prettier path/to/file.php`
+
+# run a single test
```
loilo: I'd either clarify or go into detail what running the `kitchen_sink` test actually *does*. I don't know it, and I cannot infer it from this comment. Quite the contrary — "run a single" test sounds to me like "this is how you run one specific test of your choice", which this command obviously does not do.

If this command does what I think it does, something like `# run a quick allrounder test against Prettier` may be more fitting.

CONTRIBUTING.md
```diff
@@ -5,17 +5,51 @@ To get up and running, install the dependencies and run the tests:
...
+            "request": "launch",
+            "name": "Prettify test.php",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "program": "${workspaceRoot}/node_modules/.bin/prettier",
+            "args": ["--plugin=.", "--parser=php", "foo.php"],
+            "cwd": "${workspaceRoot}",
+            "outputCapture": "std"
+        }
```
loilo: This snippet seems to be unintentionally indented by 8 spaces. 🙃

CONTRIBUTING.md
```diff
@@ -5,17 +5,51 @@ To get up and running, install the dependencies and run the tests:
...
+```
+
+Afterwards, paste some PHP code in `test.php`, add a breakpoint, and start the debugger.
```
loilo: We're all not native speakers, so I might be completely wrong... but from the gut, I'd say you'd paste *to* a file, not *in* a file.

```suggestion
Afterwards, paste some PHP code to `test.php`, add a breakpoint, and start the debugger.
```

CONTRIBUTING.md
czosel: The only thing that's special about "kitchen_sink" is that it's not specific to a topic, which makes it a good candidate if one just wants to run a very specific test. I removed the reference to kitchen_sink, because it might be something that only I am using that way :wink:

CONTRIBUTING.md
czosel: I copy-pasted it out of VScode - fixed.

CONTRIBUTING.md
czosel: [Deepl](https://www.deepl.com/translator#de/en/Code%20in%20Datei%20einf%C3%BCgen) suggests "into", which seems like a good compromise :wink:

CONTRIBUTING.md
loilo: I didn't dislike it though. Having a quick allrounder test for checking whether you broke something fundamental sounds valuable to me — but since I haven't done any coding on the plugin itself, I cannot tell.

When do you typically use it?

czosel: I usually replace the entire content of kitchen_sink with the code I'm currently playing around with. So the actual content of kitchen_sink.php doesn't even add much value. That's why I prefer to remove it.

CONTRIBUTING.md
loilo: I considered it too, but was not clever enough to just ask a translation tool. 🙃

CONTRIBUTING.md
loilo: Oh. Yeah that's probably a bit too convulted to put it into the docs. 😁

I think the "format a single file" command should be sufficient then.

czosel: Right :wink: 

Let's leave it in since it was asked for in https://github.com/prettier/plugin-php/issues/1053#issuecomment-570082839

czosel:
@loilo Good to go? :slightly_smiling_face:

loilo:
Good to go. 🚀

PR #1270, opened 2020-01-02
feat: update parser

### LINKED ISSUE #1009: Stripping of brackets leads to corrupt code
**Prettier 1.16.4**

**PHP Plugin 0.10.2**

**Input:**

```php
echo '<select name="foo"><option value="bar"' . (@$_GET['foo'] === 'bar' ? ' 
 selected="selected"' : '') . '>bar</option></select>';
```

**Output:**

```php
echo '<select name="foo"><option value="bar"' . @$_GET['foo'] === 'bar' ? ' 
 selected="selected"' : '' . '>bar</option></select>';
```

**Expected behavior:**
```php
echo '<select name="foo"><option value="bar"' . (@$_GET['foo'] === 'bar' ? ' 
 selected="selected"' : '') . '>bar</option></select>';
```

czosel:
This adds support for the spread operator for arrays (PHP 7.4) and fixes
an issue with silent / retif precedence.

Fixes #1009

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1502,7 +1502,7 @@ function printAssignmentRight(
     return indent(concat([hardline, ref, printedRight]));
   }
 
-  const pureRightNode = rightNode.kind === "cast" ? rightNode.what : rightNode;
+  const pureRightNode = rightNode.kind === "cast" ? rightNode.expr : rightNode;
 
   const canBreak =
     (pureRightNode.kind === "bin" &&
@@ -2505,16 +2505,21 @@ function printNode(path, options, print) {
         { shouldBreak }
       );
     }
-    case "entry":
-      return printAssignment(
-        node.key,
-        path.call(print, "key"),
-        " =>",
-        node.value,
-        path.call(print, "value"),
-        false,
-        options
-      );
+    case "entry": {
+      const ref = node.byRef ? "&" : "";
+      const unpack = node.unpack ? "..." : "";
+      return node.key
+        ? printAssignment(
+            node.key,
+            path.call(print, "key"),
+            " =>",
+            node.value,
+            path.call(print, "value"),
+            false,
+            options
+          )
+        : concat([ref, unpack, path.call(print, "value")]);
+    }
     case "yield": {
       const printedKeyAndValue = concat([
         node.key ? concat([path.call(print, "key"), " => "]) : "",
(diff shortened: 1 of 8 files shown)


czosel:
@loilo ready for review. Not sure why appveyor is still running...

czosel:
You’re probably still using an old (cached) version of the parser. Try running
yarn cache clean
rm -rf node_modules
yarn

In the latest version of the parser “what” has been renamed to “expr” on cast nodes.

loilo:
Right, that fixed it. And I thought that's one of the things a lockfile should prevent. 🙄

czosel:
Seems like that's an open issue in yarn: https://github.com/yarnpkg/yarn/issues/4722

Good to go? :wink:

loilo:
Gimme a second, just doing the actual review now. Was too tired yesterday. 😁 

Everything looks about right to me, but I have some questions to better understand the changes, gonna put them in review comments.

src/printer.js
```diff
@@ -2505,16 +2505,21 @@ function printNode(path, options, print) {
...
-        " =>",
-        node.value,
-        path.call(print, "value"),
-        false,
-        options
-      );
+    case "entry": {
+      const ref = node.byRef ? "&" : "";
+      const unpack = node.unpack ? "..." : "";
+      return node.key
```
loilo: Do I understand this correctly that this differentiation is needed because of PHP 7.4 array unpacking and no `key` is available when unpacking something?

czosel: With the new parser version, `entry` objects are created for every array entry, not just for associative arrays anymore (see old behavior [here](https://astexplorer.net/#/gist/6999f9ae8437215dbc768e225f9346a1/0a16da17f1798e93541e66447d6af5bff4471a9b)). Also, `unpack` and `byRef` were added to the `entry` node for a generic way to describe the context of a specific entry.

So with the new version, we have to account for `entry`s that don't have a `key` (non-associative arrays).

loilo: Alright, this explains my follow-up question as well (why we didn't previously differentiate beetween kinds of array entries). 😁 

What I discovered (but I need to create an issue over at the parser for that): It's possible to do `[...&$var]` which should cause a parser error instead.

loilo:
See also: glayzzle/php-parser#461

evilebottnawi:
Good job!

PR #1279, opened 2020-01-04
Skip standalone build & test in development

### LINKED ISSUE #1276: Skip standalone tests during development
It would be nice if we could skip everything related to standalone tests during development, for two reasons:
* they always seem to pass when the node-based tests are passing
* faster turnaround times when running a single test / few tests

Currently it's already possible to skip the standalone tests using the env setting `SKIP_STANDALONE_TESTS=true`, but I'd like to skip the creation of the rollup bundle altogether.

@loilo Do you agree this is a good idea? If yes, would it be hard to implement?

loilo:
This PR removes the build process and Jest run for [standalone tests](README.md#in-the-browser) from the regular `test` script.

The build step can still be invoked via `yarn build-standalone`. The `test` script can be run including standalone tests when providing a `RUN_STANDALONE_TESTS` environment variable set to `true`.

Previously, only the build step was part of the `prepublishOnly` script. This PR complements that with a full test run to ensure standlone tests are still used before releasing a new version.

This PR also adds two new dev dependencies:

- [`cross-env`](https://www.npmjs.com/package/cross-env) for setting the `RUN_STANDALONE_TESTS` environment variable in a CMD-compatible way (yay for Windows users).
- [`yarpm`](https://www.npmjs.com/package/yarpm): Previously, the `pretest` script did `npm run build-standalone`, even if you used `yarn test`. This is a weird little inconsistency, but since yarn is not a hard dependency for this project, it couldn't just be assumed that a contributor has yarn installed.

  `yarpm` is a drop-in replacement for `npm run`/`yarn run` which uses the package manager that was used for running the current script.

Fixes #1276.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -20,9 +20,8 @@ module.exports = {
         STANDALONE: false
       }
     },
-    ...(process.env.SKIP_STANDALONE_TESTS === "true"
-      ? []
-      : [
+    ...(/^true$/i.test(process.env.RUN_STANDALONE_TESTS)
+      ? [
           {
             displayName: "test-standalone",
             setupFiles: ["<rootDir>/tests_config/run_spec.js"],
@@ -33,7 +32,8 @@ module.exports = {
               STANDALONE: true
             }
           }
-        ]),
+        ]
+      : []),
     {
       runner: "jest-runner-eslint",
       displayName: "lint",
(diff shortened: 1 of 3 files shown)


loilo:
On my machine this brought a `yarn test` run from 32 seconds (8s rollup, 24s jest) down to 13. ⚡️ 👌

PR #1282, opened 2020-01-05
fix(README): replace travis and appveyor badges with GH actions

czosel:


czosel:
I don’t know either - doesn’t matter 😉

Issue #939, opened 2019-01-30
Force array/list shorthand notation


bfanger:
Force the use of shorthand array's, (available since php 5.4)

## Reasoning:
The old syntax looks like a method call, not like an array. 
The new syntax is shorter and improves readability as it looks like an array.

Other issues regarding array's use the shorthand notation in their example code, indicating a preference towards shorthand.

**Input:**

```php
$my_array = array(1, 2, 3);
```

**Desired output:**

```php
$my_array = [1, 2, 3];
```

evilebottnawi:
hm, we need `--engine` option for this improvement, because some people can use prettier for `php5.3`/`php5.2` and lower. Not sure it is part of prettier, maybe better use `linter` for this.

/cc @czosel

hawkrives:
Is there any reason not to use the shorthand on newer PHP versions, though? (I, for one, would really like this transformation.)

evilebottnawi:
@hawkrives no, we officially support `php5.0`, `php5.1` and etc so we can broke their code, it can be solved only added `engine` field, but it is discussed

czosel:
These kind of transformations are not Prettier's job, but instead done by a linter. PHP-CS-Fixer seems to support this use-case as well (`array_syntax`). If PHP-CS-Fixer also support automatically fixing rule violations, a setup like eslint-prettier for JS would be possible for us as well.

Billz95:
> These kind of transformations are not Prettier's job, but instead done by a linter. PHP-CS-Fixer seems to support this use-case as well (`array_syntax`). If PHP-CS-Fixer also supports automatically fixing rule violations, a setup like eslint-prettier for JS would be possible for us as well.

For anyone who's interested in integrating `prettier` into the workflow of `php-cs-fixer` as a fixer, I've wrote a customised fixer to execute `prettier` before all other fixers.
https://gist.github.com/Billz95/9d5fad3af728b88540fa831b73261733

evilebottnawi:
@Billz95 i think we can add this to docs/recipes

Billz95:
> @Billz95 i think we can add this to docs/recipes

Hmmm sorry would you mind elaborating a bit on `docs/recipes`? Couldn't find much info about it, thx

evilebottnawi:
@Billz95 we can create recipes directory and put your example, also add link on this in README, it is allow faster integrate prettier for other developers

evilebottnawi:
[...]

Issue #493, opened 2018-07-25
Behavior change: Add trailing commas to array items


gouch:
I propose a behavior change to array formatting: Prettier should add a trailing comma to the last item in a multi-line array.

Trailing commas are beneficial in that they make editing easier and make diffs smaller. But more importantly, the coding standards for several significant PHP projects (e.g. Drupal, Symfony and WordPress) **require/recommend trailing commas.**

[PSR-2](https://www.php-fig.org/psr/psr-2/) doesn’t address this topic, so I think changing this behavior to match common practice is consistent with Prettier’s goals.

Current behavior:
```php
$array = [
    'item-in-array-example-1',
    'item-in-array-example-2',
    'item-in-array-example-3'
];
```

Desired behavior:
```php
$array = [
    'item-in-array-example-1',
    'item-in-array-example-2',
    'item-in-array-example-3',
];
```

I hope this is easy to implement since the main Prettier project already has a config option for [trailing comma](https://prettier.io/docs/en/options.html#trailing-commas).

For reference, these project coding standards all recommend trailing commas:

* [Drupal Coding Standards: Arrays](https://www.drupal.org/docs/develop/standards/coding-standards#array)
* [Symfony Coding Standards: Structure](https://symfony.com/doc/current/contributing/code/standards.html#structure)
* [WordPress Coding Standards: Indentation](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation)

virgofx:
I think it does trailing commas now by default?

virgofx:
Just to clarify/confirm --- Prettier supports `trailing commas` and so does the PHP plugin ... which can be configured by using the `trailing commas` option? So the above use case should be sufficiently handled for the user ...

gouch:
@evilebottnawi I think trailing commas is a better default *for php.* Prettier also does 2 tab width by default, but this php plugin changes that to [4 spaces](https://github.com/prettier/plugin-php/blob/master/src/index.js#L328), because it's a better match for PHP (both PSR-2 and common practice).

evilebottnawi:
@gouch PSR-1/PSR-2/PSR-12 Do not make trailing comma required, unlike `4 spaces`

gouch:
@evilebottnawi I acknowledge that; PSR-2 is silent about trailing commas. But I think it's valuable to consider common practice and the coding standards of significant PHP projects. PHP is a different language after all. That's why this plugin and all its extra code exists.

[...]

PR #1287, opened 2020-01-08
fix: update parser to remove extra linebreak in nowdoc

czosel:


czosel:
Shall we start preparing a release? 🙃

loilo:
Shall we? May be worth to address #939 first as it's the last missing feature touched by the `phpVersion` option... (Probably? Not sure about #493.) Would be nice to release them all together.

czosel:
Right - i totally forgot about #939 because it's so easy to implement :sweat_smile: #493 is already implemented though.

czosel:
Done: #1289 :wink:

Issue #1274, opened 2020-01-03
Proper support for flexible heredoc/nowdoc syntax


czosel:
Currently, the new flexible heredoc/nowdoc syntax is "destroyed" during formatting:

```php
$php73FlexibleNowdoc = <<<'EOD'
      Example of string
      spanning multiple lines
      using nowdoc syntax.
      EOD;
```
is printed as
```php
$php73FlexibleNowdoc = <<<'EOD'
Example of string
spanning multiple lines
using nowdoc syntax.
turns into

EOD;
```
Also, there seems to be an extra linebreak before the end.

Concerning expected behavior, I'd propose to leave the indentation the same way we find it. Thoughts?

cc @evilebottnawi @loilo

loilo:
> First things first: I realized the indentation currently only resets for nowdoc blocks, not for heredoc. We should definitely align their behaviors.
>
> For the following, I'm going to assume that this has already been fixed and flexible nowdoc and heredoc actually work the same. I'm just gonna use "heredoc" below, please read that as "heredoc/nowdoc".

This is a tricky one.

I'd say, it would be The Prettier Way™ (and also the prettier way 🙃) to figure out one way to print heredocs and to enforce that.

One process which sounds reasonable to me: If the starting tag is on its own line, indent the content with that. Otherwise indent it one tab width further.

---

<details>

<summary><strong>Examples of the described rule</strong></summary>

### Example 1

The heredoc-opening tag goes on its own line.

**Input:**

```php
some_function(<<<'EOT'
Hello World!
EOT
);
```

**Current output:**

```php
some_function(
    <<<'EOT'
Hello World!
EOT
);
```

**Envisioned output:**

```php
some_function(
    <<<'EOT'
    Hello World!
    EOT
);
```

### Example 2

The heredoc-opening tag goes on the same line as the preceding code.

**Input:**

```php
$string = <<<'EOT'
Hello World!
EOT;
```

**Current output:**

```php
$string = <<<'EOT'
Hello World!
EOT;
```

**Envisioned output:**

```php
$string = <<<'EOT'
    Hello World!
    EOT;
```

</details>

---

However, this is not doable consistently without an `--engine` flag because we obviously cannot assume PHP 7.4 for everyone.

[...]

PR #1289, opened 2020-01-10
feat: deterministic printing of array shortform

### LINKED ISSUE #939: Force array/list shorthand notation

czosel:
Fixes #939

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -59,12 +59,16 @@ const {
   normalizeMagicMethodName
 } = require("./util");
 
+function isMinVersion(actualVersion, requiredVersion) {
+  return parseFloat(actualVersion) >= parseFloat(requiredVersion);
+}
+
 function shouldPrintComma(options, requiredVersion) {
   if (!options.trailingCommaPHP) {
     return false;
   }
 
-  return parseFloat(options.phpVersion) >= parseFloat(requiredVersion);
+  return isMinVersion(options.phpVersion, requiredVersion);
 }
 
 function shouldPrintHardlineForOpenBrace(options) {
@@ -2422,8 +2426,11 @@ function printNode(path, options, print) {
     }
     case "list":
     case "array": {
-      const open = node.shortForm ? "[" : concat([node.kind, "("]);
-      const close = node.shortForm ? "]" : ")";
+      const useShortForm =
+        (node.kind === "array" && isMinVersion(options.phpVersion, "5.4")) ||
+        (node.kind === "list" && node.shortForm);
+      const open = useShortForm ? "[" : concat([node.kind, "("]);
+      const close = useShortForm ? "]" : ")";
 
       if (node.items.length === 0) {
         if (!hasDanglingComments(node)) {
--- a/src/clean.js
+++ b/src/clean.js
@@ -34,6 +34,11 @@ function clean(node, newObj) {
     delete newObj.isDoubleQuote;
   }
 
+  if (node.kind === "array") {
+    // TODO if options are available in this method, assign instead of delete
+    delete newObj.shortForm;
+  }
+
   if (node.kind === "inline") {
     if (node.value.includes("___PSEUDO_INLINE_PLACEHOLDER___")) {
       return null;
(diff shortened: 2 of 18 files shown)


src/printer.js
```diff
@@ -59,12 +59,16 @@ const {
   normalizeMagicMethodName
 } = require("./util");
 
+function isSupported(actualVersion, requiredVersion) {
```
loilo: Being a bit nitpicky again: I think this function tells nothing about anything being supported, it just compares version strings. I'd rename it to something like `isMinVersion`.

That's debatable though.

src/printer.js
```diff
@@ -2422,8 +2426,9 @@ function printNode(path, options, print) {
     }
     case "list":
     case "array": {
-      const open = node.shortForm ? "[" : concat([node.kind, "("]);
-      const close = node.shortForm ? "]" : ")";
+      const useShortForm = isSupported(options.phpVersion, "5.4");
```
loilo: This will not work. Only arrays can use short form in 5.4, short form in `list()` assignments only work in 7.1.

czosel: Oh right, good catch! Also, AST cleaning still needs to be adapted - I'll try to finish it tonight.

loilo: I know, right? Skimming through the changed snapshots finally proved useful for once. 😅 

Trying to deepen my understanding here: What's AST cleaning?

evilebottnawi: ast after prettier should be `===` ast before prettier

loilo: Ah, right. Thanks for the explainer. 🙂

evilebottnawi: So if we change `array()` to `[]` we need change ast before prettier, it is allow to catch problems with outputting broken ast

czosel: Done. I reverted the behaviour for lists to how we used to do it - should we always use the short form for lists if phpVersion >= 7.1 though?

loilo: Not sure. I have not used it often (neither the short nor the long form) and therefore still find the long form more intuitive since I've known it for way longer. But that's just me, and I think it would be worth considering the short form in the spirit of an optinionated formatter.

Maybe make a Twitter poll? 😄

czosel: I'm also not using `list` much - let's leave it as is for now, I'll post the question in the issue about array short form, and maybe on twitter as well :wink:

src/printer.js
czosel: Agree & fixed.

czosel:
Let me know if this still needs anything :rocket:

loilo:
Also ready for a release if you ask me. 👌

loilo:
Wait a second with the release though please.

czosel:
Yeah, still need to prepare the docs anyway :slightly_smiling_face:

loilo:
Currently evaluating the current `master` build in the playground and not sure whether I'm having issues. Let me evaluate. 🙂

czosel:
I posted the question about shorthand list syntax over at https://github.com/prettier/plugin-php/issues/939 by the way. Another thing to consider: If we decide to implement this, the default for the `phpVersion` option should probably become `7.0`.

loilo:
True. 😬

loilo:
Okay, found the reason for my issues. They're based on a misconception on my side: We added a `phpVersion`, but did we actually take care of #1274? I mean, we fixed the extra blank line and we *discussed* our options in the issue comments, but we did not implement anything, right? 🤪

czosel:
I'm leaning towards short syntax for `list` because of it seems elegant and is symmetric to `array`. I'd say we set the default for `phpVersion` to `7.0`.

czosel:
> I mean, we fixed the extra blank line and we discussed our options in the issue comments, but we did not implement anything, right? 🤪

You're absolutely right :see_no_evil: I'll take a look :wink:

Issue #1280, opened 2020-01-05
RFC: New option `phpVersion`


czosel:
From the discussion in #1274 the idea to introduce a new option `phpVersion` emerged. Since this affects a few other ongoing discussions we'd like to share the plan and ask for feedback! If you agree to the proposal and have nothing to add, feel free to post a :+1: reaction!

The basic issue we'd like to solve is that in some cases we need to know the PHP version in use to achieve optimal formatting. For example, starting with PHP 7.3, heredoc/nowdoc can be formatted in a more readable (indented) way, which doesn't work in previous versions.

The following table summarizes the areas where we'd currently benefit from knowing the PHP Version:

| Feature | Available since | Benefit |
| --- | --- | --- |
| #1274 Flexible heredoc/nowdoc syntax | 7.3 | Prettier way to print heredoc/nowdoc |
| #939 Array shorthand notation | 5.4 | Print `array(...)` as shorthand `[...]` |
| #493 Trailing commas in arrays and lists | 5.0 (?) | Print trailing commas by default |
| Trailing commas in "use" | 7.2 | (same) |
| Trailing commas in function calls | 7.3 | (same) |

We propose to do the following changes:

- **Introduce new option `phpVersion`**
  As the name says, this options allows you to specify the PHP version you're using. Any valid PHP version numer can be specified as value, but they fall in different categories:
  - `7.1`, `7.0`, `5.6`, `5.5`, `5.4`: Since our formatting doesn't change between these versions, all of them are equivalent. Since most users are using one of these versions, **this category would be the default**.
  - `7.2`: Print trailing commas in "use"
  - `7.3`, `7.4`: Print trailing commas in function calls
  - `5.3`, `5.2`, `5.1`, `5.0` and older: Don't force array shorthand notation
  - older versions: Show "not supported" message

- **trailing commas**
  - remove all option values of except `none` and `all`, make it a Boolean (`none` => `false`, `all` => `true`)
  - change the default to `true` (which checks the `phpVersion` setting to determine which trailing commas can be printed) (see https://github.com/prettier/plugin-php/issues/493#issuecomment-570798898)

- **array shorthand**
  Print `array(...)` as shorthand `[...]` except when `phpVersion < 5.4`

- **flexible heredoc/nowdoc**
  Print heredoc/nowdoc in "classic" (non-flexible) style except when `phpVersion >= 7.3`

If you'd like to read more about the reasoning behind this, alternatives and trade-offs please see #1274. Thanks!

[...]

PR #1290, opened 2020-01-10
fix(docs): update README for 0.14.0

czosel:
Let's merge this when we're ready to release - otherwise it might confuse users.


README.md
```diff
@@ -158,15 +158,16 @@ prettier.format(YOUR_CODE, {
 
 ## Configuration
 
-Prettier for PHP supports the following options:
+Prettier for PHP supports the following options. We recommend that all users set the `phpVersion` option.
```
evilebottnawi: I think it is unnecessary, prettier is 0CJS, i think we should try to load `composer.json` and resolve version automatically, also default value  should be `7.2` because it is latest LTS

czosel: Hmm, we've had quite a long discussion about this and decided differently. Please see #1280 and #1274.

loilo: Also, looking up the project's `composer.json` is insufficient. It may easily be possible that a project has dependencies installed which require a higher PHP version. If we'd want to do it, we'd have to run `composer why php`, parse the output columns and analyse version constraints to pick the lowest (or rather, highest) common denominator.

czosel: :+1: 
Also, running the prettier plugin with a lower version setting than what is actually used is not really an issue. One is just missing out on some minor optimizations, but it still "just works".

evilebottnawi: I read them, and a little disagree, in real world we have:
- old projects like `wordpress` (plugins/themes/etc) they still support `5.4`
- new projects uses minimum supported LTS

So using `7.0` is not good idea, it will not suit anyone and will force both groups to setup the option, I think we should free from setuping at least one of the groups

> i think we should try to load composer.json and resolve version automatically

That we can do it in future

czosel: We went for the `7.0` default because our formatting actually doesn't change all the way down to `5.4`:

> 7.1, 7.0, 5.6, 5.5, 5.4: Since our formatting doesn't change between these versions, all of them are equivalent. Since most users are using one of these versions, this category would be the default.

evilebottnawi: I am afraid for trailing comma, it can break code

czosel: Trailing commas in arrays and lists are supported since very old PHP version (5.0?) and more trailing commas are only printed when the `phpVersion` option is set to 7.2 or higher, so we should be safe, right?

evilebottnawi: Yes, but in future versions this could potentially be dangerous (like HEREDOC/NOWDOC), if no changed why do not use `5.4` as default value, to avoid misleading for new developers?

czosel: Yeah, i guess we could also use `5.4` as a default, in a sense of a "lowest common denominator". @loilo what do you think?

czosel: The only thing I don't like about the `5.4` default is that this version reached its end of life in september 2015 - which might be confusing to people when they expect our tool to be "modern" :wink:

evilebottnawi: We can explain why 5.4 is default :smile:

loilo: I dislike it, but I don't doubt that it might be reasonable. :)

That would make a feature which parses the Composer dependency tree even more useful though...

loilo: ...that may have not been clear enough: I support the decision for 5.4, I just hate that we have to do it. 😁

evilebottnawi: Sometimes you do what you hate :smile:

czosel: Alright, fixed :slightly_smiling_face: Turns out we had an mismatch between documented and actual default version before :see_no_evil:

PR #1295, opened 2020-01-12
fix: heredoc/nowdoc trailing newlines

### LINKED ISSUE #1292: Nowdoc: invalid trailing comma
```php
call(
    $var
        ? "foo"
        : <<<'EOD'
        string
        string
        string
        EOD
);
```
is formatted as
```php
call(
    $var
        ? "foo"
        : <<<'EOD'
        string
        string
        string
        EOD,
);
```
which is invalid.

czosel:
Fixes #1292

### DIFF
--- a/tests/nowdoc/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/nowdoc/__snapshots__/jsfmt.spec.js.snap
@@ -929,8 +929,7 @@ string
 string
 string
 EOD
-    ,
-    $var
+    , $var
 );
 
 $array = [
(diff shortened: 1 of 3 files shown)


czosel:
@evilebottnawi can you take a look at thise one as well? I have the release notes for 0.14.0 already prepared :wink:

czosel:
@evilebottnawi this is the last one :wink:

tests/nowdoc/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -929,8 +929,7 @@ string
 string
 string
 EOD
-    ,
-    $var
+    , $var
```
evilebottnawi: Why it is changed?

czosel: I think this is more of a side effect, the main idea of the PR was to fix #1292. It doesn't look too bad though, does it?

czosel: @evilebottnawi what do you say, good to go? This is now literally the last open issue for our release :wink:

evilebottnawi: I think will be better to put it on new line, because we always put each argument on next line

czosel: Think we can do that in a separate PR?

evilebottnawi: Yes, just open a new issue

czosel: done! I'll make the release now :tada:

PR #1303, opened 2020-01-19
feat: update parser (noop nodes)

czosel:
See https://github.com/glayzzle/php-parser/issues/272


czosel:
This turns out to be quite tricky, but I hope I'm slowly on the right track. We're printing `noop` comments by ourselves because prettier core adds unwanted newlines when the line seems empty (see [here](https://github.com/prettier/prettier/blob/a858a315e926c21bd1ebfcde70a08419c9472a8c/src/main/comments.js#L439))

Still TODO:

- [ ] fix newlines before `inline` nodes

@evilebottnawi can you take a quick look?

tests/switch/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -169,7 +169,7 @@ endswitch;
 =====================================output=====================================
 <?php
 switch (1) {
-    default:;
+    default:
```
czosel: Not sure about this one - both seems to be valid PHP code.

czosel:
@evilebottnawi think you could take a look at the `inline` issue? Would be awesome if we could confirm this approach before the stable version of the parser is released (https://github.com/glayzzle/php-parser/issues/465).

evilebottnawi:
@czosel sorry I don't have time right now, I think we should move forward and fix other bugs, inline is not easy to solve, so I think we can implement it under an option and mark it as an experiment so it is allow to us do breaking changes in patch/minor versions

czosel:
@evilebottnawi I didn’t mean better inline support in general - I totally agree that’s out of scope for now. I meant the parser update that I’m trying to do in this PR: some tests are still failing related to inline support. Would be awesome if you could take a look and comment on if the general idea is correct, and maybe you even have an idea how to fix the tests :)

tests/switch/__snapshots__/jsfmt.spec.js.snap
evilebottnawi: It is expected, because here noop node and now we don't print them

tests/comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -490,14 +490,15 @@ printWidth: 80
 =====================================output=====================================
 <?php
 
+
+
```
evilebottnawi: Extra hard lines, it is not good

src/comments.js
```diff
@@ -870,15 +874,14 @@ function printComments(comments, options) {
...
-    parts.push(hardline);
     if (
       isNextLineEmpty(options.originalText, comment, options) &&
       comments.length > index + 1
     ) {
       parts.push(hardline);
     }
   });
-  return concat(parts);
+  return join(hardline, parts);
```
evilebottnawi: I think here problem with extra hard lines, why we need hardline here?

czosel: Comments are now attached to `noop` nodes, so trailing newlines are printed automatically - all we need is to make sure newline between the parts are still printed.

evilebottnawi: :+1:

czosel:
@evilebottnawi @loilo ready for review :tada:

tests/comments/__snapshots__/jsfmt.spec.js.snap
czosel: fixed

czosel:
Merged with #1345

PR #1362, opened 2020-03-30
Support static & return-typed arrow functions

adamaveray:
Not that commonly used but [static arrow functions are supported](https://wiki.php.net/rfc/arrow_functions_v2#this_binding_and_static_arrow_functions). Prettier was stripping the static declaration, so this PR preserves it if set. It also adds support for declared return types.


czosel:
Looks great! Thanks for contributing :100:

PR #1370, opened 2020-03-31
release notes for 0.14.0

czosel:
Here are the release notes I drafted for 0.14.0 - this PR is not intended for merging, just for getting feedback. The actual release notes are directly in the "Releases" tab in GitHub :wink:

### DIFF
--- a/release-0.14.0.md
+++ b/release-0.14.0.md
@@ -1,0 +1,27 @@
+## Breaking change
+
+This release introduces the new option `phpVersion`. Depending on the options you're using, the upgrade requires some changes to your configuration:
+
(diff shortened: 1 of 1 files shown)


czosel:
It's done! :tada: https://github.com/prettier/plugin-php/releases

Thanks @evilebottnawi and @loilo for the reviews.

loilo:
> `5.3`, `5.2`, `5.1`, `5.0` and older

I don't think we should mention four versions and then say "and older". 😁 Maybe "`5.3` and below" would be the better wording here.

czosel:
:+1: I'll change it directly in the release notes.

loilo:
Playground is now running on 0.14 as well. 👌

PR #1374, opened 2020-03-31
fix(docs): trailingCommaPHP is a Boolean

### LINKED ISSUE #1373: trailingCommaPHP not accepting string value
### Introduction

There seems to be an incongruity between the documentation and functionality of version 0.14.0.

### Producing the Error

Running prettier with `"trailingCommaPHP": "none"` yields the following error:

```
Invalid trailingCommaPHP value. Expected true or false, but received "none".
```

Changing the value from `"none"` to `false` fixed the issue.

### System
prettier-php: v0.14.0
prettier: v2.0.2
Installed locally and run using npm.
Configuration:

```javascript
"prettier": {
      "printWidth": 80,
      "tabWidth": 3,
      "useTabs": false,
      "semi": true,
      "singleQuote": false,
      "quoteProps": "as-needed",
      "trailingComma": "none",
      "bracketSpacing": true,
      "arrowParens": "always",
      "proseWrap": "preserve",
      "htmlWhitespaceSensitivity": "css",
      "endOfLine": "lf",
      "overrides": [
         {
            "files": "*.php",
            "options": {
               "phpVersion": "7.2",
               "singleQuote": true,
               "trailingCommaPHP": "none", // <-- This is the problem
               "braceStyle": "psr-2",
               "tabWidth": 4
            }
         }
      ]
   }
```

### Conclusion

The documentation says that "none" is a valid option for trailingCommaPHP, as does the release notes for v0.14.0. I must conclude that this is either a bug or incorrect documentation.

czosel:
Fixes #1373


loilo:
LGTM. Also needs to be updated in the 0.14 release notes.

README.md
```diff
@@ -169,7 +169,7 @@ Prettier for PHP supports the following options. We recommend that all users set
...
-| `trailingCommaPHP` | `"all"`   | If set to `"all"`, trailing commas will be added wherever possible. <br> If set to `"none"`, no trailing commas are printed.                                                                                                                                                                                                                                                          |
+| `trailingCommaPHP` | `true`   | If set to `true`, trailing commas will be added wherever possible. <br> If set to `false`, no trailing commas are printed.                                                                                                                                                                                                                                                          |
```
evilebottnawi: I think will be great to accept as boolean and as string, because prettier support is as string

czosel: Hmm, so you mean to keep it similar to Prettier for JS, right? If the options parser allows for mixed types, we could do this. Shall we do it in a separate PR though?

evilebottnawi: Yes and yes

PR #1416, opened 2020-05-06
Update tabWidth to 2

marcofugaro:
From the README

However, when clicking the link, https://prettier.io/docs/en/options.html#tab-width

This PR sets the default `tabWidth` to `2`


evilebottnawi:
We use `4` for php because it is recommendation from PSR

czosel:
With "Same as in Prettier" is meant that the option has the same meaning as in Prettier for JS - not that should also have the same default value :wink:

Probably we could try to be a bit clearer about this in the README.

marcofugaro:
Oh ok, my bad. Yes, it needs to be specified in the README instead of "Same in prettier"

evilebottnawi:
@marcofugaro Feel free to open an issue about it

Issue #1386, opened 2020-04-09
Inconsistent formatting for $this vs other variable for fluent interfaces


karptonite:
**Prettier 2.0.4**

**PHP Plugin 0.14.0**

```sh
# Options (if any):
(all default)
```

**Input:**

```php
$that->shouldReceive( 'signUp' )->with( anInstanceOf( 'Foo\Bar\Baz' ), $this->app['foo.bar.baz']->getEmail())->once();

$this->shouldReceive( 'signUp' )->with( anInstanceOf( 'Foo\Bar\Baz' ), $this->app['foo.bar.baz']->getEmail())->once();
```

**Output:**

```php
$that
    ->shouldReceive('signUp')
    ->with(anInstanceOf('Foo\Bar\Baz'), $this->app['foo.bar.baz']->getEmail())
    ->once();

$this->shouldReceive('signUp')
    ->with(anInstanceOf('Foo\Bar\Baz'), $this->app['foo.bar.baz']->getEmail())
    ->once();
```

**Expected behavior:**
```php
$that->shouldReceive('signUp')
    ->with(anInstanceOf('Foo\Bar\Baz'), $this->app['foo.bar.baz']->getEmail())
    ->once();

$this->shouldReceive('signUp')
    ->with(anInstanceOf('Foo\Bar\Baz'), $this->app['foo.bar.baz']->getEmail())
    ->once();
```

Namely, `$that` shouldn't be on a line by itself in the first example. It should be formatted the same way as with `$this`.
[...]

PR #1437, opened 2020-05-22
feat: update chain breaking heuristic

karptonite:
As suggested here https://github.com/prettier/plugin-php/issues/1386#issuecomment-627439042, I've committed as a pull request my starting point to implement the same heuristic as Prettier js for chain breaking. In particular, prettier PHP formats as follows:
```php
$db
    ->setQuery()
    ->setValue()
    ->execute();
```
Whereas prettier js has adopted a style of keeping the first call on the same line as the parent when the parent is shorter than the indent size, as follows:
```php
$db->setQuery()
    ->setValue()
    ->execute();
```
I briefly started trying to implement this, and got as far as putting some test data in (probably the most valuable part of this PR), but I don't know enough about prettier to figure out whether the parent is at the beginning of a line. Without that, the change would cause a lot of things to wrap that we don't want to wrap, breaking a lot of other tests (as you'll see if you run the tests on this).

It would be good, though, if the project were to get in line with how prettier handles this, i think.

### DIFF
--- a/tests/call/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/call/__snapshots__/jsfmt.spec.js.snap
@@ -602,23 +602,21 @@ string
 string
 string');
 
-$a
-    ->call(
-        $a,
-        '
+$a->call(
+    $a,
+    '
 string
 string
 string',
-        $c
-    )
-    ->call(
-        $a,
-        '
+    $c
+)->call(
+    $a,
+    '
 string
 string
 string',
-        $c
-    );
+    $c
+);
 
 call("string $var string");
 call("string $var string");
--- a/src/printer.js
+++ b/src/printer.js
@@ -404,9 +404,13 @@ function printMemberChain(path, options, print) {
       const firstNode = groups[0][0].node;
 
       return (
-        (firstNode.kind === "variable" && firstNode.name === "this") ||
+        (firstNode.kind === "variable" && (firstNode.name === "this" || isShort(firstNode.name))) ||
         isReferenceLikeNode(firstNode)
       );
+    }
+
+    function isShort(name) {
+      return name.length <= options.tabWidth;
     }
 
     const lastNode = getLast(groups[0]).node;
--- a/src/printer.js
+++ b/src/printer.js
@@ -395,6 +395,12 @@ function printMemberChain(path, options, print) {
   //    $foo->Data['key']("foo")
   //      ->method();
   //
+  // 3. expression statements with variable names shorter than the tab width
+  //
+  // Example:
+  // $foo->bar()
+  //     ->baz()
+  //     ->buzz()
 
   function shouldNotWrap(groups) {
     const hasComputed =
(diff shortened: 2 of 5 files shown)


czosel:
@karptonite I figured I'd take a shot - seems to me that all that was missing was the check if the parent is an expression statement. Can you take a look and see if the changes reflect what you're looking for? It should be identical to what prettier for JS is doing.

@evilebottnawi I'd appreciate your review as well :upside_down_face:

tests/call/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -602,23 +602,21 @@ string
...
 string',
-        $c
-    )
-    ->call(
-        $a,
-        '
+    $c
+)->call(
+    $a,
+    '
```
evilebottnawi, line 614: I think it is not good idea, ugly, prettier@2 improve this, we should copy/paste logic from prettier :smile:

czosel: Seems that prettier is doing the same: [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBDAdAIwBQEYA0ABAAYA6UAZhBOZqgE62oBexRATAJTph5HFUaUOvWKcQ+EBAAOMAJbQAzslAN6EAO4AFBgmUpUAGw2oAnssmZ6qMAGs4MAMrSbcqAHNkMegFc4kgAsYAFtDAHUAuXhFFzA4Rz0ouQA3KNNkcEULEDdFOHoYLWt3YNRkCiM8yQArRQAPACFrOwdHVGC4ABk3OHLK-xBausc3d0M4AEUfCHg+wyqQF3o8+gy6TDhDCUX6NxgwuQATGADkAA4ABklpdTyw62kMm7gV5N7JAEdp+CKZfTRFABaKBwOCHMHbehwL5yKFFVAlMpICrzAZ5YJyLy+NGjcZTGa9ZH9SQwVCYA7HU5Idgk6xyQyjADCEGCpQyLwArNsfHkACpk-QohbJPwASSg4NgjjAu1kAEEJY4YKZxnM8gBfdVAA)

karptonite:
@czosel I haven't pulled to see if the tests pass, but otherwise, it looks good. The main problem for me was that I didn't know how to check if the parent was an expression statement.

karptonite:
And it looks like all the tests pass. Thanks, @czosel for getting this working. It seems now to be consistent with prettier for JS. Hope to see it merged soon!

tests/call/__snapshots__/jsfmt.spec.js.snap
czosel: Friendly ping @evilebottnawi 🙃

czosel:
@loilo could you take a quick look? @evilebottnawi seems to be busy :wink:

czosel:
Just released this as part of `0.14.3`! :tada:

loilo:
Alright, `0.14.3` is online in the playground. 🥳

loilo:
That said, `$this` vs `$that` is [still inconsistent](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBZhqYAGpIlWfCACsAOj4gWMCG3iyVAYXQBbH2wAFAAkA-BdBQxAOVwBjKABlGABPNWswbAgAWgAmJxBoAEdBeGgA12V-PTkFEGIRCBgytgq2KoMAXxYoCngYdBQ0eCxJUDZsGDaRTtrMMEEJvBAAHgB+YzAAHUwtgBIaeGxMgD4pdHQACgBKY442CCubtgAvK4BuLd2aNhhj04vro63e4A24vS6vEDtIA) in a vanilla 4-spaces indentation setup. 🙈

karptonite:
Thanks for the merge!

@loilo I agree on this vs that, but I seem to be in the minority. The issue to address that is here: https://github.com/prettier/prettier/issues/8003

PR #1483, opened 2020-07-13
PrettierPHPFixer - Use output of exec directly instead of writing temporary file

grubersjoe:
Hey,

Instead of writing a temporary file, Prettier's output can be processed directly if the second parameter of `exec` is set. This simplifies the fixer and should make it a bit more efficient.

I've tested this with a Symfony codebase with around 950 files and everything seemed fine. On my machine with a SSD the change had little impact on performance, but if an HDD is used, it might.

What do you think?

### DIFF
--- a/docs/recipes/php-cs-fixer/PrettierPHPFixer.php
+++ b/docs/recipes/php-cs-fixer/PrettierPHPFixer.php
@@ -58,37 +58,13 @@ final class PrettierPHPFixer implements FixerInterface {
     }
 
     /**
-     * {@inheritdoc}
+     * @param SplFileInfo $file
+     * @param Tokens      $tokens
      */
-    private function applyFix(SplFileInfo $file, Tokens $tokens) {
-        $tmpFile = $this->getTmpFile($file);
-        exec("yarn exec -- prettier --write $tmpFile");
-
-        $content = file_get_contents($tmpFile);
-        $tokens->setCode($content);
-
-        (new Filesystem())->remove($tmpFile);
-    }
-
-    /**
-     * Create a Temp file with the same content as given file.
-     *
-     * @param SplFileInfo $file file to be copied
-     *
-     * @return string tmp file name
-     */
-    private function getTmpFile(SplFileInfo $file): string {
-        $fileSys = new Filesystem();
-        $tmpFolderPath = __DIR__.DIRECTORY_SEPARATOR.'tmp';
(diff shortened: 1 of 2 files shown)


czosel:
Looks great on a first glance :+1: I don't have experience with PHPfixer though, so i'll cc @Billz95 again :slightly_smiling_face:

docs/recipes/php-cs-fixer/PrettierPHPFixer.php
```diff
@@ -58,37 +58,13 @@ public function supports(SplFileInfo $file) {
...
-            '_',
-            $file->getRealPath()
-        );
-        $tmpFilePath = $tmpFolderPath.DIRECTORY_SEPARATOR.'__'.$tmpFileName;
-        $fileSys->copy($file->getRealPath(), $tmpFilePath, true);
-        return $tmpFilePath;
+    private function applyFix(SplFileInfo $file, Tokens $tokens): void
+    {
+        exec("yarn exec -- prettier $file", $prettierOutput);
+        $code = implode("\r\n", $prettierOutput);
```
Billz95, line 67: I'd suggest use `PHP_EOL` over `\r\n` for this case.

grubersjoe: First I wanted to use `PHP_EOL` too, but this does not work because it is defined as `"\n"`. However, it seems like PHP CS Fixer requires the carriage return to be present. Else the output will be empty.

Billz95: That's interesting, php-cs-fixer worked for me somehow. 
May I know if you're using macos or linux or windows? 
Setting CRLF as EOL for everyone can cause issue for some if they don't have git auto CRLF to LF conversion set correctly

grubersjoe: Ok, you're right, `PHP_EOL` does work as expected.  For some reason the output was empty when I tested first but it worked now :+1: .

czosel: @Billz95 Can I go ahead and merge? :upside_down_face:

Billz95: lgtm, go ahead, thanks :D

czosel:
Thanks @grubersjoe & @Billz95 :+1:

PR #1533, opened 2020-09-13
chore: update prettier to 2.1.1

czosel:


czosel:
Prettier 2.1.1 seems to cause a minor change in the way we're formatting inline nodes at the end of a file. What do you think @evilebottnawi - safe to merge or do we need to investigate?

evilebottnawi (commented):
Yes, we should solve it, it is regressions :disappointed:

cseufert:
I feel like it the opening `<?php` tag should be on its own line as well if we are going to leave the closing tag on its own line.

czosel:
closing in favor of #1774

Issue #1112, opened 2019-08-06
Invalid output, "<?php" turned into "<?="


warpr:
**@prettier/plugin-php v0.11.2**
[Playground link](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBYwIbeLJUBhdAFtrbAAoAJe-hCYsulhxMBjKAGUYAE81VzBsCABaACYQFmgAR0F4aHsTZTs9OQUQYhEIGDS2DLYsgwBfFigKeBh0FDR4LElQNmwYMpFK3MwwQQ68EAAeAH4wajAAHUwAegAqOem5gAIrMCCIeGVqGGXohgBGAE5lgDF0dCWZ6ZGAPmmh6gBmW9GAXmWAEgQYNWWRkbLADk-hsUGW0DA9RgQP+rxmz1uIHKQA)

**Input:**
```php
<?php
/**
* Copyright 2019 Foo
*/
?>
<h3><?= $title ?? 'Some report' ?></h3>
```

**Output:**
```php
<?=
/**
* Copyright 2019 Foo
*/
?>
<h3><?= $title ?? 'Some report' ?></h3>
```

EDIT: <?= is not the same as <?php, so this change breaks this code.

evilebottnawi:
I think it is bug in parser :di

zarv1k:
[...]

PR #1567, opened 2020-10-16
fix: formatter breaks php-view file contains phpdoc block and php short-tags

zarv1k:
Fixes: #1112
Reviewers: @evilebottnawi


zarv1k:
@evilebottnawi could you please review the PR?

tests/inline/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -2347,6 +2347,58 @@ try {
...
+<?php
+/**
+ * @var string $a
+ */
+?>
+<?php echo $a; ?>
+<?= $a ?>
+
+=====================================output=====================================
+<?php
```
evilebottnawi, line 2391: Can we add test(s) where we start it with `<?=`?

zarv1k: added `mixed-56.php`

evilebottnawi (approved):

czosel:
Looks good, thanks for contributing @zarv1k!

czosel:
Released as v0.15.1 :tada:

PR #1574, opened 2020-10-24
feat: flexible heredoc support

czosel:


evilebottnawi (commented):
Looks good, but let's add more tests for php less 7.3

czosel:
Actually all the changed tests are running for PHP < 7.3 as well. They didn't change, so no diff there :upside_down_face:

evilebottnawi:
Sounds good

czosel:
:+1: Thanks for the quick review!

Issue #1254, opened 2019-12-21
Single quotes not changed to double quotes


glen-84:
**@prettier/plugin-php v0.11.2**
[Playground link](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBYwIbeLJUBhdAFtrbAAoAJe-hCYsulhxMBjKAGUYAE81VzBsCABaACYQFmgAR0F4aHsTZTs9OQUQYhEIGDS2DLYsgwBfFigKeBh0FDR4LElQNmwYMpFK3MwwQQ68EAAeAH4wajAAHUxpgBI2AAIAXgWAcjZVgG45jmWFyZAOA82QcqA)

**Input:**
```php
<?php

$a = 'a';
$b = "b";
```

**Output:**
```php
<?php

$a = 'a';
$b = "b";
```

**Expected:**
```php
<?php

$a = "a";
$b = "b";
```

[JavaScript example](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEBDTBeTAcm0IG4AdKSKDTAI303JDudJABoQIAHGAS3TJQ2AE6iIAdwAKYhGmQhsANwj8AJpxajsYANZwYAZR67+UAObIYogK5wuACxgBbADYB1R-3hpTYOCN5H35lHwBPRTA0BS5zNDhRGGkdCxdsZAAzbDcErgArNAAPACEdfUMjbBc4ABlzOCycvJBCoqNzCzc4AEVbCHgm3IcQU1EE0UU6bDo4Ny0eUXMYDw0YR2QADgAGLkWIBI8dHkVFuAnlRq4AR374FN4FFGw0AFooODh1L61ROFv+H8Utg0hkkNlhlwEi5+NY7CM0J1un0Bo1wc0RjAZqt1OtkAAmLg2bD8NydADCEBc6UUUGgVxAtgSABUZk8IS1lPYAJJQb6wIxgJZ8ACCfKMMHC3SGCQAvrKgA)

Related: #633.

czosel:
[...]

Issue #1545, opened 2020-09-23
Chained method arguments split up unnecessarily


hackel:
**@prettier/plugin-php v0.14.3**
[Playground link](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AjAEwAMANCDAIYBG5VteAFjYBXbFAAq3XHgBmHADZi22YgHN5UAIrD0MKPhgRhUNmGpgAakhVZ8IAOwA6ASDaGO8eWoDC6ALZ+HAAKABJBBkYmIFwQHADGUADKMACeGnZg2BAAtAyuINAAjsLw0EGxqoH4copRxGIQMOUclRzVCmIAvmxQFPAw6Cho8FjSoBzYMO213SDEYMJTeCAAPAD8ZmAAOpg7ACTFSCk7AARnJ9kAfGIacTAAShwA7gAUAORx6MIkLwBUAJQnCYnAAGn2+MBBLBOFA4egA+jJBoEYC8wdA4b14XCoSctiAAKQATWyBL82SYdHxgOBINhehBb3+p3OVxkEH872Kuig8LEHAgcWoUGwTJZZyuT2F0He8XBJHhlDe0L2cq+JAAkhRmZhzhdLlKkFB3nEMXoKNiYMqTm9LgBea3eAVcLB4PCYdCvf5XSYCmDIGQAWSwNBe3uuwi4wZI1GwLz2fhDsf+Or1V1UHOEYAAQil3vSoEyANw7ECdIA)

**Input:**
```php
<?php

$query
    ->selectRaw('count(*) as `count`, date_format(`created_at`, "%Y-%m-01") as `date`')
    ->from('quote_searches')
    ->where('account_id', $accountId)
    ->where('created_at', '>=', Carbon::now()->startOfMonth()->subMonths($months))
    ->groupBy('date');

```

**Output:**
```php
<?php

$query
    ->selectRaw('count(*) as `count`, date_format(`created_at`, "%Y-%m-01") as `date`')
    ->from('quote_searches')
    ->where('account_id', $accountId)
    ->where(
        'created_at',
        '>=',
        Carbon::now()
            ->startOfMonth()
            ->subMonths($months),
    )
    ->groupBy('date');

```

[...]

PR #1582, opened 2020-11-02
fix: Single quotes not changed to double quotes (#1254)

utkarshgupta137:
I'm actually not sure if this is 100% correct. I've eye-balled most of the diff.
Basically if !node.isDoubleQuote & !options.singleQuote that is the node has single quote and single quote option is not enabled, then invert the regex.
I also had to explicitly exclude strings with dollar sign ($) when converting from single to double quote, but they're not required when going the other way around. I'm not sure how it is not required. Right now, I've modified it too to keep regexs in sync.

Tested using "AST_COMPARE=1 yarn test".


utkarshgupta137:
@czosel

src/util.js
```diff
@@ -590,12 +590,22 @@ function getNodeKindIncludingLogical(node) {
...
+      );
+    }
+  } else {
+    if (!options.singleQuote) {
+      const rawValue = node.raw.slice(node.raw[0] === "b" ? 2 : 1, -1);
+
+      return !rawValue.match(
+        /\\([$nrtfve]|[xX][0-9a-fA-F]{1,2}|[0-7]{1,3}|u{([0-9a-fA-F]+)})|\r?\n|'|"|\$/
+      );
+    }
```
czosel, line 608: A little shorter (but potentially also more confusing?) way to write this might be:
```js
if (node.isDoubleQuote === options.singleQuote) {
  const rawValue = node.raw.slice(node.raw[0] === "b" ? 2 : 1, -1);
  const isComplex = !rawValue.match(
    /\\([$nrtfve]|[xX][0-9a-fA-F]{1,2}|[0-7]{1,3}|u{([0-9a-fA-F]+)})|\r?\n|'|"|\$/
  );
  return node.isDoubleQuote ? isComplex : !isComplex;
}
```

czosel (commented):
This looks already really good to me! :+1:  I think this captures what was discussed in #1254: The conversion should be symmetric, in a sense that:

- double quotes are changed to single quotes if the string is "simple"
- single quotes are changed to double quotes if the string is "simple"

cc @evilebottnawi

src/util.js
utkarshgupta137: It makes me sad that I didn't automatically think of this.

src/util.js
```diff
@@ -580,24 +580,24 @@ function getNodeKindIncludingLogical(node) {
...
 function useDoubleQuote(node, options) {
-  if (node.isDoubleQuote && options.singleQuote) {
+  if (node.isDoubleQuote === options.singleQuote) {
+    // We have a double quote and the user passed singleQuote:true, or the other way around.
     const rawValue = node.raw.slice(node.raw[0] === "b" ? 2 : 1, -1);
-
-    return rawValue.match(
-      /\\([$nrtfve]|[xX][0-9a-fA-F]{1,2}|[0-7]{1,3}|u{([0-9a-fA-F]+)})|\r?\n|'|"/
+    const isComplex = rawValue.match(
+      /\\([$nrtfve]|[xX][0-9a-fA-F]{1,2}|[0-7]{1,3}|u{([0-9a-fA-F]+)})|\r?\n|'|"|\$/
```
evilebottnawi, line 597: I am very afraid of such things, but it is better than we have, so let's merge

evilebottnawi (approved):

src/util.js
czosel: I didn't even notice this diff when I first looked at the PR :see_no_evil: 
The reason we didn't need to check for `$` when converting double to single quotes is that the parser splits them into `string` and `encapsedpart` nodes, such that the parts with the `dollar` are not even seen by the code that handles `string` ([example](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBZhqYAGpIlWfCACsAOj4gWMCG3iyVAYXQBbH2wAFAAkA-BdBQxAOVwBjKABlGABPNWswbAgAWgAmJxBoAEdBeGgA12V-PTkFEGIRCBgytgq2KoMAXxYoCngYdBQ0eCxJUDZsGDCICM7azDBBCbwQAB4AfmMwAB1Mbc2QKXR0AAIAEg42CD2AbhB2oA)).

The only unwanted side-effect this might introduce is that we wouldn't convert a string like `$ 5.00` even though it is actually "simple" (didn't test this yet, cc @utkarshgupta137)

utkarshgupta137: Yeah, that's why I mentioned it in the PR itself. I verified it doesn't affect behavior using AST_COMPARE=1.
For some reason, it doesn't work the other way around. So I had to add this regex too.

utkarshgupta137: We can either expand the syntax to \$\w or we can have the parser split single quotes too.

czosel: It seems that strings containing the `$` sign like `"5 $"` are currently not part of our test suite, so the tests don't show that we actually handle them differently with this change:
- `"5 $"` -> `'5 $'` with single-quote=true (before)
- `"5 $"` -> `"5 $"` with single-quote=true (now)

I think we should try to fix this by adding the official regex for PHP variable names, `^[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*$` (see [here](https://www.php.net/language.variables.basics)). Think that would work @utkarshgupta137?

cc @evilebottnawi

utkarshgupta137: This regex is only for variable names. This regex doesn't contain all syntaxes for string parsing.

czosel: That's right - I meant we have to add it to the regex that we already have. Something like
```
/\\([$nrtfve]|[xX][0-9a-fA-F]{1,2}|[0-7]{1,3}|u{([0-9a-fA-F]+)})|\r?\n|'|"|\$[a-zA-Z_\x80-\xff][a-zA-Z0-9_\x80-\xff]*/
```

utkarshgupta137: Yeah, but it doesn't cover complex string parsing.

czosel: How do you mean? Can you give an example?

utkarshgupta137: https://www.php.net/manual/en/language.types.string.php#language.types.string.parsing
See complex examples.

czosel: Ah I see, some example of what we'd miss if we just check for variables:
```php
'He drank some juice made of ${juice}s.'
'result: {${getName()}}'
```
Since all the complex expressions seem to contain the `$` character I'm fine with adding just that to our regex when converting from single to double quotes. When converting the other way around, I'd prefer to leave `$` out though, such that we can still convert "5 $" to '5 $' when `single-quote` is set to `true`.

utkarshgupta137: Yes but if by chance it says $bill or something like that, then converting from single to double quote will result in an error or change in syntax

czosel: You mean converting from double to single quotes? That won't happen because they're already split in the parser. And the other way around is caught by the `$` part that you added in the regex.

I'd implement it roughly like this:
```js
const isComplex = 
  (!node.isDoubleQuote && rawValue.includes("$")) || 
  rawValue.match(/\\([$nrtfve]|[xX][0-9a-fA-F]{1,2}|[0-7]{1,3}|u{([0-9a-fA-F]+)})|\r?\n|'|"/)
}
```
Or am I missing something?

utkarshgupta137: Sorry, I misunderstood what you said.
This would work. But I personally don't like the idea of things working one way but not the other.
I'll leave the final call to you.

czosel: Fair point, ist would make ist easier to think about this feature if the same regex is used. 
@evilebottnawi @loilo what’s your take on this?

utkarshgupta137: Any updates?

czosel: Sorry, I was a bit absorbed with other things in the last few days. What I didn't consider before is that the regression mentioned above,

> * `"5 $"` -> `'5 $'` with single-quote=true (before)
> * `"5 $"` -> `"5 $"` with single-quote=true (now)

is actually only affecting users who convert their non-pretty codebase for the first time using `single-quote=true`. Anyone who's using the plugin already and updates to the new release, won't see any regressions because `"5 $"` has already been converted to `'5 $'` and will stay that way.

So I'd stay let's stick with the current implementation :+1:

czosel (approved):
Thanks again for your contribution and patience @utkarshgupta137!

utkarshgupta137:
Thanks. Let me know if there are any other issues I could look into.

czosel:
Thanks for asking @utkarshgupta137! I think we currently have several different classes of open issues:

1. [enhancements](https://github.com/prettier/plugin-php/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Aenhancement+-label%3Adiscussion) that try to optimize things (usually plain PHP) that are already printed in an acceptable way.

2. Issues around [comments](https://github.com/prettier/plugin-php/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Acomments+-label%3Ainline).

3. Issues around [inline](https://github.com/prettier/plugin-php/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Ainline) formatting.

I'd say in terms of difficulty (1) and (2) are similar while (3) is very tough. In terms of impact, I'd say (3) > (1) > (2). So if you don't want to dig too deep just yet, I'd start with an issue in class (1). In particular, #1545 might be interesting to take a look at.

PR #1656, opened 2021-02-25
feat: add support for named arguments and nullsafepropertylookup (php8)

czosel:
Fixes #1599 

cc @Selion05

Ready in this PR:

- [x] nullsafepropertylookup
- [x] named arguments

Landed in parser:

- [x] match expression: https://github.com/glayzzle/php-parser/pull/686
- [x] union types: https://github.com/glayzzle/php-parser/pull/692, https://github.com/glayzzle/php-parser/pull/693

Open:

- [ ] promoted properties: https://github.com/glayzzle/php-parser/pull/697

PR #1744, opened 2021-06-02
feat: PHP8 Promoted Props / Match Statement / Throw Statement / Union Types

cseufert:
Adding Support for PHP8 Parser Features into Prettier

- [x] Promoted Constructor Properties
- [x] Match Statement
- [x] Throw Statement (not expression)
- [x] Union Types

See #1656 by @czosel for 
- [x] nullsafe property lookup
- [x] named arguments

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -2925,6 +2925,53 @@ function printNode(path, options, print) {
 
       return path.call(print, "name");
     }
+    case "match": {
+      if (!isMinVersion(options.phpVersion, "8.0")) {
+        throw new Error("PHP Version must be > 8.0 for match expression");
+      }
+
+      const arms = path.map((armPath, armIdx) => {
+        const conds =
+          armPath.getValue().conds === null
+            ? "default"
+            : concat(
+                armPath.map(
+                  (condPath, condIdx) =>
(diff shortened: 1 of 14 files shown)


cseufert:
@czosel Can you please give this code a look over, I have tried to follow similar functionality conventions where possible.

czosel:
@cseufert sorry about the delay, I'm currently swamped with work. I hope I'll get to it on the weekend. @alexander-akait maybe you could take a look as well?

czosel:
Looks great @cseufert :100:

src/printer.js
```diff
@@ -2916,6 +2936,54 @@ function printNode(path, options, print) {
 
       return path.call(print, "name");
     }
+    case "match": {
+      if (!isMinVersion(options.phpVersion, "8.0")) {
+        throw new Error("PHP Version must be > 8.0 for match expression");
```
czosel, line 2941: This line is still lacking coverage. I think we can drop the entire check though, because we're only using version checks when we have version-dependent output (e.g. short array syntax).

czosel:
I applied the changes over at #1753

PR #1761, opened 2021-06-17
feat(php8-attribute): PHP 8 Attribute output

cseufert:
Partial work in progress, currently planning on having all class, class method, class const, and class variables output the attributes one per line each with the attribute group `#[...]`, and for inline having them all in a single group, inline if there is room (inline functions, short functions and anonymous classes).

It is pretty rough right now, things left to figure out:
- [ ] Comment handling for  attribute with leading comments
- [x] Spaces vs line breaks
- [x] Function arguments
- [x] Printing Attribute Arguments
- [x] Anonymous inline classes operator order


cseufert:
@czosel can you take a look when you get a chance, still have an issue with comments appearing in the wrong order, but I suspect this may be a parser issue, still need to investigate that futher.

src/printer.js
```diff
@@ -1197,10 +1197,44 @@ function printClassPart(
   );
 }
 
+function printAttrs(path, options, print, inline = false) {
```
czosel, line 1200: Usually we're using objects instead of "non-speaking" boolean arguments in cases like this, such that the callsite looks like this:
```js
printAttrs(path, options, print, { inline: true });
```
See for example `getEncapsedQuotes`.

czosel (commented):
Looks good! Just a minor comment.

src/printer.js
cseufert: I have changed it, definately makes code easier to read, be nice if there is was a quick way to parse that param to `path.call(...)`

jaulz:
@cseufert not sure if you noticed it as well but with the recent version (0.17.3) attributes are completely removed from the files which is quite annoying since it can reset a lot of work. The only way to undo it is to roughly know the names of the files that you touched in your editor and use the undo operation (which obviously only works if you spotted it before you applied any other changes).

cseufert:
> @cseufert not sure if you noticed it as well but with the recent version (0.17.3) attributes are completely removed from the files which is quite annoying since it can reset a lot of work. The only way to undo it is to roughly know the names of the files that you touched in your editor and use the undo operation (which obviously only works if you spotted it before you applied any other changes).

This will be due to the attribute parser code being merged and included in 0.17.3. The main problem with this code currently is the handling of comments in and around attributes is hit and miss.

chrisrhymes:
> @cseufert not sure if you noticed it as well but with the recent version (0.17.3) attributes are completely removed from the files which is quite annoying since it can reset a lot of work. The only way to undo it is to roughly know the names of the files that you touched in your editor and use the undo operation (which obviously only works if you spotted it before you applied any other changes).

I have added a list of PHP files that use attributes to my projects `.prettierignore` file so the attributes don't get removed when prettier is run. 

I was running prettier with husky in a post-commit hook. The code worked fine locally, but broke when pushed up and being tested in the CI pipeline due to the attributes being removed.

cseufert:
We could merge this PR, and open bug reports as needed when we find issues with comments & attributes mingling together. I know there are issues with it, I just haven't had the time to discover what they are.

There is also another issues not mentioned, that currently all attributes are simplified:
eg:
```php
#[
  Attr1(1),
  Attr2
]
class Test {}
```
would become:
```php
#[Attr(1)]
#[Attr2()]
class Test {}
```

which is my personal preference (makes diffs easier to read / reason about, however it may not be everyone's preference, or if they are using attribute groups to have some semantic value. This also causes some tests to fail, as the AST does change.

czosel:
@cseufert sorry about the late reply! I totally agree, I'll go ahead and release this so that we can get feedback.

czosel:
Released in v0.17.4 :tada:

fernando-rivas-smtp:
Hello i have found a bug in this version when creating a class with multiple methods and you add an attribute just in the first one... all the following methods will have the same attribute... like the last one is stuck for all the rests...

alexander-akait:
@fernando-rivas-smtp please open a new issue with example of the problem, thanks

cseufert:
@fernando-rivas-smtp I have found the problem in the parser. I will create a PR for it.

PR #1763, opened 2021-06-19
fix(php8): non-capturing exception catches

### LINKED ISSUE #1760: PHP8: catch block without variable declaration
**Prettier 2.3.1**

**PHP Plugin 0.17.1**

**Input:**

```php
try {
    // Something
} catch (\Exception) {
    // Nothing
}
```

**Expected:**
Valid PHP 8 code; file should be formatted.

**Result:**

```
Parse Error : syntax error, unexpected ')' on line 13
  11 | try {
  12 |     // Something
> 13 | } catch (\Exception) {
     |                   ^
  14 |     // Nothing
  15 | }
```

czosel:
Fixes #1760

### DIFF
--- a/tests/errors/errors.php
+++ b/tests/errors/errors.php
@@ -46,3 +46,16 @@ try {
 } finally {
     // Comment
 }
+
+// PHP 8.0 non-capturing exception catch
+try {
+    // Something
+} catch (\Exception) {
+    // Nothing
+}
+
+try {
+    // Something
+} catch (MyException | OtherException) {
+    // Nothing
+}
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "linguist-languages": "^7.5.1",
     "mem": "^8.0.0",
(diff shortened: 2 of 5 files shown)


czosel:
@cseufert Could you look over this?

tests/errors/errors.php
```diff
@@ -46,3 +46,16 @@
 } finally {
     // Comment
 }
+
+// PHP 8.0 non-capturing exception catch
+try {
+    // Something
+} catch (\Exception) {
+    // Nothing
+}
```
cseufert, line 55: Is it worth making the example something that will actually be reformatted?

cseufert (commented):
Looks good

tests/errors/errors.php
czosel: I think in this case it's not needed; we're testing the formatting of try/catch already thoroughly, here we just make sure that non-capturing exceptions also work.

cseufert: Makes sense, I think we can merge this then.

cseufert (approved):

PR #1764, opened 2021-06-20
fix(#1785) workaround for extra end tag appearing when first line line has comment

### LINKED ISSUE #1758: PHP Code is broken after formatting.
**Prettier 2.3.1**

**PHP Plugin 0.17.1**

```sh
      arrowParens: avoid
      # jsx and tsx rules
      jsxBracketSameLine: false
      printWidth: 200
      tabWidth: 2
      useTabs: false
      singleQuote: true
      phpVersion: "7.2"
      braceStyle: "1tbs"
      endOfLine: "auto"
```

**Input:**

```php
<?php // P1T_SPECIALITE_03
					echo Input::selectList('contratPDRControl.contratPDR'.$id.'.specialites', 
							array ('typeSelect' => 'SpecialitePDR', 
							'multiple' => true,
							'horizontal' => true,
							/*'texte' => 'Spécialité',*/
							'select' => 'contratPDRControl.selectedSpecialite',
							'methode' => 'contratPDRControl.changeSelectedSpecialites',
							'reRender' => $reRenderAllTarif,
							'class' => 'mb-3'
							), $controleur->mapDonneesReference['PXX_PDR_SPECIALITE'], true, false, false); ?>

```

**Output:**

```php
<?php // P1T_SPECIALITE_03
// P1T_SPECIALITE_03
?>echo Input::selectList(
       'contratPDRControl.contratPDR' . $id . '.specialites',
       [
         'typeSelect' => 'SpecialitePDR',
         'multiple' => true,
         'horizontal' => true,
         /*'texte' => 'Spécialité',*/
         'select' => 'contratPDRControl.selectedSpecialite',
         'methode' => 'contratPDRControl.changeSelectedSpecialites',
         'reRender' => $reRenderAllTarif,
         'class' => 'mb-3',
       ],
       $controleur->mapDonneesReference['PXX_PDR_SPECIALITE'],
       true,
       false,
       false
     ); ?>
```

**Expected behavior:**
Here the php code is broken : the ``<?php`` is closed just before the ``echo``.

cseufert:
This is an attempt to fix #1758, I really need input from @czosel on this one.

It appears the output from the parser is good, comment line is only in output once, however somewhere it seems this formatted is somehow duplicating the comment node.

See: https://github.com/moddengine/php-parser/blob/a88bd19dadb2eaa76fe61ba3ccc84db7b70fae9d/test/snapshot/__snapshots__/lexer.test.js.snap#L57-L134

Somewhere along the way the inline tag is ending up with the leading comment from the echo below it. This patch does work and doesn't break any other tests, however, I think its got the potential to cause more problems in the future.


czosel:
Hmm, I agree that this might cause more problems in the future. I think it would help to understand where the second comment comes from. I’d start debugging in comments.js. I can try to take a closer look in the coming days.

cseufert:
I think the handling of start and end tags for PHP needs an overhaul, it's quite complex and hard to follow. Perhaps this code will do the job for now?

czosel:
Agreed, the handling needs improvement. I'd like to take a closer look at this before merging though - i'm still a bit hopeful that there is a simpler solution. I hope I'll find some time on the weekend.

czosel:
Sorry, I don't really find the time to dig deeper on this these days. @alexander-akait maybe you could have a look?

alexander-akait:
@czosel unfortunately there is no time at all now... In the coming months, I will have to work with a codebase on PHP and we need supports PHP 8/8.1/etc and fix inline nodes, so if everything goes according to my plan I will fix many cases/edge cases. Anyway based on test workaround looks fine, so maybe we can merge it without any problems

src/printer.js
```diff
@@ -1047,14 +1047,22 @@ function printLines(path, options, print, childrenAttribute = "children") {
           : isProgramLikeNode(node) && isFirstNode && node.kind !== "namespace"
           ? ""
           : concat([beforeCloseTagInlineNode, "?>"]);
+
+      const nextV = path.getNode(index + 1);
+      // process.stdout.write(`Inline: ${JSON.stringify(nextV)}\n`);
```
czosel, line 1052: Can you remove this line?

czosel:
Alright, then i'd say we merge it - @cseufert just a minor cleanup, then it's good to go.

src/printer.js
cseufert: Done.

PR #1774, opened 2021-06-26
chore: update prettier to 2.1.2

czosel:


src/printer.js
```diff
@@ -1102,7 +1102,7 @@ function printLines(path, options, print, childrenAttribute = "children") {
       const beforeCloseTag = lastNode
         ? concat([
             hasNewlineInRange(
-              options.originalText,
+              options.originalText.trimEnd(),
```
czosel, line 1105: see https://github.com/prettier/prettier/commit/441f2325b5dfd79eb15e37716741f471ab278fb9

tests/markdown/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -103,6 +103,7 @@ $foo = "bar";
 \`\`\`
 
 \`\`\`php
+
```
czosel, line 106: Upstream prettier does this as well (see [playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBiABAFTgZxgHSkIANSArHE0kAGhAgAcYBLaHZUAQwCduIB3AAo8E7FJwA2-TgE92dAEbdOYANZwYAZQYrmUAObIY3AK5w6cALYK4AE1t2AMpwMnO+uADEI3S5xgsBsggnCYwELQgABYwlhIA6lHM8Dg6YHCaosnMAG7JMsFgOPIgejhw3DCCyvp+yABmkuV0FAAeAELKahqanJZwjnpwDU3mIG2aevoScACKJhDwIxLNIDrc5dzBftyqtgJQkQzcejDxzLYwUcgAHAAMdMcQ5fHKDMHHuBU5w3QAjgt4NVGGIQjgALRQOB2OyRbhwAHMeHVdx1JCNFZjcqWZhGUxYqYzeaLYbo0Z0GCcBTnS7XJAAJgpymYEimAGEIJY0SBcABWSImcqYKliDGrHJmACSUAcsE0YBOTAAgjLNDAZDNluUAL7aoA)); the change was introduced here: https://github.com/prettier/prettier/commit/97b899b7bce31640b869b502c3f0c74abc0f37cb#diff-0a8624d0b561d808de85774597ac5c1de4ac029c6f47e5f68a31ffb457db60eeR36

src/printer.js
cseufert: makes sense based on prettier changes

cseufert (approved):

PR #1775, opened 2021-06-28
feat(CI) Update NodeJS versions to current stable releases

cseufert:
Node 13 is no longer supported by the NodeJS Foundation, we should focus on currently supported versions.


czosel (approved):

czosel:
We can also drop v10, right? Sorry, just realized it after merging 🙈

cseufert:
v10 is only just out of LTS, so it's probably fine for now.

PR #1857, opened 2021-10-30
fix: remove jest config from package.json

czosel:
Jest was warning about the config in package.json:
```
Multiple configurations found:
    * /home/christian/dev/plugin-php/jest.config.js
    * `jest` key in /home/christian/dev/plugin-php/package.json

  Implicit config resolution does not allow multiple configuration files.
  Either remove unused config files or select one explicitly with `--config`.
```


cseufert (approved):
Makes sense, no need to confuse things with multiple jest config's

PR #1890, opened 2021-12-29
fix(attributes): remove extra line break

### LINKED ISSUE #1884: PHP8 - Multi line attributes - Is the empty line afterwards by design?
Using PHP8 with mutli line attributes result in formatting like this by prettier:

```php
    #[
        ORM\Column(...),
        ORM\CustomIdGenerator(class: ...),
        ORM\GeneratedValue(strategy: 'CUSTOM'),
        ORM\Id
    ]

    private string $id;
```

however, for single line attributes it is like this:

```php
    #[ORM\Column(...)]
    private string $id;
```

**Prettier 2.5.1**

**PHP Plugin 0.17.6**

`.prettierrc`
```json
{
    "$schema": "http://json.schemastore.org/prettierrc",
    "arrowParens": "always",
    "braceStyle": "psr-2",
    "bracketSameLine": false,
    "bracketSpacing": true,
    "phpVersion": "8.0",
    "printWidth": 120,
    "singleQuote": true,
    "tabWidth": 4,
    "trailingComma": "es5",
    "trailingCommaPHP": true
}

```

It is weird that the multi-line syntax adds an empty new line afterwards, but the single line does not. 

Is this by design or a bug?

This issue might be related: https://github.com/prettier/plugin-php/issues/1885

czosel:
Fixes #1884

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1214,14 +1214,14 @@ function printAttrs(path, options, print, { inline = false } = {}) {
   }
   path.each((agPath) => {
     const attrGroup = ["#["];
-    if (allAttrs.length > 0) {
-      allAttrs.push(inline ? " " : hardline);
+    if (!inline && allAttrs.length > 0) {
+      allAttrs.push(hardline);
     }
-    attrGroup.push(ifBreak(softline));
+    attrGroup.push(softline);
     agPath.each((attrPath) => {
       const attrNode = attrPath.getValue();
       if (attrGroup.length > 2) {
-        attrGroup.push(",", ifBreak(softline, " "));
+        attrGroup.push(",", line);
       }
       const attrStmt = [attrNode.name];
       if (attrNode.args.length > 0) {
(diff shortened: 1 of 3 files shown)


tests/attributes/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -81,7 +81,7 @@ class D
      */
     #[I, J]
     function k(#[L] int $m): callable {
-        return #[N, O]  #[P] fn(#[Q] int $r) => $r * 2;
+        return #[N, O] #[P] fn(#[Q] int $r) => $r * 2;
```
czosel, line 84: drive-by fix: remove double space between multiple attr group if `inline==true`

Epskampie:
Thanks for fixing this! Would love to see it merged.

czosel:
@cseufert could you take a quick look? :slightly_smiling_face:

src/printer.js
```diff
@@ -1215,13 +1215,13 @@ function printAttrs(path, options, print, { inline = false } = {}) {
   path.each((agPath) => {
     const attrGroup = ["#["];
     if (allAttrs.length > 0) {
-      allAttrs.push(inline ? " " : hardline);
+      allAttrs.push(inline ? "" : hardline);
     }
```
cseufert, line 1219: Could this be simplified to 
```js
    if (!inline && allAttrs.length > 0) {
      allAttrs.push(hardline);
    }
```

cseufert (approved):
Look great, just could clean up conditions for empty non-inline attrgroup

src/printer.js
czosel: Absolutely - fixed.

PR #1922, opened 2022-02-08
Improve doc print

fisker:
I was going to help remove [the deprecated `concat` command](https://prettier.io/blog/2021/05/09/2.3.0.html#use-arrays-instead-of-concat-9733httpsgithubcomprettierprettierpull9733-by-fiskerhttpsgithubcomfisker-thorn0httpsgithubcomthorn0), but I found we are still running tests with `prettier@2.1.2`, are we going to continue to support old version?

Anyway, `eslint-plugin-prettier-doc` should  be useful whenever you decide to remove `concat()`.

Changes in this PR are problems caught by other rules in `eslint-plugin-prettier-doc`.


czosel:
Wow nice, I didn't know about `eslint-plugin-prettier-doc` yet - thanks @fisker! We're still on prettier 2.1 just because we didn't get to look at the formatting changes newer releases introduced yet (see https://github.com/prettier/plugin-php/runs/5057210765?check_suite_focus=true). It would be great to get that dependency updated.

cseufert:
It is probably time to upgrade to more modern prettier and drop support for older nodejs versions. I tried this somewhat recently, and there is a lot to fix, this is a great first step.

czosel:
@alexander-akait done :tada:

cseufert:
This looks great.

czosel:
It took us long enough to get this merged 🙈 thanks again @fisker!

czosel:
Released in `v0.19.4` :tada:

PR #1949, opened 2022-03-20
Fix trailing comma on match body

### LINKED ISSUE #1948: Prettier is removing trailing commas from match expressions when `trailingCommaPHP=true`.
Prettier is removing trailing commas from match expressions when `trailingCommaPHP=true`.

Wasn't sure if you wanted a separate issue for this or just to track it here.

_Originally posted by @hackel in https://github.com/prettier/plugin-php/issues/1691#issuecomment-1073304082_

cseufert:
fixes #1948

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -3062,7 +3062,11 @@ function printNode(path, options, print) {
             ])
           ),
           ") {",
-          indent(concat(arms)),
+          group(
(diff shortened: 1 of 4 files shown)


czosel:
Released in `v0.18.4` :tada:

PR #2003, opened 2022-06-03
Attempt to fix the formatting of long cases in match statement

cseufert:
This is outputting valid PHP however, for some reason I do not fully understand, it breaks short multiple match cases onto multiple lines.


cseufert:
@czosel do you have any time to play with this, and see if you can get a more consistent output for match cases? I am unsure why the groups are working the way they are.

czosel:
@cseufert Sorry for the delay - I think I have a working version in #2054

czosel:
Superseded by #2054

PR #2059, opened 2022-09-06
chore!: upgrade dependencies

czosel:
BREAKING CHANGE: This drops support for node v12.


cseufert (approved):
Looks good, however, did you try updating the mem package to ^9.0 or does the ESM module thing break things?

czosel:
Jep, for ESM support we'll have to upgrade to prettier v3.

PR #2070, opened 2022-09-21
Deprecated PSR-2 braceStyle and use PER-CS instead

### LINKED ISSUE #2060: PSR-12 is this properly supported yet?
Hello, I am trying to use this with psr-12 standard, but I think I am doing something wrong. I want to use this plugin before PHP_CodeSniffer, which uses the PSR-12 standard. I found several threads already speaking about it, but the I did not find anything in the documentation. 

Can you help me to set this up?

```bash
# my script

#! /usr/bin/env bash

run_prettier(){
files=$1  
 if [ "$files" != "" ]
then
  echo "Running Prettier... on $files"
  prettier --write $files
  if [ $? != 0 ]
  then
    echo "Fix the error before commit."
    exit 1
  else
    # Add back the modified/prettier files to staging
    echo "$files" | xargs git add
  fi
fi
}

# output
Running Prettier... on src/LoggingServicePackageProvider.php
[error] Invalid braceStyle value. Expected "1tbs" or "psr-2", but received "psr-12".
Fix the error before commit.
```
.prettierrc:
```json
{
    "phpVersion": "8.1",
    "printWidth": 80,
    "tabWidth": 4,
    "useTabs": false,
    "singleQuote": false,
    "trailingCommaPHP": true,
    "braceStyle": "psr-12",
    "requirePragma": false,
    "insertPragma": false
}
```

cseufert:
Update braceStyle to PSR-2 option to PSR-12, and deprecated PSR-2

PSR-2 and PSR-12 produce the exact same output.

Fixes #2060


czosel (approved):
👍

czosel:
Released in `v0.19.2` :tada:

PR #2088, opened 2022-10-17
WIP Fix for multiple inline expressions

### LINKED ISSUE #2022: Multiple shorthand echoes in a row break when formatted
**Prettier 2.4.0**

**PHP Plugin 0.18.4**

**Input:**

```php
<?= $var1 ?><?= $var2 ?>
```

**Output:**

```php
<?=
$var1
$var2
?>
```

**Expected behavior:**

The above output does not work, it is fundamentally broken code. 

There should be no change:

```php
<?= $var1 ?><?= $var2 ?>
```
...or the variables should be concatenated:

```php
<?= $var1 . $var2 ?>
```

Note that if there is a space in the middle, this works fine and is not changed:

```php
<?= $var1 ?> <?= $var2 ?>
```

cseufert:
Currently breaks outputing an inline comment

closes #2022


cseufert:
Not sure why this is causing a comment to disappear, @czosel, @alexander-akait any ideas?

PR #2089, opened 2022-10-18
Fix for function curly brace with attrs

### LINKED ISSUE #1870: PHP8: Inconsistent formatting when using attributes
**Prettier 2.4.1**

**PHP Plugin 0.17.6**

**Input:**

```php
use JetBrains\PhpStorm\Pure;

class Test
{
    /**
     * Method with an attribute.
     * @param string $foo
     * @return string
     */
    #[Pure]
    public function withAttribute(string $foo): string
    {
        return $foo;
    }

    /**
     * Method without any attributes.
     * @param string $foo
     * @return string
     */
    public function withoutAttribute(string $foo): string
    {
        return $foo;
    }
}
```

**Expected:**
Code is already correctly formatted => runnning Prettier should not result in any changes.

**Result:**
The opening brace of the method that contains the attribute gets moved, the other one does not.
```php
    public function withAttribute(string $foo): string {
        return $foo;
    }
```

**Additional note:**
Removing the `string $foo` parameter from the method signature makes the problem go away. So it seems to be a rather weird combination of factors that trigger the bug.

cseufert:
This fix excludes the function block attributes from the function definition, when chcking the line length.

closes #1870

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -1370,9 +1370,10 @@ function printClass(path, options, print) {
 
 function printFunction(path, options, print) {
   const node = path.getValue();
-  const declaration = [
-    ...printAttrs(path, options, print, { inline: node.kind === "closure" }),
-  ];
+  const declAttrs = printAttrs(path, options, print, {
+    inline: node.kind === "closure",
+  });
+  const declaration = [];
 
   if (node.isFinal) {
     declaration.push("final ");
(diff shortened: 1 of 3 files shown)


czosel (approved):
Nice work :100:

czosel:
Released in `v0.19.2` :tada:

PR #2102, opened 2022-11-03
Change php version in readme

jdeniau:
According to https://github.com/prettier/plugin-php/blob/v0.19.2/src/options.js#L10, the default value is now `7.0`


cseufert (approved):

czosel:
Sorry about the delay, thanks @jdeniau!

PR #2152, opened 2023-03-03
fix: update prettier

czosel:


czosel:
Friendly ping @cseufert :slightly_smiling_face: 

cc @alexander-akait @loilo

src/comments.js
```diff
@@ -8,15 +8,8 @@ const {
...
-const {
-  concat,
-  join,
-  indent,
-  hardline,
-  cursor,
-  lineSuffix,
-  breakParent,
-} = require("prettier").doc.builders;
+const { concat, join, indent, hardline, cursor, lineSuffix, breakParent } =
```
alexander-akait, line 11: As I remeber concat is deprecated, but it is not a problem for such update

alexander-akait (approved):
Also it will be good for dependeices, because now you can have only one prettier in your node_moduels

src/comments.js
czosel: That's right - we'll tackle that in the context of #1922 in a follow-up PR :+1:

czosel:
Thanks for the review @alexander-akait! :+1:

PR #2158, opened 2023-03-14
Use `clean.ignoredProperties`

fisker:
It's supported since v2(not sure exact version, but not important since it's a debug feature).


czosel (approved):
👍 nice!

czosel:
Released in `v0.19.4` :tada:

PR #2179, opened 2023-05-21
fix: add option for PHP 8.2

czosel:


cseufert (approved):
LGTM

however we still do not support DNF syntax for types, eg:
```php
  public function test((A & B) | C $request) {}
```

probably more of a note for the changelog

czosel:
Released in `v0.19.5` :tada:

PR #2197, opened 2023-07-25
Update package.json to allow prettier 3.x as peerDependency

ryankshaw:
It looks like this should run just fine with prettier 3.x so this just updates the package.json to reflect that


czosel:
Unfortunately tests start failing once you update the prettier devDependency. See #2200 for a WIP state of the upgrade.

PR #2200, opened 2023-08-12
feat: upgrade to prettier 3

czosel:


czosel:
This is WIP, and any help is appreciated. I'll print some of the open issues, since for some reason they don't show up in CI.

```
 FAIL   test-node  tests/enum/jsfmt.spec.js                                                                                                                                            
  ● enum.php                                                                                                                                                                           
                                                                                                                                                                                       
    TypeError: Cannot convert undefined or null to object                                                                                                                              
        at Function.keys (<anonymous>)                                                                                                                                                 
                                                                                                                                                                                       
      903 |                                                                                                                                                                            
      904 |   const getChildNodes = (node) =>                                                                                                                                          
    > 905 |     Object.keys(node)                                                                                                                                                      
          |            ^                                                                                                                                                               
      906 |       .filter(                                                                                                                                                             
      907 |         (n) =>                                                                                                                                                             
      908 |           n !== "kind" &&                                                                                                                                                  
                                                                                                                                                                                       
      at keys (src/comments.js:905:12)                                                                                                                                                 
      at getChildNodes (src/comments.js:920:10)                                                                                                                                        
      at getSortedChildNodes (node_modules/prettier/index.mjs:17203:64)                                                                                                                
      at node_modules/prettier/index.mjs:17208:52                                                                                                                                      
          at Array.flatMap (<anonymous>)                                                                                                                                               
      at getSortedChildNodes (node_modules/prettier/index.mjs:17207:6)                                                                                                                 
      at decorateComment (node_modules/prettier/index.mjs:17220:22)                                                                                                                    
      at decorateComment (node_modules/prettier/index.mjs:17231:14)                                                                                                                    
      at node_modules/prettier/index.mjs:17287:8                                                                                                                                       
          at Array.map (<anonymous>)                                                                                                                                                   
      at attachComments (node_modules/prettier/index.mjs:17286:38)                                                                                                                     
      at prepareToPrint (node_modules/prettier/index.mjs:17787:3)                                                                                                                      
      at printAstToDoc (node_modules/prettier/index.mjs:17721:20)                                                                                                                      
      at coreFormat (node_modules/prettier/index.mjs:18037:20)                                                                                                                         
      at formatWithCursor (node_modules/prettier/index.mjs:18235:14)                                                                                                                   
      at format (tests_config/run_spec.js:128:18)                                                                                                                                      
      at Object.<anonymous> (tests_config/run_spec.js:71:22) 
```

```
 FAIL   test-node  tests/class/jsfmt.spec.js                                                                                                                                           
  ● anonymous.php                                                                                                                                                                      
                                                                                                                                                                                       
    TypeError: Invalid value used as weak map key                                                                                                                                      
        at WeakMap.set (<anonymous>)                                                                                                                                                   

      126 |
      127 | async function format(source, filename, options) {
    > 128 |   const result = await prettier.formatWithCursor(
          |                  ^
      129 |     source,
      130 |     Object.assign({ filepath: filename }, options)
      131 |   );

      at getSortedChildNodes (node_modules/prettier/index.mjs:17213:19)
      at node_modules/prettier/index.mjs:17208:52
          at Array.flatMap (<anonymous>)
      at getSortedChildNodes (node_modules/prettier/index.mjs:17207:6)
      at decorateComment (node_modules/prettier/index.mjs:17220:22)
      at decorateComment (node_modules/prettier/index.mjs:17231:14)
      at decorateComment (node_modules/prettier/index.mjs:17231:14)
      at node_modules/prettier/index.mjs:17287:8
          at Array.map (<anonymous>)
      at attachComments (node_modules/prettier/index.mjs:17286:38)
      at prepareToPrint (node_modules/prettier/index.mjs:17787:3)
      at printAstToDoc (node_modules/prettier/index.mjs:17721:20)
      at coreFormat (node_modules/prettier/index.mjs:18037:20)
      at formatWithCursor (node_modules/prettier/index.mjs:18235:14)
      at format (tests_config/run_spec.js:128:18)
      at Object.<anonymous> (tests_config/run_spec.js:71:22)
```

czosel:
I was able to narrow it down a bit - the issues come from `getCommentChildNodes(node)`, which seems to behave differently to prettier v2.

src/comments.js
```diff
@@ -915,7 +919,9 @@ function getCommentChildNodes(node) {
           n !== "precedingNode" &&
           n !== "followingNode"
       )
-      .map((n) => node[n]);
+      .map((n) => node[n])
+      // make sure always AST nodes are returned
+      .filter((n) => typeof n === "object");
```
czosel, line 924: This fixed the `array` tests (we were returning non-AST nodes before)

src/comments.js
```diff
@@ -897,6 +897,10 @@ function isBlockComment(comment) {
 }
 
 function getCommentChildNodes(node) {
+  // TODO why is this called with node == `null`??
+  if (!node) {
```
fisker, line 901: This is unexpected. https://github.com/prettier/prettier/blob/e22a5516f5b2c44e6d68be4c0a7ba0b32c1b9f86/src/main/comments/attach.js#L43C5-L43C21


Prettier should not pass non-object here.

src/comments.js
fisker: This should also done in Prettier core.

src/comments.js
```diff
@@ -915,14 +919,16 @@ function getCommentChildNodes(node) {
...
+      // make sure always AST nodes are returned
+      .filter((n) => typeof n === "object");
 
   return getChildNodes(node);
 }
 
 function canAttachComment(node) {
   return (
-    node.kind && node.kind !== "commentblock" && node.kind !== "commentline"
+    node?.kind && node?.kind !== "commentblock" && node?.kind !== "commentline"
```
fisker, line 931: Same here.

fisker:
I'd suggest use new property `getVisitorKeys` instead of `getCommentChildNodes` since we are handle special cases inside.

```js
const ignoredKeys = new Set(["kind", "loc", "errors", "extra"]);

function getVisitorKeys(node, nonTraversableKeys) {
  return Object.keys(node).filter(
    (key) => !nonTraversableKeys.has(key) && !ignoredKeys.has(key),
  );
}
```

https://prettier.io/docs/en/plugins#optional-getvisitorkeys

czosel:
Thank you @fisker!  `getVisitorKeys` works much better than our previous implementation. We're down to 3 failing tests now :+1: (some detail about comment handling seems to have changed as well, I don't look into it yet)

czosel:
Update: It seems that the way `followingNode` is determined for comments slightly changed in Prettier 3. E.g. for the following input
```php
<?php
$a = new class (
    // 1
    $a,
    // 2
    $b
) {
  const foo = "bar";
};
```
`followingNode` of `// 1` is
- the `Variable` node `$a` for Prettier 2
- the `classconstant` node (`const foo = "bar";`) for Prettier 3

For this example, this looks like a regression - but I didn't check what exactly changed upstream yet and why. @fisker any pointers are much appreciated, as usual :wink:

fisker:
I took a quick look.

When attaching `// 1`, Prettier found `class` is the "enclosingNode", and it's children doesn't include `variable[name="a"]` or `variable[name="b"]` (since they belong to `new`, the AST looks strange to me). 

Not sure how it works in Prettier v2, but it won't be easy to fix.

fisker:
@czosel I pushed an ugly workaround, seems the result is acceptable. https://github.com/prettier/plugin-php/pull/2200/commits/8cfceb9259efb2c7a913a701f882492cb55ccb16

czosel:
Thanks so much @fisker, this looks great! :100: The `new class` is not super common in PHP anyway, so the hack seems fine to me.

PR #2212, opened 2023-08-16
Simplify `print()`

fisker:
Since we are going to drop support for Prettier<3, we can use [simplified `print()` introduced in Prettier v2.3.0](https://prettier.io/blog/2021/05/09/2.3.0#simplified-print-callback-10557httpsgithubcomprettierprettierpull10557-by-fiskerhttpsgithubcomfisker).


czosel (approved):
Very nice, I didn't know about this :+1:

PR #2214, opened 2023-08-17
Use `AstPath` getters

fisker:


src/printer.js
```diff
@@ -1027,6 +1018,7 @@ function printLines(path, options, print, childrenAttribute = "children") {
           ? ""
           : [beforeCloseTagInlineNode, "?>"];
 
+      // Bug?
       const nextV = path.getNode(index + 1);
```
fisker, line 1022: @czosel This looks like a bug to me, `getNode` is used to get ancestors, but this seems mean to get next sibling?

czosel: @fisker It's either bad naming or a bug, yeah. I updated the comment for later investigation.

czosel:
Released in `v0.20.1` :tada:

PR #2228, opened 2023-08-28
Fix magic methods detection

fisker:
Input:

```php
<?php
class Foo
{
    public function __pRoTo__() {}
    public function cOnStRuCtOr() {}
}
```

Output:

```php
<?php
class Foo
{
    public function Have not implemented kind undefined yet.()
    {
    }
    public function Have not implemented kind undefined yet.()
    {
    }
}
```

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -2888,7 +2888,7 @@ function printNode(path, options, print) {
     case "error":
     default:
       // istanbul ignore next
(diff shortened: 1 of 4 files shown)


czosel (approved):
Nice! :+1: IIRC, there is an open issue about this somewhere in our issue tracker :upside_down_face:

PR #2235, opened 2023-08-31
Update build script

fisker:


src/needs-parens.js
```diff
@@ -92,8 +91,6 @@ function needsParens(path) {
           }
 
           if (pp === np && key === "right") {
-            assert.strictEqual(parent.right, node);
```
fisker, line 95: This actually meanless, since `key === "right"` already ensure it's accessing `parent.right`.

czosel:
@loilo could you take a look at this?

src/needs-parens.js
loilo: I mean, I kind of fear that we could be running into a Chesterton's Fence problem here. But I don't see any case where that line would be needed either. 🤷‍♂️

loilo (commented):
Looks all good to me. Standalone file builds and works fine, and also works in the playground. 👍

loilo (approved):

PR #2236, opened 2023-08-31
Fix coverage report, ast compare test

fisker:
coverage report is currently broken.


czosel (approved):
Nice! :100: 

For the coverage check to be meaningful, it would be nice to have CI fail if coverage decreased, or even set the 100% target (and add explicit ignores where we still need coverage), wouldn't it?

czosel:
:+1: sure! looks like this one needs a rebase now.

PR #2258, opened 2023-10-06
fix(match): Indent match conditions

claytonrcarter:
This updates `match($cond)` printing to ... er ... match `if ($cond)` (etc) when the condition can't fit onto a single line.

## Input
```php
match($really_really_really_really_really_really_really_really_long_variable_name) {
default => null
};
```

## Current behavior on `master`
```php
match(
$really_really_really_really_really_really_really_really_long_variable_name
) {
    default => null
};
```

## Behavior with this change
```php
match(
    $really_really_really_really_really_really_really_really_long_variable_name
) {
    default => null
};
```

I added a test to the snapsnots, and this change also corrects/changes one of the other `match` snapshots.

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -2847,7 +2847,7 @@ function printNode(path, options, print) {
       }, "arms");
       return group([
         "match (",
-        group([softline, indent(print("cond")), softline]),
+        group([indent([softline, print("cond")]), softline]),
         ") {",
         group(indent([...arms])),
         " ",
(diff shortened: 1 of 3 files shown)


cseufert (approved):
Thanks, and looks good

czosel (approved):
LGTM as well, thanks for contributing and sorry for the delay!

czosel:
Released in `v0.21.0` :tada:

PR #2320, opened 2024-01-29
Make chain breaking as similar to original prettier (js) as much as possible

eldair:
There is a hardcoded line which is the cause of chain breaks even though the expression would fit in one line. This is a quick and easy solution for https://github.com/prettier/plugin-php/issues/1545

I have adjusted test so they all pass but if this is merged than some of those could be removed as redundant.


eldair:
I have updated the pull request to include a new option where user can define their own max calls in a chain before break occurs. The default option is left as is (I think the number 3 is based on similar option in main prettier repo) so it would not affect current codebases.

czosel:
Hi @eldair, thanks for contributing! I think adding an option for this doesn't fit to the [Prettier option philosophy](https://prettier.io/docs/en/option-philosophy). Did you already look into how Prettier for JS handles this? It would be awesome if we could solve it the same way.

eldair:
Me again, I have started digging into `isSimpleCallArgument` of prettier. I will get back if I manage to port it for php to be on par with js version.

eldair:
@czosel Pull request can be checked now. The code is not perfect and I tried to test as many combination as I could think of but it is much closer to the JS counterpart and I think better than the current implementation.
Let me know if something should be changed/fixed/improved.

cseufert:
Change looks good to me, however, can you reformat with prettier and commit? See CI / Lint action, at least `print.mjs` needs reformating.

eldair:
@cseufert thanks for checking it out, i've fixed the file (on other note, it's weird that vscode prettier formats those lines differently from cli prettier :/ )

czosel:
LGTM, thanks for the awesome contribution @eldair! :tada:

czosel:
Released in `v0.22.2` :tada:

eldair:
Woohoo  🎉

PR #2321, opened 2024-02-03
chore: upgrade dependencies

czosel:


src/pragma.mjs
```diff
@@ -1,4 +1,4 @@
-import memoize from "mem";
+import memoize from "memoize";
```
fisker, line 1: Maybe we can replace it with this?

```js
const memoize = getPageLevelDocBlock => {
	const cache = new Map();
	return text => {
		if (!cache.has(text)) {
			cache.set(text, getPageLevelDocBlock(text));
		}
		return cache.get(text);
	};
};
```

czosel: Thanks for the review @fisker! I put a generic memoize function in the utils and dropped the dependency. Shall we still drop support for Node 16?

fisker: I don't know, personally I don't use old versions, but Prettier still support Node.js v14, maybe worth to keep support for 16 if not cost much to maintenance.

fisker: Looks like we don't have `engine` field at all, does that mean we can drop them anytime :smile:

czosel: :+1: I reverted the change to the node versions in CI. Where do I see the v14 baseline in prettier? `engines` is set to `>=18`: https://github.com/prettier/prettier/blob/main/package.json#L24

fisker: `engines` in https://unpkg.com/browse/prettier@3.2.4/package.json

fisker: https://github.com/prettier/prettier/blob/8cbee2e217baad7acf4cb3947834e8c1b41ed647/scripts/build/build-package-json.js#L40

czosel: Ah I see, thanks :) shall we add engines with node >= 16 here?

fisker: I don't know, I never use `engines` myself.

fisker (approved):

src/pragma.mjs
czosel: Then I’d say let’s leave it as is, ok?

PR #2322, opened 2024-02-04
Adjustment for overlooked nested lookups in chain breaks

eldair:
I have adjusted chain breaks for nested lookups which were overlooked in previous pull request. Also the break should now be consistent between `Test::foo()->`, `(new Test())->`, `$test['foo']` etc.


cseufert (approved):
Looks good, LGTM

czosel (approved):
LGTM as well, sorry about the delay @eldair!

PR #2396, opened 2024-08-13
Support PHP 8.3 typed class constants

mreiden:
Depends on php-parser dependency support.

https://github.com/Kenneth-Sills/php-parser/pull/1


czosel:
@mreiden The PR in the parser has landed in `v3.2.0` :tada: Do you want to update your PR?

src/parser.mjs
```diff
@@ -14,6 +14,7 @@ function parse(text, opts) {
   const parser = new engine({
     parser: {
       extractDoc: true,
+      version: opts.phpVersion,
```
czosel, line 17: This won't be needed once https://github.com/glayzzle/php-parser/pull/1150 has landed.

genintho:
@czosel this will need a new release of php-parser to include the fix to the original implementation.

src/parser.mjs
czosel: I think the idea is that we always try to parse with the latest PHP version, and only use the prettier version setting to control our printing (i.e. don't print features that are unsupported in the configured version). Since https://github.com/glayzzle/php-parser/pull/1150 is now landed, I don't really see a reason to change this. I'd be very open to change update the default for our version setting to something more reasonable though! :+1:

czosel:
> @czosel this will need a new release of php-parser to include the fix to the original implementation.

Done! Sorry about the delay.

PR #2422, opened 2025-04-17
feat: format new without parenthesis in PHP 8.4

claytonrcarter:
This adds support for formatting things like `(new Foo())->bar` as `new Foo()->bar`. Support for this syntax was [added in PHP 8.4](https://www.php.net/releases/8.4/en.php#new_without_parentheses) and [very recently merged and released](https://github.com/glayzzle/php-parser/pull/1146) in php-parser. 

~~**Note** that php-parser 3.2.2 introduced a parser bug that affects how we format specifically placed comments in class constant assignments. My opinion is that this is an extreme edge case issue and it's worth the temporary regression in order to gain support for one of the highlighted features of php 8.4. (There is already a [pending PR](https://github.com/glayzzle/php-parser/pull/1152) to fix this in php-parser, and we can rollout an update to fix it here when that lands.~~ edit: this has been fixed

- bumps php-parser to the newly released v3.2.3
- ~~**breaking:** removes (temporarily) support for some comments in class constant assigments~~ 
- adds a new `phpVersion` option for `8.4`
- adds support for the above syntax, with tests
- misc light refactoring to support these changes

**Verification**
1. I used this to format all text fixture files and confirmed that now (relevant) issues were introduced.
2. I used this to format a small Laravel project of mine (300+ files, 27k lines of code) and confirmed that the tests still passed.

**Try it out**
```php
<?php

class Foo 
{
    public $bar = 123;
}

echo (new Foo)->bar;
```
This should be reformatted to
```diff
-echo (new Foo)->bar;
+echo new Foo()->bar;
```
And `php foo.php` should print `123` in both cases (though the latter will fail in php <= 8.3).

### DIFF
--- a/src/parser.mjs
+++ b/src/parser.mjs
@@ -1,4 +1,5 @@
 import engine from "php-parser";
+import options from "./options.mjs";
 
 function parse(text, opts) {
   const inMarkdown = opts && opts.parentParser === "markdown";
@@ -10,10 +11,15 @@ function parse(text, opts) {
   // Todo https://github.com/glayzzle/php-parser/issues/170
   text = text.replace(/\?>\n<\?/g, "?>\n___PSEUDO_INLINE_PLACEHOLDER___<?");
 
+  const latestSupportedPhpVersion = Math.max(
+    ...options.phpVersion.choices.map((c) => parseFloat(c.value))
+  );
+
   // initialize a new parser instance
   const parser = new engine({
     parser: {
       extractDoc: true,
+      version: `${latestSupportedPhpVersion}`,
     },
     ast: {
       withPositions: true,
--- a/src/printer.mjs
+++ b/src/printer.mjs
@@ -37,6 +37,7 @@ import {
   isReferenceLikeNode,
   normalizeMagicMethodName,
   isSimpleCallArgument,
+  isMinVersion,
 } from "./util.mjs";
 
 const {
@@ -64,10 +65,6 @@ const {
   isNextLineEmpty,
   isPreviousLineEmpty,
 } = prettierUtil;
-
-function isMinVersion(actualVersion, requiredVersion) {
-  return parseFloat(actualVersion) >= parseFloat(requiredVersion);
-}
 
 function shouldPrintComma(options, requiredVersion) {
   if (!options.trailingCommaPHP) {
--- a/tests/parens/jsfmt.spec.mjs
+++ b/tests/parens/jsfmt.spec.mjs
@@ -1,1 +1,2 @@
 run_spec(import.meta, ["php"]);
+run_spec(import.meta, ["php"], { phpVersion: "8.4" });
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   ],
   "dependencies": {
     "linguist-languages": "^8.0.0",
-    "php-parser": "^3.2.4"
+    "php-parser": "^3.2.5"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.27.2",
--- a/src/needs-parens.mjs
+++ b/src/needs-parens.mjs
@@ -1,4 +1,9 @@
-import { getPrecedence, shouldFlatten, isBitwiseOperator } from "./util.mjs";
+import {
+  getPrecedence,
+  shouldFlatten,
+  isBitwiseOperator,
+  isMinVersion,
+} from "./util.mjs";
 
 function needsParens(path, options) {
   const { parent } = path;
@@ -128,13 +133,16 @@ function needsParens(path, options) {
     }
     case "clone":
     case "new": {
+      const requiresParens =
+        node.kind === "clone" ||
+        (node.kind === "new" && !isMinVersion(options.phpVersion, "8.4"));
       switch (parent.kind) {
         case "propertylookup":
         case "nullsafepropertylookup":
         case "staticlookup":
         case "offsetlookup":
         case "call":
-          return key === "what";
+          return key === "what" && requiresParens;
         default:
           return false;
       }
--- a/src/needs-parens.mjs
+++ b/src/needs-parens.mjs
@@ -1,6 +1,6 @@
 import { getPrecedence, shouldFlatten, isBitwiseOperator } from "./util.mjs";
 
-function needsParens(path) {
+function needsParens(path, options) {
   const { parent } = path;
 
   if (!parent) {
--- a/src/options.mjs
+++ b/src/options.mjs
@@ -24,6 +24,7 @@ export default {
       { value: "8.1" },
       { value: "8.2" },
       { value: "8.3" },
+      { value: "8.4" },
     ],
   },
   trailingCommaPHP: {
--- a/src/util.mjs
+++ b/src/util.mjs
@@ -709,6 +709,10 @@ function memoize(fn) {
   };
 }
 
+function isMinVersion(actualVersion, requiredVersion) {
+  return parseFloat(actualVersion) >= parseFloat(requiredVersion);
+}
+
 export {
   printNumber,
   getPrecedence,
@@ -740,4 +744,5 @@ export {
   normalizeMagicMethodName,
   isSimpleCallArgument,
   memoize,
+  isMinVersion,
 };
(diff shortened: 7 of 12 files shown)


claytonrcarter:
Marking as draft for now because I want to do some additional testing, and it looks like there are CI failures.

claytonrcarter:
Thank you for nudging me on this (I had been away). I've updated the PR to use `php-parser` 3.2.4 and dropped the feature regression related to https://github.com/glayzzle/php-parser/pull/1152. I am also seeing CI failures, but I recall from when I first started this that some of those may be related to parsing the AST. I'll try to dig in and share what I find.

claytonrcarter:
I can reproduce the CI failure locally when running `AST_COMPARE=1 yarn test:standalone` (which is [what CI does](https://github.com/prettier/plugin-php/blob/main/.github/workflows/nodejs.yml#L37)). 

I'm seeing 2 failures:
1. there is a diff in the AST between the old and new files
2. the new files are failing to parse

Here's an example of # 1:
```diff
        Object {
        "expression": Object {
            "kind": "assign",
            "left": Object {
            "curly": false,
            "kind": "variable",
            "name": "class",
            },
            "operator": "=",
            "right": Object {
+           "kind": "bin",
+           "left": Object {
                "arguments": Array [],
-           "kind": "call",
+             "kind": "new",
                "what": Object {
-             "kind": "propertylookup",
-             "offset": Object {
-               "kind": "identifier",
-               "name": "veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongMethod",
+               "kind": "name",
+               "name": "Foo",
+               "resolution": "uqn",
+             },
            },
-             "what": Object {
+           "right": Object {
                "arguments": Array [],
-               "kind": "new",
+             "kind": "call",
                "what": Object {
                "kind": "name",
-                 "name": "Foo",
+               "name": "veryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongMethod",
                "resolution": "uqn",
                },
            },
-           },
+           "type": "->",
            },
        },
```

I don't understand all of the details in that, but it looks like `(new Foo)->method()` is treated as a binary expression with a call expression, while `new Foo()->method()` is treated as just a call expression. It seems to me that this is the intended behavior with this change, but I don't know how to work around that in CI. Perhaps we need to update the AST tests to ... what? Allow specific ART diff regressions for specific files and/or versions? Or we could snapshot the the intended diff and use that to see if a diff is "known"? I'm stabbing in the dark on this; and input is welcome!

As for # 2, an example error is
```
/Users/crcarter/src/Other/plugin-php/tests/parens/new.php parse
    SyntaxError: Parse Error : syntax error, unexpected '[', expecting ';' on line 34
      32 | $var = new $foo()->bar;
      33 | $var = new $bar->y()->x;
    > 34 | $var = new foo()[0];
         |                ^
      35 | $var = new foo()[0]["string"];
      36 | $var = new $a->b();
      37 | $var = new $a->b();
```

In this case, the input snippet `$var = (new foo)[0];` has been reformatted to `$var = new foo()[0];` (which is valid in PHP 8.4). My hunch is that the parser is trying to parse the code fixtures as <8.4, but I am attempting to [pass the correct version to the parser](https://github.com/prettier/plugin-php/pull/2422/files#diff-21ed98bd3e0c36a80dfb910ee85e3b729243575288fd7d5db3b6748736b67271R13-R19) already.

I'll keep tinkering with this; but if someone can confirm that I'm on the right track (or not!) that would be really helpful. Thank you!

czosel:
I could reproduce the issue in the php-parser repository by adapting a test case. We'll have to fix this in the parser.

```
  ● Test classes › 8.4 allow new without parenthesis

    SyntaxError: Parse Error : syntax error, unexpected '[', expecting ';' on line 1

      347 |   message += " on line " + this.lexer.yylloc.first_line;
      348 |   if (!this.suppressErrors) {
    > 349 |     const err = new SyntaxError(
          |                 ^
      350 |       message,
      351 |       this.filename,
      352 |       this.lexer.yylloc.first_line,

      at Parser.raiseError (src/parser.js:349:17)
      at Parser.raiseError [as error] (src/parser.js:397:15)
      at Parser.error [as expectEndOfStatement] (src/parser.js:508:10)
      at Parser.expectEndOfStatement [as read_statement] (src/parser/statement.js:418:14)
      at Parser.read_statement [as read_top_statement] (src/parser/statement.js:78:21)
      at Parser.read_top_statement [as read_start] (src/parser/main.js:18:19)
```

src/printer.mjs
```diff
@@ -777,11 +774,12 @@ function printBinaryExpression(
       parts.push(print("left"));
     }
 
-    const shouldInline = shouldInlineLogicalExpression(node);
-
-    const right = shouldInline
-      ? [node.type, " ", print("right")]
-      : [node.type, line, print("right")];
+    const right =
+      node.left.kind === "new" && node.type !== "."
```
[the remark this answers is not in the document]
claytonrcarter: Done. I was able to remove several pieces of really fiddly code, like this. 👍

claytonrcarter:
This now includes the fix in https://github.com/glayzzle/php-parser/pull/1162, and ... CI is passing! No AST massaging needed! Thank you all for your help and guidance; truly a team effort! 

@hackel Ryan, I added a some code to the snapshots based on your examples and they seem to formatting as you would expect. Can you can try this out again on your code base and see if it's any better?

src/parser.mjs
```diff
@@ -10,11 +10,16 @@ function parse(text, opts) {
   // Todo https://github.com/glayzzle/php-parser/issues/170
   text = text.replace(/\?>\n<\?/g, "?>\n___PSEUDO_INLINE_PLACEHOLDER___<?");
 
+  const parserOpts = Object.assign(
+    { extractDoc: true },
+    // only pass the version if user requested 8.4 syntax; parser is stricter
+    // about allowed syntax than we are and currenly defaults to support for 8.3
+    opts && opts.phpVersion === "8.4" ? { version: opts.phpVersion } : {}
```
cseufert, line 17: What happens when PHP 8.5 is released here?

cseufert:
> > CI is passing
> 
> Well ... _almost_ passing. Codecov seems to be flagging me on changes to `util.mjs`, even though it looks like my only change to that file _is_ covered. I may not be reading the report correctly, though. Suggestions?

I dont think codecov knows what is going on there

src/parser.mjs
claytonrcarter: > What happens when PHP 8.5 is released here?

Good point. This should be more flexible.

> the parser's default version is the latest supported one

Do you mean that we should be using the latest version supported by `php-parser`, or *our* latest supported version? I don't see a way to do the former because I don't see that `php-parser` lists its supported versions in any single location, but I have implemented the latter in my latest push by inspecting the available options listed at https://github.com/prettier/plugin-php/blob/main/src/options.mjs#L10-L27

The upside of this is that we can parse php 8.4 code and [reformat it to an older version](https://github.com/prettier/plugin-php/pull/2422/files#diff-bddfb80378854eaac144c32c01cfaa2d5839c699f718b8b60657dcd737b6e49dR4-R38), and I can remove the `new84` snapshots I had added, instead incorporating those into the regular `new` snapshots. 👍 

The downside is that – and this speculative; I haven't thought it through very much – if there is a backward incompatible syntax change between the PHP version the user is using and the one that we are using, it's possible that this may prevent us from parsing and formatting valid code. These happen infrequently and after being deprecated, but it's possible. 🤷

czosel: I actually meant just changing the default version in the parser, as I'm proposing here: https://github.com/glayzzle/php-parser/pull/1163
But I like your approach of explicitly setting the latest version that this plugin supports. That way we are a little more resilient and the code is more explicit.

I'd value being able to reformat code to an older version higher than the (currently more theoretical) risk of not being able to parse old code. If we ever see such an issue, we can solve it when it comes to it :-)

tests/parens/jsfmt.spec.mjs
```diff
@@ -1 +1,2 @@
 run_spec(import.meta, ["php"]);
+run_spec(import.meta, ["php"], { phpVersion: "8.4" });
```
czosel, line 2: Do we really need this? The only differences would be related to `new`, which are covered in the `new` tests, right?

czosel:
Apart from the last minor question, this is ready to be merged from my point of view :100:

hackel:
@claytonrcarter I ran the new code and it all looks great on my end.  Great job!

tests/parens/jsfmt.spec.mjs
czosel: Alright! Thanks :)

PR #2424, opened 2025-05-02
fix: attribute PSR compatibility

### LINKED ISSUE #2360: Breaks #[Route] lines in Symfony
Overall, your plugin is doing great work! The only hiccup I've noticed is that when it comes to Symfony, the formatting of #[Route] lines isn't as sleek as it could be. Specifically, it tends to spread out over multiple lines when it could neatly fit into just one.

For instance, this:

`
#[Route('route/path', name: 'very_very_very_very_very_very_long_route_name', methods: ['GET'])]
`

would be converted to this

`
#[
    Route('
        'route/path',
        name: 'very_very_very_very_very_very_long_route_name',
        methods: ['GET']
    )
]
`

It's a bit awkward, isn't it? Most PHP developers would agree that since #[Route] statements are typically used in a single line, nor does it wrap anything like functions with arguments; so it looks awkward.

I've tried delving into the plugin's source to find a workaround, but no luck so far in implementing an ignore for the #[Route] pattern or anything similar. Any suggestions or insights would be greatly appreciated!

apollisa:
This partially fixes #2360, in the sense that formatting now matches PSR guidelines. It will still split across multiple lines if need be though.

### DIFF
--- a/src/printer.mjs
+++ b/src/printer.mjs
@@ -1170,30 +1170,25 @@ function printAttrs(path, options, print, { inline = false } = {}) {
     return [];
   }
   path.each(() => {
-    const attrGroup = ["#["];
+    const attrGroup = [];
     if (!inline && allAttrs.length > 0) {
       allAttrs.push(hardline);
     }
-    attrGroup.push(softline);
-    path.each(() => {
+    path.each((_, index) => {
       const attrNode = path.node;
-      if (attrGroup.length > 2) {
-        attrGroup.push(",", line);
-      }
       const attrStmt = [attrNode.name];
       if (attrNode.args.length > 0) {
         attrStmt.push(printArgumentsList(path, options, print, "args"));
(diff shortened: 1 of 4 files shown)


fisker:
Hello, @apollisa The test fails because the AST seems changed, run `AST_COMPARE=1 yarn test` you'll see the failure on local.

fisker:
I'm not sure if this will be accepted, but if it's expected to produce different AST, you'll need change the old shape to new shape in https://github.com/prettier/plugin-php/blob/main/src/clean.mjs

fisker:
Looking at how `printAttrs` is called, I think you can 

```js
if (node.kind === 'arrowfunc' || node.kind === 'parameter' /* || other kinds */) {
  // Update node.attrGroups
}
```

`clean` should also have a third argument to access parent, I believe.

PR #2434, opened 2025-07-08
feat!: bump default formatting version to 8.3

czosel:
BREAKING CHANGE: If you didn't set the `phpVersion` option explicitly, the formatting will assume PHP 8.3 now (instead of 7.0).


src/options.mjs
```diff
@@ -1,11 +1,66 @@
...
+import path from "path";
+
 const CATEGORY_PHP = "PHP";
 
+/**
+ * Detect minimum PHP version from the composer.json file
+ * @param def {string} Default PHP version to use if not found
+ * @return {string} The PHP version to use in the composer.json file.
+ */
+function getComposerPhpVer(def) {
```
czosel, line 11: Since we're not doing anything with the default version (`def`), would it be cleaner to just return `null` or `undefined` if we can't find any? And then change the callsite `getComposerPhpVer() ?? "8.3"`

tests/composer-version/composer-version.spec.mjs
```diff
@@ -0,0 +1,208 @@
...
+      2
+    );
+
+    fs.writeFileSync(tempComposerPath, composerContent);
+
+    process.chdir(tempDir);
+
+    const result = getComposerPhpVer("default");
+    expect(result).toBe(expectedVersion);
+  });
```
czosel, line 107: Is this an accidental copy of the previous test case?

czosel (commented):
Just some minor things, nice work!!

tests/composer-version/composer-version.spec.mjs
cseufert: Ooops, have sorted this now.

src/options.mjs
cseufert: yer, null return is much cleaner, I have changed it

cseufert:
@czosel 
I feel like this is ready to merge, however I did have to change the jest configuration, I think its for the best though.

cseufert:
I have updated the docs, however thinking about it further, maybe `auto` should be the default, and prettier should return an error if the proejcts php version cannot be determined?

cseufert:
I have re-implemented the default version, check out the readme, but in essence `auto` is default, `composer` is commposer or fail, and existing versions still work. Fallback is now the latest version supported (8.4), I have re-based this on the main branch as well, and updated the test logic for some of the 8.4 tests.

Apologies, as this has turned into a rather large PR though now, as multiple things have changed at once.

https://github.com/prettier/plugin-php/blob/1ea117767920d97873cdcea7ded9af88d2a903c9/README.md?plain=1#L191-L205

src/options.mjs
```diff
@@ -1,30 +1,125 @@
...
+      getComposerPhpVersion() ?? LATEST_SUPPORTED_PHP_VERSION;
+  } else if (options.phpVersion === "composer") {
+    const v = getComposerPhpVersion();
+    if (v === null) {
+      throw new Error(
+        `Could not determine PHP version from composer; ${getComposerError}`
+      );
+    } else {
+      options.phpVersion = v;
+    }
```
czosel, line 100: Nitpick: extra else
```suggestion
    }
    options.phpVersion = v;
```

README.md
```diff
@@ -176,17 +176,33 @@ await prettier.format(YOUR_CODE, {
...
+current directory or any parent directory, the plugin will use a minimum supported php version from
+`{"require":{"php":"..."}}` to set the phpVersion. If not found or not able to be parsed, it will default to latest
+supported PHP version, which is currently `8.4`.
+
+You set the `phpVersion` to `composer` and this will only use the `composer.json` file to determine the php
+version, prettier will crash if the PHP cannot be determined.
+
+You can also set the `phpVersion` to a specific version, such as `7.4`, `8.0`, `8.1`, `8.2`, or `8.3`.
+
+**Please Note:** If the phpVersion is not set correctly for your environment, this plugin will product code that could
```
czosel, line 203:
```suggestion
**Please Note:** If the phpVersion is not set correctly for your environment, this plugin will produce code that could
```

czosel:
@cseufert Looks great!! :100: A few very small cleanups, then we're good to go :rocket:

tests/composer-version/composer-version.spec.mjs
```diff
@@ -202,7 +170,25 @@ describe("getComposerPhpVer", () => {
...
+    expect(getComposerPhpVer()).toBe("7.2");
+  });
+
+  test("returns null when composer.json is malformed", () => {
+    // Create a malformed JSON file (invalid syntax)
+    const malformedContent = `{
+      "name": "test/package",
+      "require": {
+        "php": "^7.4"
+      }, // Invalid trailing comma
```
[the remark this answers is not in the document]
cseufert: I have simplified in the invalid json example

czosel:
All good now! Thanks @cseufert :100:

czosel:
Released in `v0.24.0` :tada:

PR #2453, opened 2026-02-05
fix: avoid adding parens to static property in new expression

### LINKED ISSUE #2441: Bug: Prettier incorrectly adds parentheses when using new with a static property
Description:

When formatting code with Prettier in PHP 8.4, parentheses are incorrectly added to a static property when instantiating an object. This causes a syntax error and breaks the application.

Example:

Before Prettier:

`$identifier = new Yii::$app->class([]);`

After Prettier:

`$identifier = new Yii::$app()->class([]);`

Expected behavior:
Prettier should leave the code as-is without adding parentheses when using new on a static property.

Actual behavior:
Prettier adds () to the static property, resulting in invalid syntax.

Impact:
Code breaks when executed, causing runtime errors in applications using this pattern.

sorafujitani:
## Summary

Fixes #2441

When formatting code like `new Yii::$app->class([])`, the formatter was incorrectly adding parentheses to produce `new Yii::$app()->class([])`. This changes the semantics from accessing a static property (`Yii::$app`) to calling a static method (`Yii::$app()`).

## Changes

- Added logic in `printer.mjs` to skip adding parentheses for `new` expressions when:
  - The `new` is inside a member chain (parent is a lookup node or call)
  - The constructor has no arguments
  - The original source did not have parentheses
  - The target is a `staticlookup` (e.g., `Yii::$app`)

## Test plan

- Added test cases in `tests/parens/new.php` for Issue #2441
- All existing tests pass (592 tests)
- AST comparison tests pass (1166 tests)

### DIFF
--- a/src/printer.mjs
+++ b/src/printer.mjs
@@ -2103,16 +2103,17 @@ function printNode(path, options, print) {
       } else {
         const isExpression = ["call", "offsetlookup"].includes(node.what.kind);
 
+        // Skip parens for staticlookup in member chain to preserve original syntax
         const isInMemberChain =
           path.parent &&
           (isLookupNode(path.parent) || path.parent.kind === "call");
-        const whatEndOffset = locEnd(node.what);
-        const nextCharIndex = getNextNonSpaceNonCommentCharacterIndex(
-          options.originalText,
-          whatEndOffset
-        );
         const hasOriginalParens =
-          options.originalText.charAt(nextCharIndex) === "(";
+          options.originalText.charAt(
+            getNextNonSpaceNonCommentCharacterIndex(
+              options.originalText,
+              locEnd(node.what)
+            )
+          ) === "(";
         const shouldSkipParens =
           isInMemberChain &&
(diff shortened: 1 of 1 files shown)


MasonD:
Indeed, the parser itself is incorrect in its handling of the new operator. It might be more illustrative to use parentheses like this:
```
new (Foo::$bar->baz)();
```
The above is identical semantically to `new Foo::$bar->baz();`, but results in the following AST, (`new Foo::$bar->baz();` should give this same AST but is bugged)  
```
{
  "kind": "expressionstatement",
  "expression": {
    "kind": "new",
    "what": {
      "kind": "propertylookup",
      "what": {
        "kind": "staticlookup",
        "what": {
          "kind": "name",
          "name": "Foo",
          "resolution": "uqn"
        },
        "offset": {
          "kind": "variable",
          "name": "bar",
          "curly": false
        }
      },
      "offset": {
        "kind": "identifier",
        "name": "baz"
      }
    },
    "arguments": []
  }
}
```

PR #2458, opened 2026-02-23
Property hook formatting

jorgsowa:
I forked prettier plugin for PHP as well, because it didn't cooperate well with property hooks. There is one commit on my fork to fix it: https://github.com/jorgsowa/plugin-php/commit/b35fa5a74deaf87683c7b357be0c25bf997b5b47

I mark the PR as draft only, because I can't remove the references to my fork today. Feel free to modify it, or I will change it when I have access to my laptop.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -17,8 +17,8 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@v6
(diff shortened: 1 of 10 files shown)


mreiden:
Thanks for starting this. The parser side has since landed upstream (php-parser 3.7.0, glayzzle/php-parser#1142), so the fork dependency is no longer needed. I've opened #2516 building on the same idea, adding hook attributes, hooks on promoted constructor parameters, the PER-CS one-line form for body-less hooks (`{ get; set; }`), and the `abstract`/`final`/`private(set)` property modifiers, which the printer was also dropping. Happy to fold it into this PR instead if you'd prefer.

PR #2502, opened 2026-06-02
Update php-parser, fix unary snapshot tests, and make multiple fixes for attribute formatting.

### LINKED ISSUE #2293: Syntax error when attributes are defined to enum cases
**@prettier/plugin-php v0.22.1**
[Playground link](https://loilo.github.io/prettier-php-playground/#N4IgDgTgpgLjCWUIHkwIPYDsDOIBcok8mMA6vACYwAW+AHAAwA0IMAhgEblW14AsLAK7YoAFU648AMzYAbES2zEA5rKgBFQehhR8M+VBZhqYAGpIlWfCACsAOj4gWMCG3iyVAYXQBbH2wAFAAkA-BdBQxAOVwBjKABlGABPNWswJABaGNwWaABHQXhoANdlfz05BRBiEQgYErYytgqDAF8WKAp4GHQUNHgsSVA2bBgWkXbqzDBBMbwQAB4AfmMwAB1MDcw2HyhsMDY4gAIAQTB1tYBRTEEfbABuDY3hKFPztbX4wXPemA+TuAQeAcWZ7D4AET2MSB-Swj02mCgNx8R3iclmbAwmGutzwR1GQMwyg2wA2R3JRwAxABtSHYaHwWGYAAUAHIALIQOysgCUAF0yRSYiNXuyAEpHAC8R1ZPggrPhgvJNLpDKZbM52G5-KVR2FIiO4viUplcuwCqemApVNpUJhWI18Gw5p1VqFIsNAEl4sbpbKneb4a0NiBWkA)

**Input:**
```php
<?php

namespace App\Enums;

use App\Support\Attributes\Description;

enum SalutationEnum: string
{
    #[Description('Mr.')]
    case MR = 'mr';

    #[Description('Mrs.')]
    case MRS = 'mrs';

    #[Description('Miss')]
    case MISS = 'miss';
}

```

**Output:**
```php
Parse Error : syntax error, unexpected 'case' (T_CASE) on line 10

   8 | {
   9 |     #[Description('Mr.')]
> 10 |     case MR = 'mr';
     |    ^
  11 |
  12 |     #[Description('Mrs.')]
  13 |     case MRS = 'mrs';
```

I got the idea of using attributes to describe enum cases from this [blogpost](https://kirschbaumdevelopment.com/insights/extending-php-enums-with-attributes). The php code runs fine without any issues, but Prettier thinks it's a syntax error.

acharron-hl:
Fixes https://github.com/prettier/plugin-php/issues/2293
Fixes https://github.com/prettier/plugin-php/issues/2486

## Dependency update note

This PR relies on a new version of php-parser: 3.7.0.

I was initially setting out to just fix the attribute formatting issues, but that required a new version of the php-parser. Unfortunately the currently published 3.6 has a regression around the unary operator reported as #2486 which means we can't just update to that.

## Attribute formatting fixes

Fixes https://github.com/prettier/plugin-php/issues/2293

The updated `php-parser` exposes attribute groups as distinct AST nodes on methods, functions, class constants, and enum cases. I madechanges so attributes and their comments format correctly.

**Changes**

- **`getCommentChildNodes`** — attribute groups are now comment attachment targets on `function`/`method`, `classconstant`, and `enumcase`, so comments between `#[...]` blocks attach to the right node instead of being skipped or misplaced.
- **`printAttrs` / `printAttrGroup`** — each attribute group is printed through `printAllComments()`, so leading and trailing comments on groups are actually emitted.
- **`enumcase` printing** — attributes on enum cases are now printed before `case` (new coverage in `tests/enum/enum-case-attributes.php`).
- **`handleClassMemberStatementComments`** — comments before a property or constant identifier are attached to the enclosing member statement, keeping modifiers like `public static` together when comments appear between modifiers and the name.

**Snapshot impact**

| Test | What changed |
|------|--------------|
| `tests/attributes/attributes.php` | Comments between `#[S]` and `#[T]` now print in place; `//Testing T` no longer incorrectly trails the previous method's `}`. |
| `tests/comments/method.php` | Dangling comment in empty `()` stays inside parens; inline comments between `static` and `$foo` are all preserved. |
| `tests/enum/enum-case-attributes.php` | New snapshot for attributed enum cases. |

---

## Unary operator / exponentiation fix (upstream parser)

Fixes https://github.com/prettier/plugin-php/issues/2486

This change comes from `php-parser`, not from new formatting logic in the plugin. PHP gives `**` higher precedence than unary `+` and `-`, so unparenthesized expressions parse differently from explicitly parenthesized ones:

```php
+$a ** 1   // PHP parses as:  +($a ** 1)
(+$a) ** 1 // PHP parses as:  (+$a) ** 1
```

The old parser incorrectly parsed `+$var ** 2` as `(+$var) ** 2`. The formatter faithfully reproduced that wrong AST, so both forms printed identically. The updated parser builds the correct tree — unary wrapping exponentiation for the unparenthesized form, and a `bin(**)` with `parenthesizedExpression: true` when parens are explicit.

Only the snapshots needed updating.

**Snapshot impact**

| Test | Input | Old output | New output |
|------|-------|------------|------------|
| `tests/parens/bin.php` | `$var = +$var ** 2;` | `(+$var) ** 2` | `+($var ** 2)` |
| `tests/parens/unary.php` | `$a = +$a ** 1;` | `(+$a) ** 1` | `+($a ** 1)` |

Explicitly parenthesized inputs such as `(+$var) ** 2` and `(+$a) ** 1` are unchanged.

This is a semantic correction: the old output did not match how PHP actually parses unparenthesized `+$x ** n` expressions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   ],
   "dependencies": {
     "linguist-languages": "^8.0.0",
-    "php-parser": "^3.4.0"
+    "php-parser": "https://github.com/glayzzle/php-parser.git#0b3dbe8833965ae14915e722a1b974e3616df78c"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.27.2",
@@ -82,5 +82,6 @@
     "include": [
       "src/**"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
--- a/package.json
+++ b/package.json
@@ -82,6 +82,5 @@
     "include": [
       "src/**"
     ]
-  },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "@prettier/plugin-php",
-  "version": "0.25.0",
+  "name": "@vanilla/prettier-plugin-php",
+  "version": "0.26.0",
   "description": "Prettier PHP Plugin",
   "repository": "prettier/prettier-php",
   "author": "Lucas Azzola <@azz>",
@@ -27,7 +27,7 @@
   ],
   "dependencies": {
     "linguist-languages": "^8.0.0",
-    "php-parser": "https://github.com/glayzzle/php-parser.git#0b3dbe8833965ae14915e722a1b974e3616df78c"
+    "php-parser": "^3.7.0"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.27.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "@vanilla/prettier-plugin-php",
-  "version": "0.26.0",
+  "name": "@prettier/plugin-php",
+  "version": "0.25.0",
   "description": "Prettier PHP Plugin",
   "repository": "prettier/prettier-php",
   "author": "Lucas Azzola <@azz>",
--- a/src/printer.mjs
+++ b/src/printer.mjs
@@ -1199,37 +1199,150 @@ function printAttrs(path, options, print, { inline = false } = {}) {
   if (!path.node.attrGroups) {
     return [];
   }
-  path.each(() => {
-    const attrGroup = ["#["];
+  path.each((attrGroupPath) => {
     if (!inline && allAttrs.length > 0) {
       allAttrs.push(hardline);
     }
-    attrGroup.push(softline);
-    path.each(() => {
-      const attrNode = path.node;
-      if (attrGroup.length > 2) {
-        attrGroup.push(",", line);
-      }
-      const attrStmt = [attrNode.name];
-      if (attrNode.args.length > 0) {
-        attrStmt.push(printArgumentsList(path, options, print, "args"));
-      }
-      attrGroup.push(group(attrStmt));
-    }, "attrs");
     allAttrs.push(
-      group([
-        indent(attrGroup),
-        ifBreak(shouldPrintComma(options, 8.0) ? "," : ""),
-        softline,
-        "]",
-        inline ? ifBreak(softline, " ") : "",
-      ])
+      printAllComments(
+        attrGroupPath,
+        () => printAttrGroup(attrGroupPath, options, print, { inline }),
+        options
+      )
     );
   }, "attrGroups");
   if (allAttrs.length === 0) {
     return [];
   }
   return [...allAttrs, inline ? "" : hardline];
+}
+
+function printAttrGroup(path, options, print, { inline = false } = {}) {
+  const attrGroup = ["#["];
+  attrGroup.push(softline);
+  path.each(() => {
+    const attrNode = path.node;
+    if (attrGroup.length > 2) {
+      attrGroup.push(",", line);
+    }
+    const attrStmt = [attrNode.name];
+    if (attrNode.args.length > 0) {
+      attrStmt.push(printArgumentsList(path, options, print, "args"));
+    }
+    attrGroup.push(group(attrStmt));
+  }, "attrs");
+  return group([
+    indent(attrGroup),
+    ifBreak(shouldPrintComma(options, 8.0) ? "," : ""),
+    softline,
+    "]",
+    inline ? ifBreak(softline, " ") : "",
+  ]);
+}
+
+function printFunction(path, options, print) {
+  const { node } = path;
+  const declAttrs = printAttrs(path, options, print, {
+    inline: node.kind === "closure",
+  });
+  const declaration = [];
+
+  if (node.isFinal) {
+    declaration.push("final ");
+  }
+
+  if (node.isAbstract) {
+    declaration.push("abstract ");
+  }
+
+  if (node.visibility) {
+    declaration.push(node.visibility, " ");
+  }
+
+  if (node.isStatic) {
+    declaration.push("static ");
+  }
+
+  declaration.push("function ");
+
+  if (node.byref) {
+    declaration.push("&");
+  }
+
+  if (node.name) {
+    declaration.push(print("name"));
+  }
+
+  declaration.push(printArgumentsList(path, options, print));
+
+  if (node.uses && node.uses.length > 0) {
+    declaration.push(
+      group([" use ", printArgumentsList(path, options, print, "uses")])
+    );
+  }
+
+  if (node.type) {
+    declaration.push([
+      ": ",
+      hasDanglingComments(node.type)
+        ? [
+            path.call(() => printDanglingComments(path, options, true), "type"),
+            " ",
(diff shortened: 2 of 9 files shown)


jorgsowa:
Please don't merge fixes and put into description only relevant information. It's hard to review such PRs.

> This PR relies on a new / unpublished version of php-parser (which will likely be version 3.6.1 or 3.7.0.

This is not acceptable solution.

acharron-hl:
> This is not acceptable solution.

I'm aware. That's why I noted in the first section

> It should not be merged until that version is published. In the meantime, this PR depends on the commit at the tip of the main branch over there (which includes [the unary operator fix](https://github.com/glayzzle/php-parser/commit/f5162323f5760da86f9b697e0c3761547277fc9c).
>
> I was initially setting out to just fix the attribute formatting issues, but that required a new version of the php-parser. Unfortunately the currently published 3.6 has a regression around the unary operator reported as https://github.com/prettier/plugin-php/issues/2486 which means we can't just update to that.

---

> put into description only relevant information

I don't think it's necessarily overly verbose? I'm mostly just explaining which snapshots are updating any why. The unary operator change in particular is a bit finicky and not immediately obvious to me what was wrong with the initial implementation, which is why I wrote up the examples.

czosel:
@acharron-hl php-parser v3.7.0 has just been released :+1:

acharron-hl:
@jorgsowa @czosel I've updated the php-parser version to 3.7.0. I removed the `packageManager` declaration that was auto-added to package.json as well when I ran `yarn install` with corepack enabled on my node version.

The tests passed locally for me at least.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@prettier/plugin-php",
-  "version": "0.25.0",
+  "name": "@vanilla/prettier-plugin-php",
+  "version": "0.26.0",
```
jorgsowa:
```suggestion
  "version": "0.25.0",
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@prettier/plugin-php",
-  "version": "0.25.0",
+  "name": "@vanilla/prettier-plugin-php",
```
jorgsowa: Why did you change this?

src/printer.mjs
```diff
@@ -1232,6 +1217,134 @@ function printAttrs(path, options, print, { inline = false } = {}) {
...
+  return group([
+    indent(attrGroup),
+    ifBreak(shouldPrintComma(options, 8.0) ? "," : ""),
+    softline,
+    "]",
+    inline ? ifBreak(softline, " ") : "",
+  ]);
+}
+
+function printFunction(path, options, print) {
```
jorgsowa: Why did you move this function? It's harder to compare the git diff with it

tests/attributes/__snapshots__/jsfmt.spec.mjs.snap
```diff
@@ -134,12 +134,12 @@ class D
     function k(#[L] int $m): callable
     {
         return #[N, O] #[P] fn(#[Q] int $r) => $r * 2;
-    } //Testing T
```
jorgsowa, line 137: You can see that something is messed up with comments. Why did this comment change its position?

Do you know whether it's a fault of the PHP parser or your PR?

acharron-hl: This is from the changes I made to comment/attribute formatting (Item 1 in my snapshot changes listed in the PR description).

So this is the raw unformatted code

```php
    }

    // Testing S
    #[S]
//Testing S-T
#[T] //Testing T
```

This is the old snapshot

```php
    } //Testing T          ← comment moved here (wrong)
    // Testing S
    //Testing S-T          ← moved above #[S] (wrong)
    #[S]
    #[T]   
```

And my udpated snapshot

```php
    }                       ← no stray comment
    // Testing S
    #[S]
    //Testing S-T          ← stays between #[S] and #[T]
    #[T] //Testing T        ← comment stays with #[T]
    private function u()
```

It's kind of hard to tell in the git diff that's for sure. One of the negatives of such a huge snapshot is that the input/output are hundreds of lines apart when looking at the snapshot.

package.json
acharron-hl: Whoops sorry, reverted.

package.json
acharron-hl: Sorry, reverted back, was doing in some internal linking for testing against our own codebase.

src/printer.mjs
acharron-hl: I moved it back, you are correct, the diff is much cleaner now.

tests/comments/__snapshots__/jsfmt.spec.mjs.snap
```diff
@@ -4922,9 +4922,9 @@ class Foo
 
     public function emptyMethod(/* comments */) {}
 
-    abstract public function sortByName(/* bool $useNaturalSort = false */); /* comment */
+    abstract public function sortByName(/* bool $useNaturalSort = false */);
 
-    /* comment */ /* comment */ protected static $foo; /* comment */
+    /* comment */ /* comment */ /* comment */ protected static $foo; /* comment */
```
jorgsowa, line 4927:
```suggestion
/* comment */ protected /* comment */ static /* comment */ $foo /* comment */;
```

Everything looks correct, except of this part. Some comments moved into wrong places.

src/printer.mjs
jorgsowa: Thanks, it's better now.

tests/comments/__snapshots__/jsfmt.spec.mjs.snap
acharron-hl: This doesn't feel quite right either?

So the unformatted was this:

We have 4 comment blocks attached to the property in total. One before the declaration, 2 inside of it, and 2 after.

```php
    abstract public function sortByName(/* bool $useNaturalSort = false */);

    /* comment */ protected /* comment */ static /* comment */ $foo /* comment */;

    public function foo( // Comment
    ) {}
```

Previously it formatted to this, which emits the leading comment in the spot and emits the inner comments before the declaration.

```php
    abstract public function sortByName(/* bool $useNaturalSort = false */); /* comment */  ← stolen

    /* comment */ /* comment */ protected static $foo; /* comment */                      ← 1 comment missing
```

It just fixed the leading comment getting emitted in the wrong spot, but did not modify the behaviour of comments getting placed in between property keywords.

```php
    abstract public function sortByName(/* bool $useNaturalSort = false */);

    /* comment */ /* comment */ /* comment */ protected static $foo; /* comment */        ← all 4 preserved
```

I guess what it really comes down to is:

**Do we want to allow comments on property keywords or not.** In my mind, that doesn't make much sense at all (If I saw someone putting PHPDoc between the modifier and the property I'd be asking wtf they are doing), so I preserve them but emit them as leading comments on the property declaration. In any case it's an improvement over it getting attached to the previous node.

acharron-hl: I'll note as well supporting preservation of property keywords gets a bit gnarlier when you consider we'd want to do it for constants as well, and there are a lot of other modifiers as well (types, readonly, attributes, multi property declarations etc).

acharron-hl:
@jorgsowa So it seems we're held up on if the project should preserve comments in between visibility modifiers or not. If you want I can implement it. Anything to get our growing list of enums out of our `.prettierignore` file.

jorgsowa (approved):

jorgsowa:
Sorry, for delay. Looks good. Thanks for contribution

